PR #2, opened 2022-04-07
Cache component deployment

goldmann:


goldmann:
@stuartwdouglas I'm leaving it for your review now. I added a few things that make it easier to work with.

There is one thing that I'm not sure about - the application works fine in development mode when run as `mvn quarkus:dev -pl cache` but it does not work correctly when run as `java -jar cache/target/quarkus-app/quarkus-run.jar` or packaged in the container. No matter if using dev profile or prod. It always returns 404 on any resource. Since I'm finishing now, I'll leave this mystery to you :)

stuartwdouglas (approved):

PR #7, opened 2022-04-08
fix(cache): instantiate the build policy only once

goldmann:
Make the build policy a singleton preventing from instantiating
it many times, for example in health check.


goldmann:
I found that when I hit the health check, it gets produced again.

PR #18, opened 2022-04-22
Initial work around creating a Recipe Database

stuartwdouglas:
@goldmann This is still pretty rough, but we might be better off just getting it in for now to reduce the risk of conflicts. 

The results of this can be seen at https://github.com/stuartwdouglas/temp-hacbs-recipe-data


goldmann (approved):

goldmann:
I merged it, we are at the stage where things should probably just be merged, we need to make sure we don't drop the ball and have everything documented well.

PR #29, opened 2022-05-10
Fix memory leak

stuartwdouglas:


goldmann:
@stuartwdouglas Can you check the failure? It looks like CI doesn't like your last change of requiring JDK 17.

goldmann (approved):

PR #35, opened 2022-05-16
Move to using OwnerReferences to track ownership

stuartwdouglas:


pkg/reconciler/artifactbuildrequest/artifactbuildrequest.go
```diff
@@ -120,106 +94,55 @@ func (r *ReconcileArtifactBuildRequest) Reconcile(ctx context.Context, request r
...
-				//now lets look for an existing build object
-				list := &v1alpha1.DependencyBuildList{}
-				lbls := map[string]string{
-					dependencybuild.IdLabel: depId,
-				}
-				listOpts := &client.ListOptions{
-					Namespace:     abr.Namespace,
-					LabelSelector: labels.SelectorFromSet(lbls),
+				if !found {
+					db.OwnerReferences = append(db.OwnerReferences, v1.OwnerReference{APIVersion: v1alpha1.SchemeGroupVersion.String(), Kind: ArtifactBuildRequestKind, Name: abr.Name, UID: abr.UID})
```
gabemontero, line 126: as with most things, controller runtime has a helper for "common done things" like setting owner refs .... see 
https://github.com/redhat-appstudio/build-service/blob/23ab1ed1114d04210b9b511f457160baeba2b535/controllers/component_build_controller.go#L213
and 
https://github.com/redhat-appstudio/build-service/blob/23ab1ed1114d04210b9b511f457160baeba2b535/controllers/component_build_controller.go#L213

taking a quick scan, feels like you could probably use that helper and replace some of the logic you've written

did you just miss the presence of that helper, or am I missing some nuance here as to why you can't use it?

pkg/reconciler/artifactbuildrequest/artifactbuildrequest.go
```diff
@@ -120,106 +94,55 @@ func (r *ReconcileArtifactBuildRequest) Reconcile(ctx context.Context, request r
...
-				var scmType string
-				var message string
-				var path string
+			//no existing build object found, lets create one
+			db := &v1alpha1.DependencyBuild{}
+			db.Namespace = abr.Namespace
+			db.Labels = lbls
+			//TODO: name should be based on the git repo, not the abr, but needs
+			//a sanitization algorithm
+			db.GenerateName = abr.Name + "-"
```
gabemontero, line 103: switching to generate name is new, right?

not a bad thing, but I'm curious why you pivoted to that.  Can you elaborate?

pkg/reconciler/pipelinerun/controller.go
```diff
@@ -13,14 +13,13 @@ func SetupNewReconcilerWithManager(mgr ctrl.Manager) error {
...
-				return true
+				return false
 			},
 			UpdateFunc: func(e event.UpdateEvent) bool {
 				return true
 			},
 			DeleteFunc: func(e event.DeleteEvent) bool {
-				//TODO possibly change to false over time
-				return true
+				return false
```
gabemontero, line 22: this is fine / not surprising, but it does remind me, I put a quick blurb in the jira, we need at some point to start thinking about our pruning strategy.  Or has thought been put into it and I missed it?

the owner ref usage will help with pruning ... i.e. deleting a DB or ABR will delete any tekton objects they create/own

pkg/reconciler/pipelinerun/controller.go
```diff
@@ -13,14 +13,13 @@ func SetupNewReconcilerWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&pipelinev1beta1.PipelineRun{}, builder.WithPredicates(predicate.Funcs{
 			CreateFunc: func(e event.CreateEvent) bool {
-				return true
+				return false
```
gabemontero, line 16: fwiw I had meant to put TODO commit here as well saying this might be switched to false

pkg/apis/jvmbuildservice/v1alpha1/artifactbuildrequest_types.go
```diff
@@ -39,7 +43,8 @@ const (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=artifactbuildrequests,scope=Namespaced
-
+// +kubebuilder:printcolumn:name="GAV",type=string,JSONPath=`.spec.gav`
```
gabemontero, line 46: ah cool, you are diving into the multitude of kubebuilder options

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -48,29 +47,22 @@ func (r *ReconcilePipelineRunRequest) Reconcile(ctx context.Context, request rec
...
-		err = r.client.List(ctx, &list, listOpts)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
 		success := pr.Status.GetCondition(apis.ConditionSucceeded).IsTrue()
-		for _, abr := range list.Items {
-			if abr.Status.State == v1alpha1.DependencyBuildStateBuilding {
+		for _, ref := range pr.OwnerReferences {
+			db := v1alpha1.DependencyBuild{}
+			if err = r.client.Get(ctx, types.NamespacedName{Namespace: pr.Namespace, Name: ref.Name}, &db); err != nil {
```
gabemontero, line 56: awesome

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -45,24 +49,43 @@ func (r *ReconcileTaskRunRequest) Reconcile(ctx context.Context, request reconci
...
-			LabelSelector: labels.SelectorFromSet(lbls),
-		}
-		err = r.client.List(ctx, &list, listOpts)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		for _, abr := range list.Items {
+		for _, ref := range tr.OwnerReferences {
+			abr := v1alpha1.ArtifactBuildRequest{}
+			if err = r.client.Get(ctx, types.NamespacedName{Namespace: tr.Namespace, Name: ref.Name}, &abr); err != nil {
```
gabemontero, line 79: awesome

gabemontero (changes requested):
mostly anecdotal comments, but 1 suggestion in line (apply to where you set owner refs)

but overall you are obviously getting the hang of this quickly @stuartwdouglas 

now on task_run_controller_test.go, I missed that before, but that looks like an e2e test vs. a unit test

typically, we keep unit tests in line with source code, but e2e's in a separate top level dir from the source code.

we can dive into this when we have the "testing" discussion you alluded to in slack @stuartwdouglas  ... I can certainly spend cycles on such a reorganization if it make sense for you pivot elsehwere ... we'll sort that out when we talk.

pkg/reconciler/artifactbuildrequest/artifactbuildrequest.go
stuartwdouglas: Nope, it was moved from here: https://github.com/redhat-appstudio/jvm-build-service/pull/35/files#diff-512dbf53bb69666cabbc97bb8bc3bd9caa90fbfedaac29c57c14101c513d5e68L186

generatedName is used because ATM this name is not guaranteed to be stable, it depends on which ABR generates the build. If we had a way of generating human readable stable names from SCM URI + context path that meet the kube naming rules then we could use that, but it did not seem worth spending time on at this point.

pkg/reconciler/pipelinerun/controller.go
stuartwdouglas: Once the component is built we can technically prune everything, but we will likely want to stash the results somewhere in case people want info from the logs. We probably want to stash these in the same long term storage as the artifacts (e.g. S3 ATM).

stuartwdouglas:
@gabemontero I have updated the PR.

gabemontero (approved):
yeah let's merge this, and then deal with the e2e / unit test restructuring in separate PRs

I'm *thinking* I could bootstrap some of that, including reorganizing task_run_controller_test.go, while you have plenty of other things to focus on, but we can drill down on that when we talk voice to voice @stuartwdouglas

PR #40, opened 2022-05-18
Report on contamination of rebuilt dependencies

stuartwdouglas:
This will not allow the deployment of rebuilt deps that have had non-rebuilt dependencies shaded in. It also reports on these dependencies, but will not automatically resolve the situation yet.


gabemontero (approved):

gabemontero:
@stuartwdouglas assuming this PR falls under `PLNSRVCE-292` umbrella, if you put `PLNSRVCE-292: ` at the beginning of the PR title the git/jira bots will automatically link this PR to the jira.  I'd like for us to start doing that, if you are amenable.

My guess is we don't have prow configuration set up for this repo, so I cannot do it myself via the `/retitle` prow command.

PR #42, opened 2022-05-19
POC to vet only updating CRD status in the associated CRD's reconciler

gabemontero:
ok @stuartwdouglas - in the end I deferred on looking at testing to dive a bit into this PR's subject, which we go into during our last call.

Part of that "holistic review" I mentioned in the team channel slack threads my Thursday AM 

Preface:
- I checked out #41 locally and worked off of that
- just compiled no test (of course I'm not set up to test yet)

This is very much still a first pass.  Most notably, I might want to take another pass at registering pipelinerun success/failure on the dependencybuild in a different way.  I might switch to adding an annotation on the DB with the lookup information for the pipelinerun to inspect of success / failure.  If I read it right you are only setting a label on the pipelinerun to designate it is from a DB, not the specific DB.  Which is fine.

So potential reasons to move forward with some form of this:
1) I may have eliminated the need for the ABR discovered state, but of course testing needs to vet this
2) this is more in line with best practices and single writer for status 
3) I thought about it some more since we last talked, and if for scaling purposes we had to break out these reconcilers into separate processes / deployments, we really need at that point to adhere to 2), both from an RBAC perspective as well as a single writer perspective.

Take a peek, see what you think, and if need be, let's chat my sometime prior to the Tuesday workshop.

thanks


pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
```diff
@@ -13,19 +13,21 @@ const (
...
 type DependencyBuildSpec struct {
-	SCMURL  string `json:"scmURL,omitempty"`
-	SCMType string `json:"scmType,omitempty"`
-	Tag     string `json:"tag,omitempty"`
-	Path    string `json:"path,omitempty"`
+	SCMURL             string   `json:"scmURL,omitempty"`
+	SCMType            string   `json:"scmType,omitempty"`
+	Tag                string   `json:"tag,omitempty"`
+	Path               string   `json:"path,omitempty"`
+	PipelineRunSuccess bool     `json:"pipelineRunSuccess,omitempty"`
```
stuartwdouglas, line 20: Maybe a string state field would make more sense here.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -62,6 +62,8 @@ func (r *ReconcilePipelineRunRequest) Reconcile(ctx context.Context, request rec
 		for _, ref := range pr.OwnerReferences {
 			dep := v1alpha1.DependencyBuild{}
 			if err = r.client.Get(ctx, types.NamespacedName{Namespace: pr.Namespace, Name: ref.Name}, &dep); err != nil {
+				//TODO do we want to loop through all the owner refs and only return and error if we do not get success on
```
stuartwdouglas, line 65: Yes, this should just continue I think rather than returning on error.

pkg/apis/jvmbuildservice/v1alpha1/artifactbuildrequest_types.go
```diff
@@ -7,17 +7,20 @@ import (
 type ArtifactBuildRequestSpec struct {
 	// GAV is the groupID:artifacdtID:version tuple seen in maven pom.xml files
 	GAV string `json:"gav,omitempty"`
+	Message           string `json:"message,omitempty"`
+	SCMURL            string `json:"scmURL,omitempty"`
```
stuartwdouglas: It does feel like this change means our implementation is now leaking out into the CRD, as conceptually these really are part of the status, they are not things that a user would update.

gabemontero: yeah admittedly I'm back and forth a bit on these vectors of consideration:
- which reconcilers maintain which fields, and do we need to adhere to the CRD specific reconciler conventions. 
- do these CRDs, this data model, truly fit the "spec/status" paradigm ... I can point you to serveral "request" types in the openshift ecosystem that do *NOT* have spec/status.  For example:  the [openshift build request](https://github.com/openshift/api/blob/master/build/v1/types.go#L1247-L1280) ... I wonder if the ABR for example is conceptually more akin to that.  If so, we could split things out for example, and have an `ArtifactBuildRequest` that results in the creation of an `ArtifactBuild`
- who exactly are the "users" of these CRD specifically (vs. the overall system in general) and how do they interact with the system.  Even after watching your demo videos, I'm not 100% sure on the answers to that yet.

That said, I think all that ^^ dovetails precisely into the theme of tomorrow's next workshop that @sbose78 has set up.  If anything else, I think this PR could kick start discussion on the CRD API, where we will get to consensus one way or the other, and even if we throw this PR away, as long as we get clarity and consensus, the greater purpose has been served.

PR #43, opened 2022-05-23
Updating java base image to openjdk-17

Michkov:
Required java version is 17, updating base images as well.


stuartwdouglas:
So these dockerfiles should actually be deleted, they are auto generated when the project is created but I ended up using Jib instead (and just forgot to delete the uneeded dockerfiles).

Alternatively we could move away from jib and use docker instead, but in general jib based builds are likely going to be easier to maintain as Quarkus controls the image generation process.

PR #51, opened 2022-05-26
Expand testing for DependencyBuild

stuartwdouglas:


pkg/apis/jvmbuildservice/v1alpha1/artifactbuildrequest_types.go
```diff
@@ -5,18 +5,15 @@ import (
...
-	Message       string `json:"message,omitempty"`
-	RecipeGitHash string `json:"recipeGitHash,omitempty"`
-	SCMURL        string `json:"scmURL,omitempty"`
-	SCMType       string `json:"scmType,omitempty"`
-	Tag           string `json:"tag,omitempty"`
-	Path          string `json:"path,omitempty"`
+	//TODO: conditions?
+	State   string  `json:"state,omitempty"`
+	Message string  `json:"message,omitempty"`
+	ScmInfo SCMInfo `json:"scm,omitempty"`
```
gabemontero, line 16:
```suggestion
	SCMInfo SCMInfo `json:"scm,omitempty"`
```

I think we need to be case consistent here

pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
```diff
@@ -13,10 +13,7 @@ const (
 )
 
 type DependencyBuildSpec struct {
-	SCMURL  string `json:"scmURL,omitempty"`
-	SCMType string `json:"scmType,omitempty"`
-	Tag     string `json:"tag,omitempty"`
-	Path    string `json:"path,omitempty"`
+	ScmInfo SCMInfo `json:"scmInfo,omitempty,inline"`
```
gabemontero, line 16:
```suggestion
	SCMInfo SCMInfo `json:",inline"`
```

I've never seen folks inline the way you had it.  Did you try it with just inline first and it did not work for some reason?

gabemontero (changes requested):

pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
stuartwdouglas: ah, that would be why it was not working

PR #54, opened 2022-05-30
Multiple builder images

stuartwdouglas:
This fixes a lot of problems with the existing code, and also adds the ability to try builds with multiple different recipes.

The main issue was that when the reconciler was notified about a change in state for a PR/TR the client might still have stale state cached, so the change that we were being notified about might not actually be visible to the client.

I worked around this by using the non-caching client when doing operations on PR/TR so we can be sure to see up-to-date results.

@gabemontero the current code is very broken (creates way to many builds as it does not see the existing builds from the client), so it would be great to get this in ASAP.


stuartwdouglas:
@gabemontero the current master is broken in a lot of different ways, and this seems to fix it. IMHO we should get this in ASAP and if you have any concerns deal with them in a follow up PR

gabemontero:
On Mon, May 30, 2022 at 8:17 AM Stuart Douglas ***@***.***>
wrote:

> @gabemontero <https://github.com/gabemontero> the current master is
> broken in a lot of different ways, and this seems to fix it. IMHO we should
> get this in ASAP and if you have any concerns deal with them in a follow up
> PR
>
> sure get this in and will level set when I am back in the office

I still wonder if your caching client issues ate in fact the server you are
running against but we will dive into that after I look in detail and we
can talk

Using the non caching client long term is not ideal

stuartwdouglas:
You think it might be a minikube issue? 

It was definitely reproducible, around 20% of builds would have the issue. Basically once I had created a TR/PR sometimes it would just not show up in a subsequent List() call. Requeing with a timeout would 'fix' it, but in a really fragile way.

This would mostly (exclusivly?) happen in events caused by the watch on the PR/TR rather than on the main DependencyBuild/ArtifactBuildRequest resource, so I wonder if that is related.

stuartwdouglas:
From what I read today it sounds like the cache is invalidated by its own watch, so I am wondering if sometimes we see the watch event before the cache invalidation happens?

Michkov (approved):

gabemontero:
> You think it might be a minikube issue?
> 

If not minikube per se, a function of running mini kube on your laptop where the api server becmes resource constrainted.  API servers can throttle, watch events are not guaranteed (hence controller relist for eventual consistency), though in a non constrained system it should be really ray.

I learned last week that some of the team has started using clusters provisioned on upshift.  We need to get you to try thiins on either that (I suspect @Michkov can help you get an upshift cluster), or al least have you try to use the "Cluster Bot" app on slack (heads up, it used to be "cluster-bot" but they just renamed it ever so slightly. I'll follow up with @Michkov during scrum on the specifics, unless you and he can talk before your day is done.

From Clsuter Bot, you should be able to say `launch aws` and get a cluster in about 30 minutes.  It will stay up for 2 to 3 hous.

Conversely, not knowing yeah all the items waiting for me when I officially ready all email /messages tomorrow, I want to bring up a cluster, run @Michkov 's latest intra-deployment to bring up the jvm buidl service, and try the service registry build.

Aside from applying https://github.com/redhat-appstudio/jvm-build-service/blob/main/hack/examples/run-service-registry.yaml, what else do I need to do to try it out @stuartwdouglas ?

thanks

> It was definitely reproducible, around 20% of builds would have the issue. Basically once I had created a TR/PR sometimes it would just not show up in a subsequent List() call. Requeing with a timeout would 'fix' it, but in a really fragile way.
> 
> This would mostly (exclusivly?) happen in events caused by the watch on the PR/TR rather than on the main DependencyBuild/ArtifactBuildRequest resource, so I wonder if that is related.

gabemontero:
> From what I read today it sounds like the cache is invalidated by its own watch, so I am wondering if sometimes we see the watch event before the cache invalidation happens?

the controller runtime watch feature is a bit new to me, so maybe that is what is going on, but in gernal with controllers, you should not have to repeatedly use the real client vs. controller client.

I'll add tracking down your question to the list of jvm-build-service items I try to take on Tuesday, if not WEdnesday.

stuartwdouglas:
So at the moment there is a little bit of stuffing around to deploy. For minikube you can do:

```
export QUAY_USERNAME=myquayusername
make dev
```
This will build your local workspace and push it to your quay repo, and then run ./deploy/development.sh to add this to your cluster.

One complexity is that by default the quay repositories will be private, so you need to go to the quay UI and make them public or all your pods will fail to pull the images (this only happens the first time you run it). 

If you seen an error you may need to run ./deploy/development.sh again, as the tekton CRD's might not be ready when we try and install our tasks.

You can then run a build with:

```
./hack/example/run-sample-build.sh
```

For deployment to openshift it is a bit different as you don't want to install Tekton, so you need to comment out the line that installs it in `deploy/base/kustomization.yaml`, and instead install openshift pipelines through the operator hub. Otherwise the instructions should be mostly the same.

Once this is running if you want to see code changes in the operator you can either run make dev-image to push your changes to quay and then restart the operator, or you can shut down the operator deployment and just run it locally.

gabemontero:
OK after bumping maven to 3.8.5 (from 3.6.3) and installing java 17 I was able to get ^^  for the service registry build to work, at least along the order of magnitude I believe @stuartwdouglas has conveyed to me in the past.

Some details
- 1 pipelinerun, was successful
- 128 taskruns, 108 successful, 20 failed
- 84 artifactbuilds
- 29 dependencybuilds

First blush, I was expecting to see artifactbuild/dependencybuild in failed state to correspond with the failed task runs, but I may just not have the latest state machine details down.

I'll see if I can discover why / see if that makes sense as I change the watcher per the discussion @stuartwdouglas and I had last night.

PR #60, opened 2022-06-01
Initial discovery prototype

stuartwdouglas:
Initial prototype for build discovery.

The build request processor will examine the repo, and put any information it finds into a JSON field on the artifactbuildrequest. 

The reconciler will then turn this into actual build recipes, and create the DependencyBuild with this information.

This is very much just a PoC, but is the start of supporting gradle builds.


gabemontero:
probably makes sense @dwalluck for you to start looking a this as well, since it is a building block to gradle builds

deploy/operator/base/rbac.yaml
```diff
@@ -47,6 +47,13 @@ rules:
       - patch
       - update
       - watch
+  - apiGroups:
+      - ''
+      - events.k8s.io
+    resources:
+      - events
+    verbs:
+      - create
```
gabemontero, line 56: ooops forgot this when I started adding events / apologies

deploy/overlays/dev-template/kustomization.yaml
```diff
@@ -46,7 +46,7 @@ patches:
          name: QUARKUS_S3_ENDPOINT_OVERRIDE
          value: "http://localstack.default.svc.cluster.local:4572"
    target:
-     kind: Task
+     kind: ClusterTask
```
gabemontero, line 49: interesting

can you clarify the reason it make sense to go from namespace scope to cluster scope?

java-components/adhoc-analyser-tool/src/main/java/com/redhat/hacbs/analyser/kube/ResetDependencyBuildsCommand.java
```diff
@@ -36,7 +36,7 @@ public void run() {
             } else {
                 List<DependencyBuild> items = client.list().getItems();
                 for (var request : items) {
-                    if (!failedOnly || request.getStatus().getState().equals("DependencyBuildFailed")) {
+                    if (!failedOnly || request.getStatus().getState().equals("DependencyBuildStartFailed")) {
```
gabemontero, line 39: missed this in previously PR etc. examinations

why do we have java runnables manipulating status for a CRD being handled by our golang operartor / controller?
should be be concerned with multiple writers getting in the way of each other, possibly corrupting the flow through the state machine?

pkg/reconciler/artifactbuild/artifactbuildrequest.go
```diff
@@ -191,7 +218,7 @@ func (r *ReconcileArtifactBuild) handleStateDiscovering(ctx context.Context, abr
 			SCMType: abr.Status.SCMInfo.SCMType,
 			Tag:     abr.Status.SCMInfo.Tag,
 			Path:    abr.Status.SCMInfo.Path,
-		}}
+		}, BuildRecipes: buildRecipes}
```
gabemontero, line 221: my comment applies to the line below, which is unchanged, https://github.com/redhat-appstudio/jvm-build-service/pull/60/files#diff-e8e0091cd27c2b3a10a3f5ec14aa8c07390ca462d30847514323d45e2470ff6eR222 ... i've missed this detail previously as well.

just noticed we are calling status().update(), but we are manipulating the spec too

and this is actually working?  my guess is it is working, otherwise you would have ran into issues already, but at the lower levels of the k8s api, when you define a status subresource a la https://github.com/redhat-appstudio/jvm-build-service/blob/main/pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go#L41 and https://github.com/redhat-appstudio/jvm-build-service/blob/main/deploy/crds/base/jvmbuildservice.io_dependencybuilds.yaml#L175-L176 you have to call status.update to update the status and update when you update the spec/metadata

either controller-gen is not crafting status subresource correctly, or controller-gen's client must do some machinations to handle things regardless of which update you call

or I'm just misremembering how the lower level k8s update status works

maybe I investigate to satisfy my curiosity if you say "yes, it just works", but we'll see if time permits

gabemontero (commented):
ok here you go @stuartwdouglas 

more "macro" items vs. detailed logic analysis (the logic itself seemed pretty straight forward)

java-components/adhoc-analyser-tool/src/main/java/com/redhat/hacbs/analyser/kube/ResetDependencyBuildsCommand.java
stuartwdouglas: Just to resolve the conversation here, this is a testing utility that makes it easy to retry failed jobs when doing local development. It is not actually running on the cluster.

pkg/reconciler/artifactbuild/artifactbuildrequest.go
stuartwdouglas: We are working on two different objects. We update the status of the AB and create a new DependencyBuild.

java-components/adhoc-analyser-tool/src/main/java/com/redhat/hacbs/analyser/kube/ResetDependencyBuildsCommand.java
gabemontero: yep per our conversation last night, occasionally my old eyes struggle with my 4K monitor :-)

deploy/overlays/dev-template/kustomization.yaml
gabemontero: we got into this one last night too ... this is in response to Michal changing the target over in infra / build defs as part of his resource limiting effort

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/LookupBuildRecipesCommand.java
```diff
@@ -162,4 +183,32 @@ public void run() {
...
+            if (context != null) {
+                path = path.resolve(context);
+            }
+            BuildInfo info = new BuildInfo();
+            info.tools.put("jdk", new VersionRange("8", "17", "11"));
+            if (Files.isRegularFile(path.resolve("pom.xml"))) {
+                info.tools.put("maven", new VersionRange("3.8", "3.8", "3.8"));
+                info.invocations.add(
+                        new ArrayList<>(List.of("clean", "install", "-DskipTests", "-Denforcer.skip", "-Dcheckstyle.skip")));
+            } else if (Files.isRegularFile(path.resolve("pom.xml"))) {
```
dwalluck, line 200: We would want `build.gradle` or `build.gradle.kts` here.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/LookupBuildRecipesCommand.java
```diff
@@ -162,4 +183,32 @@ public void run() {
...
+                info.tools.put("gradle", new VersionRange("7.3", "7.3", "7.3"));
+                info.invocations.add(new ArrayList<>(List.of("gradle", "build")));
```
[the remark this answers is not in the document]
stuartwdouglas: Actually it's just completely broken, as the current images hard code 'mvn'. We need to fix this once we have images that can actually do a gradle deployment.

stuartwdouglas:
I think this should be ready to go now, but some of the tests seem to hang, and I am not sure if it is related. I also see failures on main but in a different way.

Michkov (approved):

gabemontero:
On Tue, Jun 14, 2022 at 8:00 AM Stuart Douglas ***@***.***>
wrote:

> I think this should be ready to go now, but some of the tests seem to
> hang, and I am not sure if it is related. I also see failures on main but
> in a different way.
>
If you are talking about the jenkins bld srvc e2e s Stuart ignore them and
read the status I put in my Jira (maybe PLNSRVCE-363 but not sure off the
top of my head)

PR #63, opened 2022-06-02
move off non-caching client

gabemontero:


gabemontero:
this is just the dependencybuild changes @stuartwdouglas 
but I wanted to
- get CI run in to compare with my test run against my openshit cluster
- add a quick sanity check to the one we did last night

I'll be starting on artifactrequests here in a sec

To reiterate what I mentioned in slack, same situation with the caching client and lists of taskruns right after they are created ... only an 8 /47 ratio on successful hits

That said, I think this approach is better than using the non-caching client per our prior discussions.

gabemontero:
oh, and I left my list debug in intially so you see it in commit https://github.com/redhat-appstudio/jvm-build-service/pull/63/commits/a92d90f17b7eb866cdafb635f7ced57d53ff5de8 but will be removing it in a subsequent commit here shortly

pkg/reconciler/dependencybuild/controller.go
```diff
@@ -32,6 +36,25 @@ func SetupNewReconcilerWithManager(mgr ctrl.Manager, nonCachingClient client.Cli
...
 		})).
-		Watches(&source.Kind{Type: &v1beta1.TaskRun{}}, &handler.EnqueueRequestForOwner{OwnerType: &v1alpha1.DependencyBuild{}, IsController: false}).
+		Watches(&source.Kind{Type: &v1beta1.TaskRun{}}, handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+			taskRun := o.(*v1beta1.TaskRun)
+
+			// check if the TaskRun is related to DependencyBuild
+			if taskRun.GetLabels() == nil {
+				return []reconcile.Request{}
+			}
+			_, ok := taskRun.GetLabels()[artifactbuild.TaskRunLabel]
```
gabemontero, line 46: looking at the artifactbuild usage now @stuartwdouglas, should we filter out taskruns that have the taskrunlabel AND the artifactbuildidlabel (as those would be intended for the artifactbuild reconciler?

that said, I at least did not observe overt issues with the code like this

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -166,61 +195,66 @@ func (r *ReconcileDependencyBuild) handleStateSubmitBuild(ctx context.Context, d
...
+		if ownerRefs == nil || len(ownerRefs) == 0 {
+			msg := "taskrun missing onwerrefs %s:%s"
+			r.eventRecorder.Eventf(tr, v1.EventTypeWarning, msg, tr.Namespace, tr.Name)
+			log.Info(msg, tr.Namespace, tr.Name)
+			return reconcile.Result{}, nil
 		}
-	}
-	if pr.Status.CompletionTime != nil {
-		if pr.Name == db.Status.LastCompletedBuildTaskRun {
+		if len(ownerRefs) > 1 {
```
gabemontero, line 219: looking at the artifactbuild flow now, I think I now see that a taskrun could have both a db and abr owner ref

assuming that can occur at the same time for a given taskrun, I need to distinguish based on ownerref kind above here @stuartwdouglas , right?

that said, as I said before, whatever negative effect this might have on the test run, I was unable to discern it (though that most likely is my not having sufficient understanding of how to analyze the results

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -166,61 +195,66 @@ func (r *ReconcileDependencyBuild) handleStateSubmitBuild(ctx context.Context, d
...
+			msg := "taskrun %s:%s has %s ownerrefs but only using the first"
+			r.eventRecorder.Eventf(tr, v1.EventTypeWarning, msg, tr.Namespace, tr.Name, count)
+			log.Info(msg, tr.Namespace, tr.Name, count)
+		}
+		ownerRef := ownerRefs[0]
+		key := types.NamespacedName{Namespace: tr.Namespace, Name: ownerRef.Name}
+		db := v1alpha1.DependencyBuild{}
+		err := r.client.Get(ctx, key, &db)
+		if err != nil {
+			msg := "get for taskrun %s:%s owning db %s:%s yielded error %s (GGM DO WE STILL NEED RAW CLIENT!?!?!?)"
```
gabemontero, line 231: realized my prior analysis may have not distinguished from an issue here

I've reset my test system though, so I left this here, will add something similar in the artifactbuild flow, and scan again next time

gabemontero:
ok @stuartwdouglas this pr has now removed the non caching client for both dependencybuilds and artifactbuilds

my test run with the service registry looked good, and came up like it has previously and like we discussed last night

removed my temp "GGM" debug message, as I confirmed we did not hit them during the run

I will wait before hitting the merge button until you've either had a chance to look at it and are OK, of if you say you are too busy with the conference and just go for it

I will point out change you may want to visit.  My change from using GenerateName to using the dependency ID hash for the dependencybuild name.  https://github.com/redhat-appstudio/jvm-build-service/pull/63/files#diff-e8e0091cd27c2b3a10a3f5ec14aa8c07390ca462d30847514323d45e2470ff6eR295-R296

I've convinced myself this is OK.  But if you disagree, let's discuss.

thanks

stuartwdouglas:
This LGTM, but I am not sure why CI is now failing

gabemontero:
On Sat, Jun 4, 2022 at 2:29 AM Stuart Douglas ***@***.***>
wrote:

> This LGTM, but I am not sure why CI is now failing
>
ok cool on the lgty and I think I will have some time Sunday to take a look
at CI

gabemontero:
Yeah I have no idea what the `sast-go` check is, but they are failing in all the jobs with

```
sanity-inspect-image test FAIL
Unable to inspect base image - quay.io/redhat-appstudio/pull-request-builds@sha256:3ba210dda1196a7c1704eaf9e5c447f1d5a7905992c3cf3cc3a70b28b33e31bf
```
even with images unrelated to my change.

all the other tests for each job passed

would seem unrelated to my change.

/retest

gabemontero:
OK this time to job actually produced logs when I tailed it live.  It is running `gosec` https://github.com/securego/gosec 

Now, in the job, it redirects to a file and then greps it.

I ran locally and viewed the logs in detail.  It is complaining about a few things, most notably the use of hash and that it is a weak crypto algorithm.

Sometime must have changed on the staging cluster or something and the pipeline as code stuff from the .tekton directly that Michal wired in a week or so ago is now complaining. This seems like a bit of overkill to me for our use here.  We are only hashing to get a useable string.

I'll have to ask in stand up what is going on, can we turn this off / exempt ourselves, or do we have to in facdt deal with these pedantic complaints.

gabemontero:
Hmm ... I looked in the build-service repo, and with merged PRs there, the `sast-go` job is claimed successful even though you look at its logs and you see

```
-------------
sanity-inspect-image test FAIL
Unable to inspect base image - quay.io/redhat-appstudio/pull-request-builds@sha256:0e23d095dc128a1f66eebb94d8a62fa31eff213f985a8cd5bf4e069f8cd85a4b
-------------
```

However, running `gosec -no-fail -fmt=sarif -out=gosec_output.json ./...` in the build-service repo does not produce any errors like in jvm-build-service repo.

So yeah I'll find out the deal here during Monday stand up.  If the `gosec` errors have to be addressed @stuartwdouglas , since they are independent of this PR, I may still merge this, and then work on addressing them in a subsequent PR.

PR #65, opened 2022-06-06
PLNSRVCE-363: add unit and e2etest for golang github actions for PRs

gabemontero:
note - the envtests work a bit different in CI for jvm-build-service compared to build-service because we include the vendor tree in-repo a la openshift


gabemontero:
@stuartwdouglas FYI

feel free to merge if you have a chance otherwise I'll do it in the morning

PR #69, opened 2022-06-09
PLNSRVCE-363: preliminary e2e-test integration (will get verification once openshift/release is updated to consume)

gabemontero:
the true verification of this will be once I update openshift/release for jvm-build-service a la https://github.com/openshift/release/blob/33aac19fbdcee42298798dc0f69d8873e55e50c2/ci-operator/config/redhat-appstudio/application-service/redhat-appstudio-application-service-main.yaml#L1-L3 and https://github.com/openshift/release/blob/33aac19fbdcee42298798dc0f69d8873e55e50c2/ci-operator/config/redhat-appstudio/application-service/redhat-appstudio-application-service-main.yaml#L21-L60

so quite possibly if the rehearsals in that upcoming openshift/release PR fail, I'll need to come back here and make updates

But minimally, this PR will verify
- that the initial openshift/release bootstrapping done via https://github.com/openshift/release/pull/29191 produced the consequences we want
- that these change do not affect the existing CI

FYI
@stuartwdouglas 
@mmorhun 
@psturc 
@Michkov 

Of course, if anybody in Europe is still online and sees something very obvious before I merge this, let me know and I'll update :-)

### DIFF
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -1,0 +1,22 @@
+# Build Service CI documentation
+
+Currently in build-service all tests are running in [Openshift CI](https://prow.ci.openshift.org/?job=*jvm-build*service*).
+
+## Openshift CI
+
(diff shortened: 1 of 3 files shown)


gabemontero:
ok nothing broke so merging

PR #83, opened 2022-06-17
maven-component-build: reduce number of steps, set limits

Michkov:
Stage cluster has quota set to 20 CPU per namespace. Each step is
container which is taking quota, reducing number of steps and also
setting limits for non-building steps.

### DIFF
--- a/deploy/base/run-maven-component-build-v0.1.yaml
+++ b/deploy/base/run-maven-component-build-v0.1.yaml
@@ -84,22 +84,33 @@ spec:
   steps:
     - name: git-clone
       image: "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.21.0"
+      resources:
+        requests:
+          memory: "128Mi"
+          cpu: "10m"
+        limits:
+          memory: "512Mi"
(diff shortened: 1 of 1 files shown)


deploy/base/run-maven-component-build-v0.1.yaml
```diff
@@ -84,22 +84,33 @@ spec:
   steps:
     - name: git-clone
       image: "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.21.0"
+      resources:
+        requests:
+          memory: "128Mi"
+          cpu: "10m"
+        limits:
+          memory: "512Mi"
```
stuartwdouglas, line 92: Pretty sure this will fail for Quarkus. I will follow up and make this be parameterized so it can be overridden for large builds.

stuartwdouglas (approved):

deploy/base/run-maven-component-build-v0.1.yaml
Michkov: this on git-clone, mvn-goals is not modified

PR #85, opened 2022-06-24
Update dockerfile to use redhat ubi image

mbharatk:


Michkov:
Also change `FROM gcr.io/distroless/static:nonroot` to `ubi8/ubi-minimal:8.6-751`

PR #92, opened 2022-07-11
PLNSRVCE-452: get run sample build running on openshift using jvm-build-service deployed by appstudio

gabemontero:
Basically, testing against minikube/kind, or against openshift but using the `default` namespace, bypasses all of openshift's basic SCC related pod level security checks.

After this merges, I'll work with @psturc to see if and how we might want or not want to incorporate running our non-component based, example cache population flow (essentially a pre-cursor to doing `java-builder` appstudio builds that leverage the our cache) out of https://github.com/redhat-appstudio/e2e-tests, of we want to contain that element of things to our e2e specific logic in this repo.

I also updated the `make dev` flow to become opinionated on using the `jvm-build-service` namespace to make it consistent enough with appstudio deploying jvm-build-service that we can still share yaml.

/assign @stuartwdouglas

### DIFF
--- a/deploy/base/maven-v0.2.yaml
+++ b/deploy/base/maven-v0.2.yaml
@@ -89,6 +89,8 @@ spec:
   steps:
     - name: mvn-settings
       image: registry.access.redhat.com/ubi8/ubi-minimal:8.2
+      securityContext:
+        runAsUser: 0
       script: |
         #!/usr/bin/env bash
         chown 1001:1001 -R $(workspaces.source.path)
@@ -160,6 +162,8 @@ spec:
 
     - name: mvn-goals
       image: $(params.MAVEN_IMAGE)
+      securityContext:
+        runAsUser: 0
       workingDir: $(workspaces.source.path)/$(params.CONTEXT_DIR)
       script: |
         #!/usr/bin/env bash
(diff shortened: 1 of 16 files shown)


deploy/base/maven-v0.2.yaml
```diff
@@ -174,7 +178,9 @@ spec:
         /usr/bin/mvn -s "$(workspaces.maven-settings.path)/settings.xml" -DskipTests clean package
 
     - name: analyse-dependencies
-      image: hacbs-jvm-dependency-analyser
+      securityContext:
+        runAsUser: 0
+      image: quay.io/redhat-appstudio/hacbs-jvm-dependency-analyser:3dcd8603061c7ab2cedfe128c80cac9726721a74
```
gabemontero, line 183: presumably need to incorporate QUAY_USERNAME, right?

deploy/base/maven-v0.2.yaml
```diff
@@ -185,11 +191,13 @@ spec:
         - -u
         - "$(results.untrusted-dependencies.path)"
   sidecars:
-    - image: hacbs-jvm-sidecar
+    - image: quay.io/redhat-appstudio/hacbs-jvm-sidecar:3dcd8603061c7ab2cedfe128c80cac9726721a74
```
gabemontero, line 194: presumably need to incorporate QUAY_USERNAME, right?

deploy/base/maven-v0.2.yaml
gabemontero: I sorted it out with a change to development.sh and leaving this file as is

deploy/base/maven-v0.2.yaml
gabemontero: I sorted it out with a change to development.sh and leaving this file as is

gabemontero:
so I think my move in the operator yaml to the `jvm-build-service` namespace may have broke some of the assumption in the envtest/e2etest around the `default` namespace.

I can dive into that on Tuesday.

gabemontero:
> so I think my move in the operator yaml to the `jvm-build-service` namespace may have broke some of the assumption in the envtest/e2etest around the `default` namespace.
> 
> I can dive into that on Tuesday.

main branch fails for me the same way

going to create a dummy PR off of main to confirm whether it is consistent or not

@stuartwdouglas FYI

stuartwdouglas (approved):

PR #93, opened 2022-07-12
Adding a container registry storage option

phillip-kruger:
This PR ad a Container registry as a storage option. This to discuss:

- Naming is hard, I am not sure about `OCI_REGISTRY`, maybe more general `CONTAINER_REGISTRY` ?
- Current layer 3 downloads contains all artifacts for a project, even if only one is asked for. So this extracts into tmp folder
- (1) Should this be temp or should we configure this 
- (2) Should we do this at all, or just re-download ? 
- (3) We keep a map of hash to know if this is already downloaded, this can be changed to rather create a folder structure in the temp that use the hash in the folder name. This might be safer.
- Do we need to add support for other image types like `application/vnd.docker.distribution.manifest.v2+json` ?
- This PR still contains Tests (Disabled) against Quay.io and Docker Hub and a locally started registry, we probably want to remove those. There are a test that setup a test registry in test containers and test against this, so this is probably enough


java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
```diff
@@ -0,0 +1,229 @@
...
+
+import io.quarkus.logging.Log;
+
+public class OCIRegistryRepositoryClient implements RepositoryClient {
+
+    private final String registry;
+    private final String owner;
+    private final boolean enableHttpAndInsecureFailover;
+
+    // TODO: We can also use the digestHash as the temp root folder then we do not need this ?
```
stuartwdouglas, line 48: digest hash should be used as the folder name IMHO.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
```diff
@@ -0,0 +1,229 @@
...
+    private Path getLocalCachePath(RegistryClient registryClient, ManifestTemplate manifest, String digestHash)
+            throws IOException {
+        if (imageCache.containsKey(digestHash)) {
+            return pullStale(registryClient, manifest, digestHash);
+        } else {
+            return pullFreshAndCache(registryClient, manifest, digestHash);
+        }
+    }
+
+    private Path pullStale(RegistryClient registryClient, ManifestTemplate manifest, String digestHash) throws IOException {
```
stuartwdouglas, line 139: Minor nit, but this is not really 'state', it is just already existing.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
```diff
@@ -0,0 +1,229 @@
...
+            List<BuildableManifestTemplate.ContentDescriptorTemplate> layers = ((OciManifestTemplate) manifest).getLayers();
+
+            // Layer 2 is artifacts
+            BuildableManifestTemplate.ContentDescriptorTemplate artifactsLayer = layers.get(2);
+
+            Blob blob = registryClient.pullBlob(artifactsLayer.getDigest(), s -> {
+            }, s -> {
+            });
+
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
```
stuartwdouglas, line 164: This should stream directly to disk. It would be really easy to run out of memory otherwise.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
```diff
@@ -0,0 +1,229 @@
...
+            blob.writeTo(out);
+
+            // Extract to temp (TODO: Configure this ?)
+            String osTmpDir = System.getProperty(SYS_PROP_TEMP_DIR);
+            Path tempFullPath = Paths.get(osTmpDir, HACBS);
+            Path outputPath = Files.createDirectories(tempFullPath);
+            extractTarArchive(out.toByteArray(), outputPath.toString());
+            Path repoRoot = Paths.get(outputPath.toString(), ARTIFACTS);
```
stuartwdouglas, line 172: Shouldn't this have the hash in the path? Otherwise they are all getting mixed into the same directory?

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
```diff
@@ -0,0 +1,229 @@
...
+
+            Blob blob = registryClient.pullBlob(artifactsLayer.getDigest(), s -> {
+            }, s -> {
+            });
+
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            blob.writeTo(out);
+
+            // Extract to temp (TODO: Configure this ?)
+            String osTmpDir = System.getProperty(SYS_PROP_TEMP_DIR);
```
stuartwdouglas, line 168: This should all be downloaded to the cache directory (in specific sup directories), as we don't really know how big tmp will be.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
```diff
@@ -0,0 +1,229 @@
...
+    private Path pullFreshAndCache(RegistryClient registryClient, ManifestTemplate manifest, String digestHash)
+            throws IOException {
+
+        String manifestMediaType = manifest.getManifestMediaType();
+
+        if (OCI_MEDIA_TYPE.equalsIgnoreCase(manifestMediaType)) {
+            List<BuildableManifestTemplate.ContentDescriptorTemplate> layers = ((OciManifestTemplate) manifest).getLayers();
+
+            // Layer 2 is artifacts
+            BuildableManifestTemplate.ContentDescriptorTemplate artifactsLayer = layers.get(2);
```
stuartwdouglas, line 158: We probably need some error handling here.

stuartwdouglas (changes requested):

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
phillip-kruger: Not really as the version is in the path, but now the hash will be in the path anyway

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
phillip-kruger: Meaning ? If there is not a layer 2 ?

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
stuartwdouglas: What do you mean? The path is just  osTmpDir + HACBS ?

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
stuartwdouglas: yea, if the image is not valid it should probably log an error and return.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
phillip-kruger: Yea but then follow it a mvn like structure that contains the version.

stuartwdouglas: This does allow images to potentially overwrite each other though. These really need to be in a directory that matches the hash. You can also then use the existence of this directory to know if the layer is present or not.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
phillip-kruger: Cool. Adding that.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
phillip-kruger: Yes, that is what I am doing not. So removing the map that holds the cache and just look at local cache

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
phillip-kruger: So this is the tar (that I keep in memory) and then extract that to disk. Do you think I should store the tar too ?

phillip-kruger:
@stuartwdouglas - as discussed, I removed the test data and now generate it in the test

stuartwdouglas (approved):

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java
```diff
@@ -127,53 +136,46 @@ private RegistryClient getRegistryClient(String registry, String owner, String g
...
-            Path outputPath = Files.createDirectories(tempFullPath);
-            extractTarArchive(out.toByteArray(), outputPath.toString());
-            Path repoRoot = Paths.get(outputPath.toString(), ARTIFACTS);
-
-            imageCache.put(digestHash, repoRoot);
-
-            return repoRoot;
+                Path tarFile = Files.createFile(Paths.get(outputPath.toString(), digestHash + ".tar"));
+                try (OutputStream tarOutputStream = Files.newOutputStream(tarFile)) {
+                    blob.writeTo(Files.newOutputStream(tarFile));
```
gabemontero, line 168: easy enough / cool :-)

stuartwdouglas: oops, I missed that in review, @phillip-kruger shouldn't think be using the tarOutputStream ?

PR #97, opened 2022-07-12
Fix local dev flow

stuartwdouglas:


deploy/development.sh
```diff
@@ -8,8 +8,6 @@ kubectl delete deployments.apps localstack -n jvm-build-service
 DIR=`dirname $0`
 kubectl apply -f $DIR/namespace.yaml
 kubectl config set-context --current --namespace=jvm-build-service
-find $DIR -name maven-v0.2.yaml -exec sed -i s/redhat-appstudio/${QUAY_USERNAME}/ {} \;
```
stuartwdouglas, line 11: @gabemontero why did you add this? It should already be replaced by an overlay. This is modifying the version controlled files, which means it is likely the changes will be accidentally commited.

gabemontero: yeah this was meant to be a placeholder while I vetted things today ... I just didn't put comments to that effect before I signed off last night

deploy/development.sh
```diff
@@ -8,8 +8,6 @@ kubectl delete deployments.apps localstack -n jvm-build-service
...
-find $DIR -name maven-v0.2.yaml -exec sed -i s/3dcd8603061c7ab2cedfe128c80cac9726721a74/dev/ {} \;
```
gabemontero, line 12: that said, I'm still not clear how the specific sha tag will get changed to `dev` instead

I'll check out your PR today and discover one way or the other

gabemontero: ok I finally found the `kustomization.yaml` file that was changing this ... not sure why I was having such a hard time finding this the other day

PR #99, opened 2022-07-13
Revert move to pipeline runs

stuartwdouglas:


stuartwdouglas:
@gabemontero I am just going to revert this for now, and we can re-do it once we have discussed the implications some more.

PR #100, opened 2022-07-13
Fix build discovery

stuartwdouglas:


deploy/development.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
 
 kubectl delete deployments.apps hacbs-jvm-operator -n jvm-build-service
-kubectl delete deployments.apps hacbs-jvm-cache -n jvm-build-service
-kubectl delete deployments.apps localstack -n jvm-build-service
+#kubectl delete deployments.apps hacbs-jvm-cache -n jvm-build-service
```
gabemontero, line 4: can you clarify via comment why we do not want to recycle these, so in a few weeks if I'm back in this script, I don't inadvertently add these deletes back

gabemontero (commented):
just an "aside"  comment and a minor ask for a comment clarification

admittedly, I still need to find time at some point and do a end to end review of the various java based components if I'm going to be able and provide any comments of any nuance 

does feel like the operator ehancements are becoming more boilerplate and we've reached a good steady state wrt best practices, etc. ..... the number of states do keep growing.  I wonder if that is simply OK, or is indicative of the need to perhaps break up the CRDs some.

deploy/base/lookup-build-info-v0.1.yaml
```diff
@@ -0,0 +1,65 @@
+apiVersion: tekton.dev/v1beta1
+kind: ClusterTask
```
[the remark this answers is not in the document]
stuartwdouglas: I know, this was why I decided to convert to pipelines with my other PR, then realized it was premature. It's really not that much work to change, but we do need to figure out how it will work with regards to namespaces/workspaces.

deploy/development.sh
stuartwdouglas: Done. Basically this is only really necessary if you are actively working on the cache, otherwise it just slows down the development flow as builds end up taking longer.

PR #102, opened 2022-07-14
PLNSRVCE-446: switch to tekton object based pruning in the operator

gabemontero:


stuartwdouglas (approved):

stuartwdouglas:
We really need to change this to preserve the results somehow. I am seeing build failures on the staging cluster that are not failing locally, but this means the logs have been deleted.

gabemontero:
> We really need to change this to preserve the results somehow. I am seeing build failures on the staging cluster that are not failing locally, but this means the logs have been deleted.

as @stuartwdouglas and I discussed in slack, @adambkaplan 's results persistence efforts in upstream tekon are possibly the long term solution; but in the interim, a cabal like discussion with minimally @sbose78 @Michkov and us is needed to reconcile the desire to preserve logs for debug (at least for some amount of time if not permanently) vs. cleaning up namespaces to avoid violating quota.

PR #103, opened 2022-07-15
PLNSRVCE-482: Allow the overriding of the analyzer image in CI runs

brunoapimentel:
We need to override the used dependency analyzer image in order to properly run the CI tests. This PR provides a mechanism that achieves that by:

-  cloning the build-definitions repo
-  overriding the image reference in the task where its being used
-  creating a new Tekton bundle
-  pushing it to a target repository
-  creating a configmap to override the bundle used by Appstudio
-  starting the e2e tests

### DIFF
--- a/.ci/override-tekton-bundle.sh
+++ b/.ci/override-tekton-bundle.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+# This script will generate a new Tekton bundle that allows the overriding of images that are hard-coded into task steps
+TEMP_FOLDER=$WORKSPACE/tmp/bundle-override
+APPSTUDIO_QE_REPO=quay.io/redhat-appstudio-qe/test-images
+TASK_BUNDLE_IMG=$APPSTUDIO_QE_REPO:task-bundle-$JVM_BUILD_SERVICE_PR_SHA
+PIPELINE_BUNDLE_IMG=$APPSTUDIO_QE_REPO:pipeline-bundle-$JVM_BUILD_SERVICE_PR_SHA
+
+function getCurrentBuildBundle() {
+    curl -s -o "$TEMP_FOLDER"/build-kustomization.yaml https://raw.githubusercontent.com/redhat-appstudio/infra-deployments/main/components/build/kustomization.yaml
+
+    BUILD_BUNDLE=$(yq e ".configMapGenerator[].literals[]" \
+    "$TEMP_FOLDER"/build-kustomization.yaml \
+    | grep default_build_bundle \
+    | sed "s/default_build_bundle=//")
+
+    echo "New build bundle will be created based on $BUILD_BUNDLE"
+}
+
+function createPipelinesFile() {
+    pipelines=$(tkn bundle list "$BUILD_BUNDLE" | sed "s/pipeline.tekton.dev\///")
+    for pipeline in $pipelines; do
+        echo "---" >> "$TEMP_FOLDER"/pipelines.yaml
+        tkn bundle list -o yaml "$BUILD_BUNDLE" pipeline "$pipeline" >> "$TEMP_FOLDER"/pipelines.yaml
+    done
+}
+
+function createTaskFile() {
+    for bundle in $(grep bundle "$TEMP_FOLDER"/pipelines.yaml | sort -u | awk '{print $2}'); do 
+        tasks=$(tkn bundle list "$bundle" | sed "s/task.tekton.dev\///")
+
+        for task in $tasks; do
+            if [ "$task" == "s2i-java" ]; then
+                tkn bundle list -o yaml "$bundle" task "$task" >> "$TEMP_FOLDER"/task.yaml
+            fi
+        done
+    done
+}
+
+function updateAnalyzerImage() {
+    echo "jvm-build-service analyzer image set to $JVM_BUILD_SERVICE_ANALYZER_IMAGE"
+    yq e -i "select(.metadata.name == \"s2i-java\") \
+    | (.spec.steps[] | select(.name == \"analyse-dependencies-java-sbom\").image) \
+    |= \"$JVM_BUILD_SERVICE_ANALYZER_IMAGE\"" "$TEMP_FOLDER"/task.yaml
+}
+
+function updatePipelineRef() {
+    yq e -i "(.spec.tasks[].taskRef | select (.name == \"s2i-java\").bundle) \
+    |= \"$TASK_BUNDLE_IMG\"" "$TEMP_FOLDER"/pipelines.yaml
+}
+
+function createDockerConfig() {
+    mkdir ~/.docker
+    echo "$QUAY_TOKEN" | base64 -d > ~/.docker/config.json
+}
+
+function createAndPushTaskBundle() {
+    echo "Creating $TASK_BUNDLE_IMG"
+    tkn bundle push "$TASK_BUNDLE_IMG" -f "$TEMP_FOLDER"/task.yaml
+}
+
+function createAndPushPipelineBundle() {
+    echo "Creating $PIPELINE_BUNDLE_IMG"
+    tkn bundle push "$PIPELINE_BUNDLE_IMG" -f "$TEMP_FOLDER"/pipelines.yaml
+}
+
+mkdir -p "$TEMP_FOLDER"
+
+getCurrentBuildBundle
+createPipelinesFile
+createTaskFile
+updateAnalyzerImage
+updatePipelineRef
+createDockerConfig
+createAndPushTaskBundle
+createAndPushPipelineBundle
+
+export DEFAULT_BUILD_BUNDLE="$PIPELINE_BUNDLE_IMG"
--- a/.ci/common-ci-settings-and-functions.sh
+++ b/.ci/common-ci-settings-and-functions.sh
@@ -86,6 +86,8 @@ function waitBuildToBeReady() {
     done
 }
 
+. .ci/override-tekton-bundle.sh
+
 curl https://raw.githubusercontent.com/redhat-appstudio/e2e-tests/main/scripts/install-appstudio-e2e-mode.sh | bash -s install
 
 export -f waitAppStudioToBeReady


.ci/update-tekton-bundle.sh
```diff
@@ -0,0 +1,79 @@
...
+function createAndPushPipelineBundles() {
+    tkn bundle push $PIPELINE_BUNDLE_IMG -f ${PIPELINE_TEMP}/base.yaml
+}
+
+function createConfigMap() {
+    oc apply -m build-templates -f - << CONFIGMAP
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: build-pipelines-defaults
```
gabemontero, line 69: I think you'll need to specify a well known namespace that is accessible like `default` or `openshift`, or perhaps create a new one, so that when the e2e's follow their typical pattern of creating a test namespace, our jvm build service e2e code can find this config map and copy it into the test namespace.

.ci/update-tekton-bundle.sh
```diff
@@ -0,0 +1,79 @@
...
+data:
+  default_build_bundle: "$PIPELINE_BUNDLE_IMG"
```
gabemontero, line 71: so this overrides the pipeline bundle

don't we need a similar entry for the task bundle?

stuartwdouglas: Won't the rebuilt build bundle reference the rebuilt task bundles that were created in the build process?

gabemontero: maybe?

can you post some sample output that clarifies one way or the other @brunoapimentel ?  or maybe add a "debug commit" that runs `cat` on the relevant updated yaml prior to calling `tkn` ?

gabemontero:
we will retest after https://github.com/redhat-appstudio/jvm-build-service/pull/107 merges

.ci/update-tekton-bundle.sh
brunoapimentel: In this script, we're updating all the task refs to point to the newly built ones: https://github.com/brunoapimentel/jvm-build-service/blob/ci-override-analyzer-image/.ci/update-tekton-bundle.sh#L49-L53

.ci/update-tekton-bundle.sh
brunoapimentel: I created it in this namespace because it would override all pipeline runs from all components (that is, unless a specific component had a configmap like that created in the same namespace it lives, then it would take precedence over the one in build-pipelines-defaults).

But if you think it's best to have it copied to a specific namespace as needed, we can go with that approach as well. :)

gabemontero: ah I was unaware of had forgotten that element of the config map override features you had added @brunoapimentel thanks for the clarification.  I'm fine with the global override.

That said, you have a typo in the line above.  I'll add a suggestion comment

.ci/update-tekton-bundle.sh
```diff
@@ -0,0 +1,79 @@
...
+    echo Creating $APPSTUDIO_TASKS_REPO:task-bundle-$JVM_BUILD_SERVICE_PR_SHA-$PART
+    tkn bundle push $TASKS $REF
+}
+
+function createAndPushPipelineBundles() {
+    tkn bundle push $PIPELINE_BUNDLE_IMG -f ${PIPELINE_TEMP}/base.yaml
+}
+
+function createConfigMap() {
+    oc apply -m build-templates -f - << CONFIGMAP
```
gabemontero, line 65:
```suggestion
    oc apply -n build-templates -f - << CONFIGMAP
```

.ci/update-tekton-bundle.sh
gabemontero: ok that helps @brunoapimentel thanks

just to make sure I'm understanding the bundle structure, is it correct to say that while we have 3 bundles, the pipeline bundles is the "top level bundle" with refs to the task bundle and conceivably the util bundle as well?

gabemontero:
the images error is interesting @brunoapimentel 
```
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
error: build error: error building at STEP "RUN wget https://github.com/mikefarah/yq/releases/download/v4.25.1/yq_linux_amd64 -O /usr/local/bin/yq &&     chmod +x /usr/local/bin/yq && yq --version &&     curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl &&     chmod +x ./kubectl &&     mv ./kubectl /usr/local/bin &&     curl -LO https://github.com/tektoncd/cli/releases/download/v0.24.0/tkn_0.24.0_Linux_x86_64.tar.gz &&     tar xvzf tkn_0.24.0_Linux_x86_64.tar.gz --no-same-owner -C /usr/local/bin/ tkn": error while running runtime: exit status 2
```

I wonder if something has changed with the `registry.ci.openshift.org/openshift/release:golang-1.17` used as the base when your PR running https://github.com/redhat-appstudio/jvm-build-service/blob/main/.ci/openshift-ci/Dockerfile#L16 presumably succeeded.

My best guess if the file permissions has changed and the curl is not actually storing on disk, or we can no longer assume permission writing to `/usr/local/bin`.  Or there was a networking hiccup.

Trying again in case it was a networking hiccup.

/test images

gabemontero:
it does look like there were some github.com throttling hiccups yesterday with RH repos ... could have been that

gabemontero:
ok image building again

/test jvm-build-service-e2e

.ci/update-tekton-bundle.sh
```diff
@@ -0,0 +1,87 @@
...
+
+function cloneBuildDefinitions() {
+    git clone https://$GITHUB_TOKEN@github.com/redhat-appstudio/build-definitions.git "$BUILD_DEFINITIONS_DIR"
+}
+
+function updateAnalyzerImage() {
+    yq e -i "(.[] \
+    | select(.path == \"/spec/steps/-\").value \
+    | select(.name == \"analyse-dependencies-java-sbom\").image) \
+    |= \"$JVM_BUILD_SERVICE_ANALYZER_IMAGE\"" \
```
gabemontero, line 13: so in debugging https://github.com/openshift/release/pull/30600 I realized last night that the override for the sidecar image at https://github.com/redhat-appstudio/infra-deployments/blob/d156b6cb8ad4180c1ee5d93d68cc45450940996e/hack/preview.sh#L96-L97 is incorrect.

So we now have a bit of a race between this work and https://issues.redhat.com/browse/PLNSRVCE-455

So with 455, we want to remove the use of cluster tasks.  And @stuartwdouglas has already started with the req processor, but that works different, in that the operator just injects the needed content in the taskruns it generates vs. referencing the cluster task.

There ref in the operator to the side car cluster task pulled in via https://github.com/redhat-appstudio/infra-deployments/blob/376acd8c9b4d9dedac1788ba16eee3f76aad7394/components/build/jvm-build-service/kustomization.yaml#L6 over at https://github.com/redhat-appstudio/jvm-build-service/blob/d1b637f56c4df350447685655a373943317c84c2/pkg/reconciler/dependencybuild/dependencybuild.go#L264 as part of build recipe construction.


And we also have references in the bundles to this cluster task and the images used for it, namely https://github.com/redhat-appstudio/build-definitions/blob/main/tasks/patches/add-sbom-and-push.yaml

If we remove the cluster task ref in the operator first, where we again hard code in the task def needed, then we'll want to presumably employ the same env var override pattern @stuartwdouglas is pursuing with https://github.com/redhat-appstudio/infra-deployments/pull/506

I think that is the path we want to prioritize.

But if there are hurdles with that, and we have to continue to live with the `run-maven-component-build` for a while, then we would want to update this PR to take `JVM_BUILD_SERVICE_SIDECAR_IMAGE` and update the ClusterTasks's image as well.

Thoughts @stuartwdouglas @brunoapimentel

gabemontero:
clean run, but still getting those auth errors pushing to bundle @psturc @brunoapimentel :+1: 
`Error: could not push image to registry as "quay.io/repository/redhat-appstudio-qe/test-images:task-bundle-6595aefcf9d9a3445c1efdcc17b84d9352ae3598-1": HEAD https://quay.io/v2/repository/redhat-appstudio-qe/test-images/blobs/sha256:726f445a325fc72e713d6569f26185acee0005db89c008b51f38613539482cbe: unexpected status code 401 Unauthorized (HEAD responses have no body, use GET for details)`

Is that image URL correct?  should it not be something like `quay.io/redhat-appstudio-qe/test-images:task-bundle-6595aefcf9d9a3445c1efdcc17b84d9352ae3598-1` instead, where you've removed `/respository/` from that URL?

.ci/update-tekton-bundle.sh
gabemontero: fyi I think I have the path down for hard coding / in lining the references to https://github.com/redhat-appstudio/jvm-build-service/blob/main/deploy/base/run-maven-component-build-v0.1.yaml and https://github.com/redhat-appstudio/jvm-build-service/blob/main/deploy/base/run-gradle-component-build-v0.1.yaml ... I should have a PR up before my EOB today.

Assuming @stuartwdouglas agrees I am on the right path, we can follow the pattern he did for the req processor and not worry about updating it via this tekton bundle update approach here.

gabemontero: won't make EOB today for the https://github.com/redhat-appstudio/jvm-build-service/blob/main/deploy/base/run-gradle-component-build-v0.1.yaml replacement

and @stuartwdouglas reminded me just now in slack their are 2 sidecars and in fact we still have to do the override of the one in add-sbom @brunoapimentel  ... if you can add it to this PR after sorting out the bundle image push, great.  Otherwise, someone can create a follow up PR to build on the base you have here.

.ci/update-tekton-bundle.sh
```diff
@@ -0,0 +1,87 @@
...
+}
+
+function createAndPushPipelineBundles() {
+    echo Creating $PIPELINE_BUNDLE_IMG
+    tkn bundle push $PIPELINE_BUNDLE_IMG -f ${PIPELINE_TEMP}/base.yaml
+}
+
+function createConfigMap() {
+    echo Creating configmap to override default bundle to $PIPELINE_BUNDLE_IMG
+    oc apply -n build-templates -f - << CONFIGMAP
```
Michkov, line 72: this will be overridden by argocd

.ci/update-tekton-bundle.sh
```diff
@@ -0,0 +1,87 @@
...
+    |= \"$JVM_BUILD_SERVICE_ANALYZER_IMAGE\"" \
+    $BUILD_DEFINITIONS_DIR/tasks/patches/add-sbom-and-push.yaml
+}
+
+function createDockerConfig() {
+    mkdir ~/.docker
+    echo $QUAY_TOKEN | base64 -d > ~/.docker/config.json
+}
+
+function createAndPushTaskBundle() {
```
Michkov, line 22: I would go with way of getting current bundles, storing into file, modify and upload again:
```
tkn bundle list -o yaml quay.io/redhat-appstudio/build-templates-bundle:0e0b28716550e1adc50443084adecec67f1748ef > pipelines.yaml
for bundle in $(grep bundle pipelines.yaml  | sort -u | awk '{print $2}'); do i=`echo $bundle | sed 's/.*-//'`; tkn bundle list -o yaml $bundle > tasks-$i.yaml; done
```
Find the task, modify it, upload new bundle (could be enough to upload new bundle only with that one task). Modify pipelines.yaml -> change task bundle link for your modified task.

brunoapimentel: Any suggestions on how to get the latest tag?

Michkov: from infra-deployments repository which is available (since it's running preview script)

brunoapimentel:
> clean run, but still getting those auth errors pushing to bundle @psturc @brunoapimentel `Error: could not push image to registry as "quay.io/repository/redhat-appstudio-qe/test-images:task-bundle-6595aefcf9d9a3445c1efdcc17b84d9352ae3598-1": HEAD https://quay.io/v2/repository/redhat-appstudio-qe/test-images/blobs/sha256:726f445a325fc72e713d6569f26185acee0005db89c008b51f38613539482cbe: unexpected status code 401 Unauthorized (HEAD responses have no body, use GET for details)`
> 
> Is that image URL correct? should it not be something like `quay.io/redhat-appstudio-qe/test-images:task-bundle-6595aefcf9d9a3445c1efdcc17b84d9352ae3598-1` instead, where you've removed `/respository/` from that URL?

You're absolutely right, Gabe! Removing the `repository` part solved it :facepalm:

.ci/update-tekton-bundle.sh
psturc: That's a good point.
We could create this config map in a specific namespace that will be then used by the test, but since some tests are already using namespaces which names are random-generated (and I think jvm-build-service tests should do that too), we could leave the config map creation for the test itself (similarly like [here](https://github.com/redhat-appstudio/e2e-tests/blob/1069888eeed24803a3a904fbf0f94edfdf3cfc9f/tests/build/build.go#L489-L494)) - wdyt @gabemontero?

The bundle image value could be extracted from an env var. If the env var is empty, no config map will be created by the test.

brunoapimentel: Seems like a good approach, I can remove this function here and just leave the env variable set. I

.ci/update-tekton-bundle.sh
```diff
@@ -0,0 +1,69 @@
...
+    TARGET_BUNDLE=$(yq e ".configMapGenerator[].literals[]" \
+    $WORKSPACE/tmp/infra-deployments/components/build/kustomization.yaml \
+    | grep default_build_bundle \
+    | sed "s/default_build_bundle=//")
+
+    echo New build bundle will be created based on $TARGET_BUNDLE
+}
+
+function createPipelineFile() {
+    tkn bundle list -o yaml $TARGET_BUNDLE pipeline java-builder > $TEMP_FOLDER/pipeline.yaml
```
brunoapimentel, line 17: I only fetched a single pipeline here because when `tkn` outputs the list of pipelines, it does not create a break between them, so this affects how yq parses it.

gabemontero:
let's re-run with @psturc 's commenting out of the various tests while we wait on minimally https://github.com/redhat-appstudio/infra-deployments/pull/513

gabemontero:
ok we now are hitting the issue tracked by https://github.com/redhat-appstudio/e2e-tests/pull/152 https://github.com/redhat-appstudio/e2e-tests/pull/153

gabemontero:
From https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/103/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-e2e/1550825338118868992/artifacts/jvm-build-service-e2e/jvm-build-service-e2e/build-log.txt

```
New build bundle will be created based on quay.io/redhat-appstudio/build-templates-bundle:0e0b28716550e1adc50443084adecec67f1748ef
jvm-build-service analyzer image set to registry.build01.ci.openshift.org/ci-op-gx2kkl51/pipeline
Creating quay.io/redhat-appstudio-qe/test-images:task-bundle-af9cd7b6d2f757bec2de5cdf963b3043574aa813
Creating Tekton Bundle:
	- Added Task: s2i-java to image

Pushed Tekton Bundle to quay.io/redhat-appstudio-qe/test-images@sha256:a5400cbac91d21b88c49df02788d641daca8797872538f0020695e08c9cfd186
Creating quay.io/redhat-appstudio-qe/test-images:pipeline-bundle-af9cd7b6d2f757bec2de5cdf963b3043574aa813
Creating Tekton Bundle:
	- Added Pipeline: docker-build to image
	- Added Pipeline: java-builder to image
	- Added Pipeline: nodejs-builder to image
	- Added Pipeline: noop to image
	- Added Pipeline: prototype-compliance to image

Pushed Tekton Bundle to quay.io/redhat-appstudio-qe/test-images@sha256:23d001a24032ab760fba28d893294e8d81ae97b2c4bad9478479c8bed8ebeb3f

```

That seems to meet the criteria here.  Unless I am missing something or you have other observactions @Michkov I want to merge this to possibly facilitate https://github.com/openshift/release/pull/30600

gabemontero:
We can also move toward doing the similar update of the sidecar image in the bundle.

.ci/update-tekton-bundle.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+TEMP_FOLDER=$WORKSPACE/tmp/bundle-override
+APPSTUDIO_QE_REPO=quay.io/redhat-appstudio-qe/test-images
+TASK_BUNDLE_IMG=$APPSTUDIO_QE_REPO:task-bundle-$JVM_BUILD_SERVICE_PR_SHA
+export PIPELINE_BUNDLE_IMG=$APPSTUDIO_QE_REPO:pipeline-bundle-$JVM_BUILD_SERVICE_PR_SHA
+
+function setTargetBundle() {
```
Michkov, line 7: nit: it is rather getTargetBundle or getDefaultBundle

.ci/update-tekton-bundle.sh
```diff
@@ -0,0 +1,72 @@
...
+mkdir "$TEMP_FOLDER"
+
+setTargetBundle
+createPipelinesFile
+createTaskFile
+updateAnalyzerImage
+updatePipelineRef
+createDockerConfig
+createAndPushTaskBundle
+createAndPushPipelineBundle
```
Michkov, line 72: how is new Pipeline bundle used?

gabemontero: Short term, I do not believe it is, because of the "hard coding" we sorted out with https://coreos.slack.com/archives/C032EJ007C0/p1658212125508779

The CI flow will set `JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE` to the image built in openshift CI, and then because of https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build/jvm-build-service/operator-images.yaml#L12-L13 the controller will use that openshift CI built image when it in-lines the req analyzer task per https://github.com/redhat-appstudio/jvm-build-service/blob/main/pkg/reconciler/artifactbuild/artifactbuild.go#L461 and https://github.com/redhat-appstudio/jvm-build-service/blob/main/pkg/reconciler/dependencybuild/dependencybuild.go#L508.

But longer term, as we switch to pipeline runs, and as we make progress on https://docs.google.com/document/d/1M4o4cMMYz8lM_gd4VfJGiLE8p9FBWeeQvOxcfq9WYDE/edit# minimally (and maybe on other stories?), and presumably build-service can decipher that an incoming Component needs `jvm-build-service` integration, then it will be access the bundle that is being created here, per the configmaps based overrides that @brunoapimentel added a month or 2 ago, as it constructs the PIpelineRun needed for the build, and away we go.

So rather than trying to wedge in CI / e2e updates after the fact, yes, this PR establishes the test plumbing before we actually implement features on the horizon.

gabemontero: and yes I now (re?)realize this change is not an immediate blocker to PLNSRVCE-452

brunoapimentel: I'm not sure I follow everything you wrote there, @gabemontero. :sweat_smile: 

I understood that the e2e tests would create a configmap that points to the new bundle as needed, and that's how the pipeline would be pulled in.

We discussed it on this thread: https://github.com/redhat-appstudio/jvm-build-service/pull/103#discussion_r926277674

gabemontero: yes that still holds @brunoapimentel .... what I clarified in https://github.com/redhat-appstudio/jvm-build-service/pull/103#discussion_r928909364 is my realization on _**WHEN**_ that confirmap would actually have an effect.  When the _**WHEN**_ is after we stop hardcoding those task specs in the controller and changes them to refs that point to items in the bundle.

gabemontero: and yes the wiring of all this is crazy complicated ... I myself may still not have the connections 100% memorized and accurately "recall-able".

gabemontero:
btw https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/103/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1551585375418322944 will fail until I can debug / fix the problem I noted Friday with https://coreos.slack.com/archives/C03KN2B47GU/p1658585506731059?thread_ts=1658531778.074429&cid=C03KN2B47GU

Starting off with pulling in one by one a few of @stuartwdouglas 's fix PRs .... I think I am minimally hitting some of those.

When I have something passing again locally I will push a PR and reference it here.

gabemontero:
So https://github.com/redhat-appstudio/jvm-build-service/pull/129 moves the boulder up the hill some, but I'm either going to need help or another day or so of study to sort out why at least some of the artifactbuilds / dependencybuilds are not getting to complete state.

gabemontero:
in repo is now green with #129 

/retest

.ci/update-tekton-bundle.sh
brunoapimentel: I had a chat with @Michkov and we decided to patch the kustomization script in infra-deployments, so that the new build-bundle would be available by default to all namespaces.

Notice that, to do that, we needed to change the execution of this script to run before bootstraping the cluster. We are also changing the preview script in https://github.com/redhat-appstudio/infra-deployments/pull/526.

.ci/override-tekton-bundle.sh
```diff
@@ -0,0 +1,77 @@
...
+    echo "Creating $TASK_BUNDLE_IMG"
+    tkn bundle push "$TASK_BUNDLE_IMG" -f "$TEMP_FOLDER"/task.yaml
+}
+
+function createAndPushPipelineBundle() {
+    echo "Creating $PIPELINE_BUNDLE_IMG"
+    tkn bundle push "$PIPELINE_BUNDLE_IMG" -f "$TEMP_FOLDER"/pipelines.yaml
+}
+
+mkdir -p "$TEMP_FOLDER"
```
gabemontero, line 66: interesting, since you used `-p`, but this line if failing with 

`mkdir: cannot create directory '/go/src/github.com/redhat-appstudio/jvm-build-service/tmp/bundle-override': No such file or directory`

Maybe it is a permission issue masquerading as a no such file (i.e. the permission issue prevents you from creating `tmp` ??)

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/103/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1551861261765971968

Michkov: that `-p` was added recently

brunoapimentel: The script worked in the latest run, the bundles were pushed as expected: https://quay.io/repository/redhat-appstudio-qe/test-images/manifest/sha256:41f0ded0364e0d234d99ba3613da39ceb9ac72a2b56195ce2f59083212724053

The error was some blip regarding Github, it returned a 429 message, not sure why:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/103/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-e2e/1551867931267174400

gabemontero: hmm ok yeah looking at the time for https://github.com/redhat-appstudio/jvm-build-service/compare/45cb5b52ca7f7bb6e4708f681721ecace872e9c1..d95f519cae7121824084f105b509adaa2b546f88 vs. https://github.com/redhat-appstudio/jvm-build-service/pull/103#issuecomment-1195296295 it is feasible the test run was started before the force push

so let's see what happens with https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/103/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1551867931309117440

thanks

gabemontero:
another `tekton-chanins` ns not created in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/103/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1551867931309117440

going to try @Michkov 's test in  https://coreos.slack.com/archives/C02FTKEMBU1/p1658726783159679?thread_ts=1658499647.682389&cid=C02FTKEMBU1 and see if the unveils anything

Michkov:
@gabemontero I can see -> ` fatal: unable to access 'https://github.com/redhat-appstudio-qe/infra-deployments.git/': The requested URL returned error: 429 `, it was not able to push changes, branch did not exists so the argocd was not able to download it and apply.

gabemontero:
> @gabemontero I can see -> `fatal: unable to access 'https://github.com/redhat-appstudio-qe/infra-deployments.git/': The requested URL returned error: 429`, it was not able to push changes, branch did not exists so the argocd was not able to download it and apply.

good eye @Michkov - yep in hindsight I do not recall that message in earlier incarnations of this

and certainly @brunoapimentel 's changes here introduce the repo `redhat-appstudio-qe` though it is in relation to the bundle image ref.

Is it possible the consumption of the new bundle at https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/preview.sh#L101 leads to this ref to  'https://github.com/redhat-appstudio-qe/infra-deployments.git/' ??

gabemontero:
> > @gabemontero I can see -> `fatal: unable to access 'https://github.com/redhat-appstudio-qe/infra-deployments.git/': The requested URL returned error: 429`, it was not able to push changes, branch did not exists so the argocd was not able to download it and apply.
> 
> good eye @Michkov - yep in hindsight I do not recall that message in earlier incarnations of this
> 
> and certainly @brunoapimentel 's changes here introduce the repo `redhat-appstudio-qe` though it is in relation to the bundle image ref.
> 
> Is it possible the consumption of the new bundle at https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/preview.sh#L101 leads to this ref to 'https://github.com/redhat-appstudio-qe/infra-deployments.git/' ??

oh it was an intermittent thing ... from the succeeding run:

```
Each component above is set to the following repositories
if you do not see your component in the list, please send a PR update to argo-cd-apps/overlays/development/repo-overlay.yaml
https://github.com/redhat-appstudio-qe/infra-deployments.git
---
https://github.com/redhat-appstudio-qe/infra-deployments.git
---
https://github.com/redhat-appstudio-qe/infra-deployments.git
---
https://github.com/redhat-appstudio-qe/infra-deployments.git
---
https://github.com/redhat-appstudio-qe/infra-deployments.git
---
https://github.com/redhat-appstudio-qe/infra-deployments.git
---
https://github.com/redhat-appstudio-qe/infra-deployments.git
---
https://github.com/redhat-appstudio-qe/infra-deployments.git
---
https://github.com/redhat-appstudio-qe/infra-deployments.git
---
https://github.com/redhat-appstudio-qe/infra-deployments.git
Changing AppStudio Gitlab Org to "redhat-appstudio-qe"
```

PR #105, opened 2022-07-18
  PLNSRVCE-547 Ability to configure JVM builds 

stuartwdouglas:
The presence of the "jvm-build-config" ConfigMap triggers the
environment setup. This config map is used to configure user related
settings in the namespace.

In addition there is the jvm-build-service/jvm-build-system-config that
configures system wide config such as builder images and other system
settings.

Presently this will:
- Create a new cache instance in the local workspace if the `jvm-build-config` configmap is present in the namespace
- Configure this cache from the map
- Configure system wide images from a system config map
- Configure builder images from a system config map

This also resolves the following:

[PLNSRVCE-465](https://issues.redhat.com//browse/PLNSRVCE-465) Java Workspace Setup
[PLNSRVCE-547](https://issues.redhat.com//browse/PLNSRVCE-547) Allow users to configure per namespace recipe repositories

Infra deployments PR: https://github.com/redhat-appstudio/infra-deployments/pull/537
Build Definitions PR: https://github.com/redhat-appstudio/build-definitions/pull/179

### DIFF
--- a/.ci/common-ci-settings-and-functions.sh
+++ b/.ci/common-ci-settings-and-functions.sh
@@ -23,6 +23,9 @@ if [ "$PR_RUN" != "notpr" ]; then
   # JVM_BUILD_SERVICE_SIDECAR_IMAGE - sidecar image built in openshift CI job workflow.
   # JVM_BUILD_SERVICE_ANALYZER_IMAGE - dependency analyzer image built in openshift CI workflow.
   # JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE - request processor image built in openshift CI workflow.
+  # JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE - JDK build image built in openshift CI workflow.
+  # JVM_BUILD_SERVICE_JDK11_BUILDER_IMAGE - JDK build image built in openshift CI workflow.
+  # JVM_BUILD_SERVICE_JDK17_BUILDER_IMAGE - JDK build image built in openshift CI workflow.
 
   # More info about how image dependencies work in ci: https://github.com/openshift/ci-tools/blob/master/TEMPLATES.md#parameters-available-to-templates
   # Container env defined at: https://github.com/openshift/release/blob/master/ci-operator/config/redhat-appstudio/jvm-build-service/redhat-appstudio-jvm-build-service-main.yaml
@@ -55,6 +58,18 @@ if [ "$PR_RUN" != "notpr" ]; then
   export JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE=${JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE%@*}
   export JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE_REPO=${JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE:-"quay.io/redhat-appstudio/hacbs-jvm-build-request-processor"}
   export JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE_TAG=${JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE_TAG:-"redhat-appstudio-jvm-build-service-reqprocessor-image"}
+
+  export JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE=${JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE%@*}
+  export JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE_REPO=${JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE:-"quay.io/redhat-appstudio/hacbs-jdk8-builder"}
+  export JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE_TAG=${JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE_TAG:-"redhat-appstudio-jvm-build-service-jdk8-builder-image"}
+
+  export JVM_BUILD_SERVICE_JDK11_BUILDER_IMAGE=${JVM_BUILD_SERVICE_JDK11_BUILDER_IMAGE%@*}
+  export JVM_BUILD_SERVICE_JDK11_BUILDER_IMAGE_REPO=${JVM_BUILD_SERVICE_JDK11_BUILDER_IMAGE:-"quay.io/redhat-appstudio/hacbs-jdk11-builder"}
+  export JVM_BUILD_SERVICE_JDK11_BUILDER_IMAGE_TAG=${JVM_BUILD_SERVICE_JDK11_BUILDER_IMAGE_TAG:-"redhat-appstudio-jvm-build-service-jdk11-builder-image"}
+
+  export JVM_BUILD_SERVICE_JDK17_BUILDER_IMAGE=${JVM_BUILD_SERVICE_JDK17_BUILDER_IMAGE%@*}
+  export JVM_BUILD_SERVICE_JDK17_BUILDER_IMAGE_REPO=${JVM_BUILD_SERVICE_JDK17_BUILDER_IMAGE:-"quay.io/redhat-appstudio/hacbs-jdk17-builder"}
+  export JVM_BUILD_SERVICE_JDK17_BUILDER_IMAGE_TAG=${JVM_BUILD_SERVICE_JDK17_BUILDER_IMAGE_TAG:-"redhat-appstudio-jvm-build-service-jdk17-builder-image"}
 
   if [[ -n "${JOB_SPEC}" && "${REPO_NAME}" == "jvm-build-service" ]]; then
       # Extract PR author and commit SHA to also override default kustomization in infra-deployments repo
--- a/Dockerfile.cache-in-ci
+++ b/Dockerfile.cache-in-ci
@@ -5,6 +5,9 @@ USER 0
 RUN cd java-components/cache && mvn clean package -DskipTests
 
 FROM registry.access.redhat.com/ubi8/openjdk-17-runtime:1.13
+USER 0
+RUN mkdir /cache && chown 185 /cache
+USER 185
 WORKDIR /work/
 
 COPY --from=builder /work/java-components/cache/target/quarkus-app/lib/ /deployments/lib/
@@ -13,7 +16,6 @@ COPY --from=builder /work/java-components/cache/target/quarkus-app/app/ /deploym
 COPY --from=builder /work/java-components/cache/target/quarkus-app/quarkus/ /deployments/quarkus/
 
 EXPOSE 8080
-USER 1001
 
 ENV AB_JOLOKIA_OFF=""
 ENV JAVA_OPTS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
--- a/deploy/base/config.yaml
+++ b/deploy/base/config.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: jvm-build-config
+data:
+  enable-rebuilds: "true"
+  maven-repository-300-jboss: "https://repository.jboss.org/nexus/content/groups/public/"
+  maven-repository-301-jitpack: "https://jitpack.io"
+  maven-repository-302-confluent: "https://packages.confluent.io/maven"
+  maven-repository-303-gradle: "https://repo.gradle.org/artifactory/libs-releases"
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -3,21 +3,26 @@ package controller
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
+	v1 "k8s.io/api/core/v1"
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	k8sscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/redhat-appstudio/jvm-build-service/pkg/apis/jvmbuildservice/v1alpha1"
 	"github.com/redhat-appstudio/jvm-build-service/pkg/reconciler/artifactbuild"
+	"github.com/redhat-appstudio/jvm-build-service/pkg/reconciler/configmap"
 	"github.com/redhat-appstudio/jvm-build-service/pkg/reconciler/dependencybuild"
 
 	pipelinev1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
@@ -64,9 +69,40 @@ func NewManager(cfg *rest.Config, options manager.Options) (manager.Manager, err
 			&pipelinev1beta1.PipelineRun{}: {Label: labels.SelectorFromSet(map[string]string{artifactbuild.PipelineRunLabel: ""})},
 			&v1alpha1.DependencyBuild{}:    {},
 			&v1alpha1.ArtifactBuild{}:      {},
+			&v1.ConfigMap{}:                {},
 		}})
 	mgr, err := manager.New(cfg, options)
 	if err != nil {
+		return nil, err
+	}
+	nonCachingClient, err := client.New(mgr.GetConfig(), client.Options{Scheme: options.Scheme})
+	if err != nil {
+		controllerLog.Error(err, "unable to initialize non cached client")
+		os.Exit(1)
+	}
+	var systemConfig = v1.ConfigMap{}
+	if err := wait.PollImmediate(time.Second*5, time.Minute*5, func() (done bool, err error) {
+		err = nonCachingClient.Get(context.TODO(), types.NamespacedName{Namespace: configmap.SystemConfigMapNamespace, Name: configmap.SystemConfigMapName}, &systemConfig)
+		if err != nil {
+			return false, err
+		}
+		var keys = map[string]bool{}
+		for k, v := range configmap.RequiredKeys {
+			keys[k] = v
+		}
+		for k := range systemConfig.Data {
+			delete(keys, k)
+		}
+		if len(keys) > 0 {
+			for k := range keys {
+				controllerLog.Info(fmt.Sprintf("Missing required keys in system config map %s", k))
+			}
+			return false, nil
+		}
+		controllerLog.Info("config map loaded and has required keys")
+		return true, nil
+	}); err != nil {
+		controllerLog.Error(err, "timed out waiting for system ConfigMap "+configmap.SystemConfigMapNamespace+"/"+configmap.SystemConfigMapName+" to be created")
 		return nil, err
 	}
 
@@ -75,6 +111,10 @@ func NewManager(cfg *rest.Config, options manager.Options) (manager.Manager, err
 	}
 
 	if err := dependencybuild.SetupNewReconcilerWithManager(mgr); err != nil {
+		return nil, err
+	}
+
+	if err := configmap.SetupNewReconcilerWithManager(mgr, systemConfig.Data); err != nil {
 		return nil, err
 	}
 
--- a/pkg/reconciler/configmap/configmap.go
+++ b/pkg/reconciler/configmap/configmap.go
@@ -1,0 +1,357 @@
+// This package contains the reconciler for the system and user level config
+// System level config is a single ConfigMap, while the user level config is
+// per namespace
+package configmap
+
+import (
+	"context"
+	"fmt"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/tools/record"
+	"regexp"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const (
+	//TODO eventually we'll need to decide if we want to make this tuneable
+	contextTimeout              = 300 * time.Second
+	UserConfigMapName           = "jvm-build-config"
+	UserConfigAdditionalRecipes = "additional-build-recipes"
+	CacheDeploymentName         = "jvm-build-workspace-artifact-cache"
+	LocalstackDeploymentName    = "jvm-build-workspace-localstack"
+	EnableRebuilds              = "enable-rebuilds"
+	SystemConfigMapName         = "jvm-build-system-config"
+	SystemConfigMapNamespace    = "jvm-build-service"
+	SystemCacheImage            = "image.cache"
+	SystemBuilderImages         = "builder-image.names"
+	SystemBuilderImageFormat    = "builder-image.%s.image"
+)
+
+var (
+	RequiredKeys = map[string]bool{SystemCacheImage: true}
+	log          = ctrl.Log.WithName("configmap")
+)
+
+type ReconcileConfigMap struct {
+	client               client.Client
+	scheme               *runtime.Scheme
+	eventRecorder        record.EventRecorder
+	configuredCacheImage string
+}
+
+func newReconciler(mgr ctrl.Manager, config map[string]string) reconcile.Reconciler {
+	return &ReconcileConfigMap{
+		client:               mgr.GetClient(),
+		scheme:               mgr.GetScheme(),
+		eventRecorder:        mgr.GetEventRecorderFor("ArtifactBuild"),
+		configuredCacheImage: config[SystemCacheImage],
+	}
+}
+
+func (r *ReconcileConfigMap) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	// Set the ctx to be Background, as the top-level context for incoming requests.
+	ctx, cancel := context.WithTimeout(ctx, contextTimeout)
+	defer cancel()
+	if request.Name == SystemConfigMapName && request.Namespace == SystemConfigMapNamespace {
+		return r.handleSystemConfigMap(ctx, request)
+	}
+	if request.Name != UserConfigMapName {
+		return reconcile.Result{}, nil
+	}
+	configMap := corev1.ConfigMap{}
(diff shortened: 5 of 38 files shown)


pkg/reconciler/configmap/configmap.go
```diff
@@ -0,0 +1,200 @@
...
+				ImagePullPolicy: "Always",
+				SecurityContext: &corev1.SecurityContext{RunAsUser: i64a(0)},
+				Ports: []corev1.ContainerPort{{
+					Name:          "http",
+					ContainerPort: 8080,
+					Protocol:      "TCP",
+				}},
+				Resources: corev1.ResourceRequirements{
+					Requests: map[corev1.ResourceName]resource.Quantity{"memory": resource.MustParse("128Mi"), "cpu": resource.MustParse("250m")},
+					Limits:   map[corev1.ResourceName]resource.Quantity{"memory": resource.MustParse("700Mi"), "cpu": resource.MustParse("500m")},
```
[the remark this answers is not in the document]
stuartwdouglas: We will need some form of configuration here. Big workspaces may need caches with more resources.

pkg/reconciler/configmap/configmap.go
```diff
@@ -0,0 +1,200 @@
...
+				},
+			}}
+			cache.Spec.Template.Spec.ServiceAccountName = "hacbs-jvm-cache"
+
+			//and setup the service
+			service := corev1.Service{
```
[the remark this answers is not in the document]
stuartwdouglas: I was thinking we could just mount the configmap in the sidecar pod and have it read this configuration directly, however for this particular bit of config I think we just want to use ${context.taskRun.namespace} to get the namespace.

pkg/reconciler/configmap/configmap.go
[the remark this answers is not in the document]
gabemontero: yep .... and since my comment we had the cabal discussion with @sbose78 and the rest of the 3 pipeline service teams and reached agreement that there is a broader concern around setting up workspaces/namespaces with more quota based on request and additional payment, not just jvm build service

that said, in the interim, I think you should define keys in the configmap for these, where the values you've picked now could be the default if those key are missing.

pkg/reconciler/configmap/configmap.go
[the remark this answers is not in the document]
gabemontero: +1 on both those ideas @stuartwdouglas

pkg/reconciler/configmap/configmap.go
```diff
@@ -0,0 +1,423 @@
...
+type ReconcileConfigMap struct {
+	client               client.Client
+	scheme               *runtime.Scheme
+	eventRecorder        record.EventRecorder
+	configuredCacheImage string
+	builderImages        *BuilderImageConfig
+}
+
+type BuilderImageConfig struct {
+	Lock   sync.Locker
```
stuartwdouglas, line 57: @gabemontero can I get your opinion on this bit? Basically the builder images are needed in the dependencybuild reconciler, and it seems really wasteful to have to explcitiy grab the config map there every time we do a build. 

Instead I just have this shared BuilderImageConfig object and access it under a lock.

.ci/common-ci-settings-and-functions.sh
```diff
@@ -53,6 +56,18 @@ export JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE=${JVM_BUILD_SERVICE_REQPROCESSOR_IMA
 export JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE_REPO=${JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE:-"quay.io/redhat-appstudio/hacbs-jvm-build-request-processor"}
 export JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE_TAG=${JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE_TAG:-"redhat-appstudio-jvm-build-service-reqprocessor-image"}
 
+export JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE=${JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE%@*}
```
stuartwdouglas, line 59: @gabemontero where are these images actually built? I will need to build the builder images now as well but I can't seem to find where this build+env var setting happens.

psturc: @stuartwdouglas [here's the openshift-ci job config file](https://github.com/openshift/release/blob/master/ci-operator/config/redhat-appstudio/jvm-build-service/redhat-appstudio-jvm-build-service-main.yaml) where they need to be added

you can find more info in the [readme](https://github.com/redhat-appstudio/jvm-build-service/blob/main/.ci/README.md)

gabemontero: @stuartwdouglas this is what we opened https://issues.redhat.com/browse/PLNSRVCE-517 after our exchange start started with my comment https://github.com/redhat-appstudio/jvm-build-service/pull/106#pullrequestreview-1042028142
and ended with your comment https://github.com/redhat-appstudio/jvm-build-service/pull/106#issuecomment-1192064239

gabemontero: I tried the capture all the relevant steps in my update to the description in but of course we can talk, divide and conquer as needed if you are ready to take on https://issues.redhat.com/browse/PLNSRVCE-517

@Michkov FYI

pkg/reconciler/configmap/configmap.go
gabemontero: yeah since we are reconciling on the config map it should be in the watch cache and hence grabbing it each time using the controller runtime client is suppose to be a "local fetch / in memory only / cache hit" with no RPC.

general rule of thumb we don't want to synchronize across events like this

.ci/common-ci-settings-and-functions.sh
stuartwdouglas: https://github.com/openshift/release/pull/30919

pkg/reconciler/configmap/configmap.go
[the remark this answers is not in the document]
gabemontero: wanted to circle back to this thread @stuartwdouglas .... I'm not seeing anything in https://github.com/redhat-appstudio/build-definitions/pull/179/files or https://github.com/redhat-appstudio/jvm-build-service/blob/c1032fe8ca4f0b0c95d16b34361ace5df649b321/pkg/reconciler/dependencybuild/buildrecipeyaml.go around mounting in the config map

and I see you deleted the env var QUARKUS_REST_CLIENT_CACHE_SERVICE_URL I mentioned with https://github.com/redhat-appstudio/jvm-build-service/pull/105/files#r923400881

is it still a to-do (perhaps in that ^^ build def PR) to add the config map mount?  or are you going in a different direction?

thanks

gabemontero (changes requested):
ok @stuartwdouglas I started looking at this PR on Monday Aug 1 after the stand up meeting we were both on and see I have a pending review from last Friday I never hit submit review on

sign - too much multi tasking for me on Friday

I've cleaned up the review up a bit and and hitting submit review now.  I'll then take an additional pass in case there was an update you made today that I need to see.

.ci/common-ci-settings-and-functions.sh
gabemontero: update for anyone coming into this thread mid stream ... https://github.com/openshift/release/pull/30919 has merged

after these changes here, next step is to add the requisite updates in infra-deps (but correct me if I'm wrong @stuartwdouglas ) to consume them.

pkg/reconciler/configmap/configmap.go
```diff
@@ -0,0 +1,200 @@
...
+	defer cancel()
+	if request.Name != ConfigMapName {
+		return reconcile.Result{}, nil
+	}
+	configMap := corev1.ConfigMap{}
+	err := r.client.Get(ctx, request.NamespacedName, &configMap)
+	if err != nil {
+		return reconcile.Result{}, nil
+	}
+	enabled := configMap.Data[EnableRebuilds] == "true"
```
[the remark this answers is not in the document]
gabemontero: fyi we have agreed that ^^ is beyond the scope of this PR

(though I do want to start pursuing the concept at least with @sbose78 when he returns from PTO)

pkg/reconciler/configmap/configmap.go
```diff
@@ -0,0 +1,396 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+	type Repo struct {
+		name     string
+		position int
+	}
+	//TODO: make configurable
```
gabemontero, line 166: I suppose my question from https://github.com/redhat-appstudio/jvm-build-service/pull/105/files#r932615502 applies here too ... need a mem refresh from one of the voice to voice chats we had last week.

pkg/reconciler/configmap/configmap.go
```diff
@@ -0,0 +1,396 @@
...
+	})
+	cache.Spec.Template.Spec.Containers[0].Image = r.configuredCacheImage
+
+	role := v1alpha1.Role{}
+	roleName := types.NamespacedName{Namespace: request.Namespace, Name: "pipeline-anyuid-role"}
+	err = r.client.Get(ctx, roleName, &role)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			role := v1alpha1.Role{}
+			role.Name = "pipeline-anyuid-role"
```
gabemontero, line 240: Preface: I understand the need for this change here with what this PR is accomplishing.

That said, a controller adding roles and rolebindings (i.e. manipulate permissions) is atypical (and can in fact be frowned upon).

Now, needing anyuid is a function of the base images we are leveraging during our builds not giving us write access to content we need.

We are not the only item under the appstudio umbrella fighting this (the buildah task is another one, where it has a path around user workspaces for eventually solving this (though that is not there in openshift 4.10 IIRC so that I think is a big change).

But I have had brief conversations with Adam Kaplan about "long term" not wanting any appstudio components to leverage anyuid.

Bottom line, *minimally* we need a big TODO comment here about we know we need to see about create images that do not need anyuid, where that is complex enough that if we do not already have a Jira story that tracks this work, we should either update an existing story, or create a new one.

@Michkov FYI

gabemontero:
OK done with PR review @stuartwdouglas 

e2e failure ` .ci/common-ci-settings-and-functions.sh: line 57: JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE: unbound variable ` should be addressed with the openshift/release merge I believe

/retest

gabemontero:
bad luck with app studio bootstrap:

```
         1.6593693589970229e+09	ERROR	controller	timed out waiting for taskrun CRD to be created	{"error": "configmaps \"jvm-build-system-config\" not found"}
        github.com/redhat-appstudio/jvm-build-service/pkg/controller.NewManager
        	/opt/app-root/src/pkg/controller/controller.go:118
        main.main
        	/opt/app-root/src/cmd/controller/main.go:39
        runtime.main
        	/usr/lib/golang/src/runtime/proc.go:255
        1.6593693589970791e+09	ERROR	main	unable to start manager	{"error": "configmaps \"jvm-build-system-config\" not found"}
        main.main
        	/opt/app-root/src/cmd/controller/main.go:48
        runtime.main
        	/usr/lib/golang/src/runtime/proc.go:255
```

/retest

gabemontero:
same error as before 

```
         1.659372107494306e+09	ERROR	controller	timed out waiting for taskrun CRD to be created	{"error": "configmaps \"jvm-build-system-config\" not found"}
        github.com/redhat-appstudio/jvm-build-service/pkg/controller.NewManager
        	/opt/app-root/src/pkg/controller/controller.go:118
        main.main
        	/opt/app-root/src/cmd/controller/main.go:39
        runtime.main
        	/usr/lib/golang/src/runtime/proc.go:255
        1.6593721074943602e+09	ERROR	main	unable to start manager	{"error": "configmaps \"jvm-build-system-config\" not found"}
        main.main
        	/opt/app-root/src/cmd/controller/main.go:48
        runtime.main
        	/usr/lib/golang/src/runtime/proc.go:255 
``` 

possibly related to my comment https://github.com/redhat-appstudio/jvm-build-service/pull/105/files#r932612117 something must be amiss with that addition, especially in light that CRD check is already in main branch at https://github.com/redhat-appstudio/jvm-build-service/blob/7d2f00b2bb9ed3de167a76e8533d03921518026d/pkg/controller/controller.go#L35-L46

I'll start comparing files in a few

pkg/controller/controller.go
```diff
@@ -45,6 +51,19 @@ func NewManager(cfg *rest.Config, options manager.Options) (manager.Manager, err
 		return nil, err
 	}
 
+	if err := wait.PollImmediate(time.Second*5, time.Minute*5, func() (done bool, err error) {
+		_, err = apiextensionsClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.TODO(), "taskruns.tekton.dev", metav1.GetOptions{})
```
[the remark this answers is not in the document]
gabemontero: OK I looked at the entire file and it became obvious.  the CRD check is duplicated @stuartwdouglas 

https://github.com/redhat-appstudio/jvm-build-service/blob/c1032fe8ca4f0b0c95d16b34361ace5df649b321/pkg/controller/controller.go#L40-L64

Remove this duplicate chunk

gabemontero:
duh - I need better glasses:

`        1.659372107494306e+09	ERROR	controller	timed out waiting for taskrun CRD to be created	{"error": "configmaps \"jvm-build-system-config\" not found"}`

it is the new code looking for the system config map
though that message should be updated to not say taskrun CRD 

(that said the duplication of the CRD check needs to be cleaned up also)

this PR's test needs the PRs @stuartwdouglas noted in the description (minimally https://github.com/redhat-appstudio/infra-deployments/pull/537)

sorry for the noise @stuartwdouglas .... I'll clean up some noisy comments

pkg/controller/controller.go
```diff
@@ -64,11 +82,42 @@ func NewManager(cfg *rest.Config, options manager.Options) (manager.Manager, err
...
+		if len(keys) > 0 {
+			for k := range keys {
+				controllerLog.Info(fmt.Sprintf("Missing required keys in system config map %s", k))
+			}
+			return false, nil
+		}
+		controllerLog.Info("config map loaded and has required keys")
+		return true, nil
+	}); err != nil {
+		controllerLog.Error(err, "timed out waiting for taskrun CRD to be created")
```
gabemontero, line 118:
```suggestion
		controllerLog.Error(err, "timed out waiting for system config map to be properly intialized")
```

pkg/reconciler/configmap/configmap.go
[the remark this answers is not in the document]
stuartwdouglas: We definitly want it to be system level configurable, probably also individually configurable (assuming we have some kind of per namespace quota, so a user can't just increment this to get more total memory).

pkg/reconciler/configmap/configmap.go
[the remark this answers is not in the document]
stuartwdouglas: We don't actually need a config map for this, we can handle the per workspace cache URI via standard tekton config expansion. If we do want to configure the sidecar we will need to mount the config map but I want to limit the blast radius of this PR so IMHO that can come later.

pkg/reconciler/configmap/configmap.go
stuartwdouglas: I have changed this to just get the config map in the db reconciler

.ci/common-ci-settings-and-functions.sh
stuartwdouglas: https://github.com/redhat-appstudio/infra-deployments/pull/537 configures the config map to reference them in infra deployments.

deploy/base/config.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: jvm-build-config
+data:
+  enable-rebuilds: "true"
```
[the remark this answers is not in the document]
stuartwdouglas: At the moment the user would create it to enable rebuilds, in a similar manner to how they need to configure the hacbs bundle. 

There are no permissions on this yet so anyone can enable this, but I don't think this is a big deal at this stage.

pkg/controller/controller.go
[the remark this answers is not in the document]
stuartwdouglas: oops

pkg/reconciler/configmap/configmap.go
stuartwdouglas: This PR actually changes the base image to the ones built in openshift, so I can probably make that image change as part of this PR. 

I will need to do some local testing.

pkg/reconciler/configmap/configmap.go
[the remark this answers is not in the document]
gabemontero: we can assume per namespace quota

and I'm good with staging system level vs. individual configuration separately (with system first) if that seems like the correct path to you @stuartwdouglas 

do you want to include  in this PR or a follow up?

pkg/reconciler/configmap/configmap.go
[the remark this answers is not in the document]
gabemontero: ah right .... and that explain why you deleted the env var ... ok yeah makes sense

I'm totally fine with deferring configurable side car / config map mount outside this PR ... might even be something we defer indefinitely - we'll see.

pkg/reconciler/configmap/configmap.go
gabemontero: yep in conjunction with https://github.com/redhat-appstudio/jvm-build-service/pull/105/files#diff-243ebed2765f75e6a54f57167212fefb08c3b2a85967ad2acbc0eb78919019c1R84 looks good  / thanks

marking resolved

.ci/common-ci-settings-and-functions.sh
gabemontero: yep thanks for that more specif ref @stuartwdouglas 

marking resolved

deploy/base/config.yaml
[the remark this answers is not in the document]
gabemontero: ok thanks for the confirm

and yeah the permission etc. we agreed is a follow up item, where my idea at https://github.com/redhat-appstudio/jvm-build-service/pull/105#discussion_r923391438 is the first attempt at coming up with the approach for that future implementation

marking as resolved

pkg/controller/controller.go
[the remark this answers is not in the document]
gabemontero: looks good now marking resolved

pkg/reconciler/configmap/configmap.go
gabemontero: ok cool - we'll play it by ear.

initial thought looking for example (but not limited to) we would tweak https://github.com/redhat-appstudio/jvm-build-service/blob/main/Dockerfile.sidecar-in-ci so that before we set the user to non-root at https://github.com/redhat-appstudio/jvm-build-service/blob/main/Dockerfile.sidecar-in-ci#L18 we create dirs and update permissions so that we no longer the various `chown` calls and use of anyuid scc in our various tasks (i.e. https://github.com/redhat-appstudio/jvm-build-service/blob/3a19938e522f16830cb6c67449ad0bf5117ffc3c/pkg/reconciler/dependencybuild/buildrecipeyaml.go#L253-L254)

The only task not unique to jvm build service is the git-clone task (which is in build-serivce, but I think is basically a copy of the tekton cluster task.  I think it would need to be solved as well long term, but since that effects everyone, probably would need a broader discussion / review.

gabemontero:
some github throttling in the image failure

appstudio bootstrap failure in jvm-build-service-e2e though I just merged a build-service fix that should help there

the sample component maven-goals pod failed, so no artifactbuild / dependencybuild generation:

```
    jvm_build_service_test.go:51: time: 2022-08-02 10:37:07.264819823 +0000 UTC m=+641.164228848: pod logs for container step-mvn-goals in pod sample-component-build-rj2gt-maven-run-pod:  [INFO] Scanning for projects...
        [INFO] 
        [INFO] ------------------< org.example:simple-java-project >-------------------
        [INFO] Building simple-java-project 1.0-SNAPSHOT
        [INFO] --------------------------------[ jar ]---------------------------------
        Downloading from mirror.default: http://localhost:2000/maven2/org/apache/maven/plugins/maven-shade-plugin/3.2.4/maven-shade-plugin-3.2.4.pom
        [INFO] ------------------------------------------------------------------------
        [INFO] BUILD FAILURE
        [INFO] ------------------------------------------------------------------------
        [INFO] Total time:  30.921 s
        [INFO] Finished at: 2022-08-02T10:29:41Z
        [INFO] ------------------------------------------------------------------------
        [ERROR] Plugin org.apache.maven.plugins:maven-shade-plugin:3.2.4 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-shade-plugin:jar:3.2.4: Could not transfer artifact org.apache.maven.plugins:maven-shade-plugin:pom:3.2.4 from/to mirror.default (http://localhost:2000/maven2): Transfer failed for http://localhost:2000/maven2/org/apache/maven/plugins/maven-shade-plugin/3.2.4/maven-shade-plugin-3.2.4.pom 500 Internal Server Error -> [Help 1]
        [ERROR] 
        [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
        [ERROR] Re-run Maven using the -X switch to enable full debug logging.
        [ERROR] 
        [ERROR] For more information about the errors and possible solutions, please read the following articles:
        [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginResolutionException
        cat: /workspace/maven-settings/sidecar.log: No such file or directory
```

gabemontero:
hmm ... images failed again ... may not be a flake ... looking again

gabemontero:
Nope - same thing - very weird / bad luck, this PR is not changing what is running elsewhere best as I can tell

/test images

stuartwdouglas:
The sidecar is running out of memory on the image build.

gabemontero:
Last test the operator was complaining about the config map not being available.

But now https://github.com/redhat-appstudio/infra-deployments/pull/537 has merged, so let's try again.

gabemontero:
OK controller still getting 

```
         1.6595440611397045e+09	ERROR	controller.configmap	Reconciler error	{"reconciler group": "", "reconciler kind": "ConfigMap", "name": "jvm-build-config", "namespace": "jvm-build-service-test-namespace-sls8w", "error": "roles.rbac.authorization.k8s.io \"pipeline-anyuid-role\" not found"}
        sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        	/opt/app-root/src/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:266
        sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        	/opt/app-root/src/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:227
```

complaining about  the role in particular ... I'll spend cycles now/this afternoon trying to sort out.

pkg/reconciler/configmap/configmap.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	cb := v13.RoleBinding{}
+	cbName := types.NamespacedName{Namespace: request.Namespace, Name: "hacbs-jvm-cache-anyuid-rolebinding"}
+	err = r.client.Get(ctx, cbName, &cb)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			cb := v13.RoleBinding{}
+			cb.Name = "hacbs-jvm-cache-anyuid-rolebinding"
+			cb.Namespace = request.Namespace
+			cb.RoleRef = v13.RoleRef{Kind: "Role", Name: "pipeline-anyuid-role", APIGroup: "rbac.authorization.k8s.io"}
```
gabemontero, line 255:
```suggestion
			cb.RoleRef = v13.RoleRef{Kind: "ClusterRole", Name: "pipeline-anyuid-role", APIGroup: "rbac.authorization.k8s.io"}
```

gabemontero (changes requested):
I see the problem with the e2e error log I noted ^^ @stuartwdouglas 

You changed `pipeline-anyuid-role` to a `ClusterRole` in https://github.com/redhat-appstudio/jvm-build-service/pull/105/files#diff-aead536da1d8e62277d27b3f65e8a0153ff084e61727d0123d15d881368bfe02 but your code here still cites a namespace scoped `Role`

stuartwdouglas:
@gabemontero jvm-build-service-in-repo-e2e is green, and I have removed the need for privileged containers. I am going to squash this and do some cleanup and it should be good to go.

gabemontero:
> @gabemontero jvm-build-service-in-repo-e2e is green, and I have removed the need for privileged containers. I am going to squash this and do some cleanup and it should be good to go.

awesome on both counts !!

still morning for me, but as of this moment, not going to merge this *yet* because we also need to really get https://github.com/redhat-appstudio/infra-deployments/pull/558 merged to facilitate the e2e work over in e2e-tests, and merging this will interrupt that effort.  And given the amount of non-jvm flakes we are already hitting there, I want to avoid any interrupts I can control.

That said, if we end up at my EOB and that PR is still struggling, I'll merge and we can pick back up on https://github.com/redhat-appstudio/infra-deployments/pull/558 and the various CI testing tomorrow / Monday.

gabemontero:
@stuartwdouglas - can you point out to me where in the PR correlates to your removing the need for privileged containers?

I *think* it relates to earlier forms of the PR where you were assigning the anyuid role, but I don't see that anymore ... is that it?

Otherwise, when I first read of, I thought of https://github.com/redhat-appstudio/jvm-build-service/blob/650846d2011d745b82c72cd59cf1e2398daeddfa/pkg/reconciler/dependencybuild/buildrecipeyaml.go#L137 for example, but of course the use of runAsUser is still there.

thanks

Dockerfile.cache-in-ci
```diff
@@ -5,6 +5,9 @@ USER 0
 RUN cd java-components/cache && mvn clean package -DskipTests
 
 FROM registry.access.redhat.com/ubi8/openjdk-17-runtime:1.13
+USER 0
+RUN mkdir /cache && chown 185 /cache
+USER 185
```
gabemontero, line 10: ahh ... I bet this is the no longer need privileged change you are talking about

side bar, is there something special about user 185?  In case you were not aware, when the pod runs in openshift, the actual UID will not be 185.

gabemontero: I would think this change would allow us to remove `runAsUse: 0` then in https://github.com/redhat-appstudio/jvm-build-service/blob/650846d2011d745b82c72cd59cf1e2398daeddfa/deploy/cache/base/deployment.yaml#L24

did you try that @stuartwdouglas ?

gabemontero:
OK https://github.com/redhat-appstudio/infra-deployments/pull/558 has merged.

That said, I'm now curious about https://github.com/redhat-appstudio/jvm-build-service/pull/105#discussion_r938031556 and a) did you try it and it did not work, or b) did not try it, but are willing to do so in this PR

So I'll defer hitting the merge button @stuartwdouglas to you, where either 
- you merge because the answer to a) was "it did not work" ... I can look into why on Monday unless you already know
- the answer to b) is you want to defer that change to a follow up PR
- you decide to make the change to remove `runAsUser: 0` and then you merge this after CI passes with that change

Dockerfile.cache-in-ci
stuartwdouglas: User 185 comes from the base image, that file is only still there to allow infra deployments to work in the short term, as now the cache is workspace scoped.

stuartwdouglas:
@gabemontero the cache deployment yaml is a transitional thing, once this is merged and infra deployments is updated it will be deleted, which is why I did not look at fixing the user.

I also looked removing it from the tasks but there seems to be an issue with the git clone task not specifying a user and the Java build ones do, which results in the workspace not being writable.

gabemontero:
> @gabemontero the cache deployment yaml is a transitional thing, once this is merged and infra deployments is updated it will be deleted, which is why I did not look at fixing the user.
> 
> I also looked removing it from the tasks but there seems to be an issue with the git clone task not specifying a user and the Java build ones do, which results in the workspace not being writable.

sounds good cache deployment and yeah forgot about the git clone (I remember having to make and update for that now) ... would think that is a app studio wide issue, though I'm not sure if I've seen the need for that in the other build tasks.  I'll revisit that next week.

So yeah merge away at your leisure @stuartwdouglas

PR #106, opened 2022-07-18
PLNSRVCE-325: Add Builder Images

stuartwdouglas:
@gabemontero @Michkov this is my first attempt at actually building the builder images in CI. At the moment there is just 3 of them although this will increase as time goes on, so I am using a template and a script to generate the docker files.

We also need some way to override the builder images that are actually used, which I will look at once this is working.

### DIFF
--- a/builder-images/generate.sh
+++ b/builder-images/generate.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+
+generate () {
+  mkdir $IMAGE_NAME
+  envsubst '$IMAGE_NAME,$BASE_IMAGE,$MAVEN_VERSION,$MAVEN_SHA,$GRADLE_VERSION,$GRADLE_SHA' < Dockerfile.template > $IMAGE_NAME/Dockerfile
+  envsubst '$IMAGE_NAME,$BASE_IMAGE' < push.yaml > ../.tekton/$IMAGE_NAME-push.yaml
+  envsubst '$IMAGE_NAME,$BASE_IMAGE' < pull-request.yaml > ../.tekton/$IMAGE_NAME-pull-request.yaml
+}
+
+export MAVEN_VERSION=3.8.6
+export MAVEN_SHA=f790857f3b1f90ae8d16281f902c689e4f136ebe584aba45e4b1fa66c80cba826d3e0e52fdd04ed44b4c66f6d3fe3584a057c26dfcac544a60b301e6d0f91c26
+export GRADLE_VERSION=7.5
+export GRADLE_SHA=cb87f222c5585bd46838ad4db78463a5c5f3d336e5e2b98dc7c0c586527351c2
+
+export IMAGE_NAME=hacbs-jdk17-builder
+export BASE_IMAGE=registry.access.redhat.com/ubi8/openjdk-17
+generate
+
+
+export IMAGE_NAME=hacbs-jdk8-builder
+export BASE_IMAGE=registry.access.redhat.com/ubi8/openjdk-8
+generate
+
+
+export IMAGE_NAME=hacbs-jdk11-builder
+export BASE_IMAGE=registry.access.redhat.com/ubi8/openjdk-11
+generate
--- a/builder-images/hacbs-jdk11-builder/Dockerfile
+++ b/builder-images/hacbs-jdk11-builder/Dockerfile
@@ -1,0 +1,56 @@
+FROM registry.access.redhat.com/ubi8/openjdk-11
+USER root
+ENV \
+    LANG="en_US.UTF-8" \
+    LANGUAGE="en_US:en" \
+    LC_ALL="en_US.UTF-8" \
+    APP_HOME="/home/hacbs" \
+    MAVEN_HOME="/usr/share/maven" \
+    MAVEN_OPTS="-XX:+TieredCompilation -XX:TieredStopAtLevel=1" \
+    MAVEN_VERSION="3.8.6" \
+    GRADLE_HOME="/usr/share/gradle" \
+    GRADLE_OPTS="-Dorg.gradle.daemon=false" \
+    GRADLE_VERSION="7.5" \
+    HOME="/home/hacbs" \
+    NSS_WRAPPER_PASSWD="/etc/passwd" \
+    GRADLE_DOWNLOAD_SHA256=cb87f222c5585bd46838ad4db78463a5c5f3d336e5e2b98dc7c0c586527351c2 \
+    PATH="$PATH:$JAVA_HOME/bin" \
+    BASE_URL=https://apache.osuosl.org/maven/maven-3/3.8.6/binaries \
+    MAVEN_SHA=f790857f3b1f90ae8d16281f902c689e4f136ebe584aba45e4b1fa66c80cba826d3e0e52fdd04ed44b4c66f6d3fe3584a057c26dfcac544a60b301e6d0f91c26
+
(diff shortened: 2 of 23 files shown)


gabemontero:
@dwalluck IIRC you asked about builder images several times while @stuartwdouglas was out.

I suspect you'll minimally want to monitor this, if not comment on it.

gabemontero (commented):
So I'm assuming you ran `generate.sh` to produce the jdk11 and jdk17 builder images @stuartwdouglas .. is that correct ?

If so, do you envision that being a one time operator, or something that needs to be done each time we build the image?  From what I see so far, I'm thinking one time operator, but want to make sure.

My motivation for all this is that while you've covered the `.tekton` based verification, if you want to include validation of these builder images via the e2e runs that test against openshift clusters running appstudio, we'll need to update https://github.com/openshift/release/blob/master/ci-operator/config/redhat-appstudio/jvm-build-service/redhat-appstudio-jvm-build-service-main.yaml#L4-L14 after this merges to build the images, and update https://github.com/openshift/release/blob/master/ci-operator/config/redhat-appstudio/jvm-build-service/redhat-appstudio-jvm-build-service-main.yaml#L51-L77 to define env vars for those built images that we can they leverage to update Tekton yaml in either infra-deployments and build-definitions.

Now, when I did the last round of connecting the dots here, I had to make minor tweaks to your Dockerfiles under `java-components` to get the images builds to work on the openshift CI cluster.  At the time, I chose to not "pollute" the Dockefiles under `java-components` with those changes (in part this may have overlapped during our PTOs and I leaned conservative).  Hence, when building items in openshift CI, I have 
- https://github.com/redhat-appstudio/jvm-build-service/blob/main/Dockerfile.analyzer-in-ci
- https://github.com/redhat-appstudio/jvm-build-service/blob/main/Dockerfile.cache-in-ci
- https://github.com/redhat-appstudio/jvm-build-service/blob/main/Dockerfile.reqprocessor-in-ci
- https://github.com/redhat-appstudio/jvm-build-service/blob/main/Dockerfile.sidecar-in-ci

Something to consider as we move on to this next stage of adding images is do we continue this pattern, or do we have just one copy of Dockerfile, were the ones in the various subdirectories you've created take on the changes I needed to build in openshift CI.

My thinking is we want to consolidate, but let's discuss.

Minimally, I suspect we may need to manipulate `USER` like we do with the `in-ci` Dockerfiles.

Also note, running scripts like `generate.sh` on a per build basis (again, I suspect that is not what is required) won't fly in openshift CI.

stuartwdouglas:
`generate.sh` is run to generate the checked in files. It allows for having multiple similar images generated from the same templates.

gabemontero:
> `generate.sh` is run to generate the checked in files. It allows for having multiple similar images generated from the same templates.

thanks for the confirmation ... makes things easier

we also talked about how we could move the top level Dockerfiles I have and replace the all-in-ones with the top level ones, typically during our developer flow the docker images are built via the `mvn` plugins and not by `docker build` on the Dockerfiles.

I'll see about that once the e2e test updates are all merged and running correctly.

builder-images/generate.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+
+generate () {
+  mkdir $IMAGE_NAME
+  cp Dockerfile.template $IMAGE_NAME/Dockerfile
```
Michkov, line 6: use `envsubst`:
```
envsubst < Dockerfile.template > $IMAGE_NAME/Dockerfile
```
So then you do not need to use `sed`. Only replace `T_BASE_IMAGE` by `${BASE_IMAGE}` in the template.

builder-images/pull-request.yaml
```diff
@@ -0,0 +1,28 @@
...
+      value: java-components
+    - name: dockerfile
+      value: builder-images/T_IMAGE_NAME/Dockerfile
+  pipelineRef:
+    name: docker-build
+    bundle: quay.io/redhat-appstudio/hacbs-core-service-templates-bundle:latest
+  workspaces:
+    - name: workspace
+      persistentVolumeClaim:
+        claimName: app-studio-default-workspace
```
Michkov, line 27: this is causing that the build takes ages, since there are only read-write-once pvc available, let's change it to template:
```
workspaces:
  - name: workspace
    volumeClaimTemplate:
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 1Gi
```

builder-images/hacbs-jdk11-builder/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+    BASE_URL=https://apache.osuosl.org/maven/maven-3/3.8.6/binaries \
+    MAVEN_SHA=f790857f3b1f90ae8d16281f902c689e4f136ebe584aba45e4b1fa66c80cba826d3e0e52fdd04ed44b4c66f6d3fe3584a057c26dfcac544a60b301e6d0f91c26
+
+RUN groupadd -r hacbs -g 1001 && useradd -u 1001 -r -g 1001 -m -d ${APP_HOME} -s /sbin/nologin -c "HACBS user" hacbs && cp /etc/passwd /home/hacbs/passwd && chown 1001:1001 /home/hacbs/passwd && mkdir /project && chown 1001:1001 /project
+
+RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y apr-devel wget tar gzip gcc glibc-devel zlib-devel shadow-utils unzip gcc-c++ glibc-langpack-en buildah findutils podman bzip2-devel gcc-c++ libcurl-devel openssl-devel bc which lsof unzip git libstdc++-static glibc-static make autoconf automake libtool tar glibc-devel libgcc.i686 glibc-devel.i686\
+    && microdnf clean all \
+    && rpm -q apr-devel wget tar gzip gcc glibc-devel zlib-devel shadow-utils unzip gcc-c++ glibc-langpack-en buildah findutils podman bzip2-devel gcc-c++ libcurl-devel openssl-devel bc which lsof unzip git libstdc++-static glibc-static make autoconf automake libtool tar glibc-devel libgcc.i686 glibc-devel.i686
```
Michkov, line 25: curious why this verification is needed?

builder-images/hacbs-jdk17-builder/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+LABEL \
+    io.cekit.version="4.1.0" \
+    io.k8s.description="HACBS builder image for building Java components." \
+    io.k8s.display-name="HACBS BuilderImage" \
+    io.openshift.tags="builder,java,hacbs" \
+    maintainer="Stuart Douglas <sdouglas@redhat.com>"
+
+# Clear package manager metadata
+RUN microdnf clean all && [ ! -d /var/cache/yum ] || rm -rf /var/cache/yum
```
Michkov, line 50: this was already executed

Michkov (commented):
Small improvements requested, also I created repositories in quay and set access to our robot account.

gabemontero:
@Michkov @stuartwdouglas do we want to track the building of these new images in openshift CI (and eventual overriding of those images during e2e's) under PLNSRVCE-253, or do we want to break it out into a separate story?

builder-images/hacbs-jdk11-builder/Dockerfile
stuartwdouglas: This was copied/adapted from the CEKit generated dockerfiles. I just assumed this was best practice because CEKit does it.

stuartwdouglas:
I created https://issues.redhat.com/browse/PLNSRVCE-517 for the override issue. I am going to merge this as is for now, then start working on actually using these images (we really need the first versions built before we start referring to them in code).

PR #109, opened 2022-07-18
PLNSRVCE-346: Rebuild Java dependencies that use Gradle

dwalluck:


deploy/base/kustomization.yaml
```diff
@@ -2,8 +2,9 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
+  - gradle-v0.1.yaml
   - https://github.com/tektoncd/pipeline/releases/download/v0.34.1/release.yaml
-  - https://raw.githubusercontent.com/redhat-appstudio/build-definitions/main/tasks/git-clone.yaml
+  - https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.5/git-clone.yaml
```
stuartwdouglas, line 7: This looks like a rebase issue.

dwalluck: I reverted that change to make the script still using the old task work. I think need to change the parameters to the new ones.

deploy/base/gradle-v0.1.yaml
```diff
@@ -0,0 +1,362 @@
...
+      image: quay.io/redhat-appstudio/hacbs-jvm-dependency-analyser:3dcd8603061c7ab2cedfe128c80cac9726721a74
+      args:
+        - path
+        - $(workspaces.source.path)/$(params.CONTEXT_DIR)/$(params.DEPENDENCY_ANALYSER_PATH)
+        - -s
+        - $(workspaces.source.path)/$(params.CONTEXT_DIR)/$(params.DEPENDENCY_ANALYSER_PATH)/java-sbom.json
+        - -c
+        - "$(results.dependencies.path)"
+        - -u
+        - "$(results.untrusted-dependencies.path)"
```
dwalluck, line 321: This is old and shouldn't be here. But, it used to kick off the build of dependencies, but broke within the last few days.

dwalluck:
```
[process-build-requests] 2022-07-18 21:39:11,959 INFO  [io.quarkus] (main) hacbs-build-request-processor 1.0-SNAPSHOT on JVM (powered by Quarkus 2.7.5.Final) started in 0.558s.
[process-build-requests] 2022-07-18 21:39:11,960 INFO  [io.quarkus] (main) Profile prod activated.
[process-build-requests] 2022-07-18 21:39:11,960 INFO  [io.quarkus] (main) Installed features: [cdi, jgit, jsch, picocli, rest-client-reactive, smallrye-context-propagation, vertx]
[process-build-requests] Missing required option: '--gav=<gav>'
[process-build-requests] Usage: <main class> [--build-info=<buildInfo>] [--context=<context>]
[process-build-requests]                     --gav=<gav> [--message=<message>] [--scm-tag=<scmTag>]
[process-build-requests]                     [--scm-type=<scmType>] [--scm-url=<scmUrl>]
[process-build-requests]                     --recipes=<recipeRepos>[,<recipeRepos>...]
[process-build-requests]                     [--recipes=<recipeRepos>[,<recipeRepos>...]]...
[process-build-requests]       --build-info=<buildInfo>
[process-build-requests]
[process-build-requests]       --context=<context>
[process-build-requests]       --gav=<gav>
[process-build-requests]       --message=<message>
[process-build-requests]       --recipes=<recipeRepos>[,<recipeRepos>...]
[process-build-requests]
[process-build-requests]       --scm-tag=<scmTag>
[process-build-requests]       --scm-type=<scmType>
[process-build-requests]       --scm-url=<scmUrl>
```

deploy/base/gradle-v0.1.yaml
```diff
@@ -0,0 +1,362 @@
+apiVersion: tekton.dev/v1beta1
```
gabemontero, line 1: so fyi, assuming my e2e change merges before this, I would expect your PR here to add an e2e similar to the maven based one I initially introduced where, but it leverages this gradle file and the associated run-gradle.yaml example you have.

hack/examples/run-gradle.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: tekton.dev/v1beta1
```
gabemontero, line 1: yeah an e2e with this 

of course you would use the sample pipeline the maven based example I initially did uses

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -258,7 +258,9 @@ func (r *ReconcileDependencyBuild) handleStateAnalyzeBuild(ctx context.Context,
 		buildRecipes := []*v1alpha1.BuildRecipe{}
 		_, maven := unmarshalled.Tools["maven"]
 		_, gradle := unmarshalled.Tools["gradle"]
-		for _, image := range []string{"quay.io/sdouglas/hacbs-jdk11-builder:latest", "quay.io/sdouglas/hacbs-jdk8-builder:latest", "quay.io/sdouglas/hacbs-jdk17-builder:latest"} {
+		// FIXME: temporarily changed image for gradle to be replaced after <https://github.com/redhat-appstudio/jvm-build-service/pull/106>
```
gabemontero, line 261: yeah maybe with @stuartwdouglas 's new configmap based configuring https://github.com/redhat-appstudio/jvm-build-service/pull/105 that it will add a maven vs. gradle switch, so that we can vary what we seed the image ref array with ?

gabemontero (commented):
some initial thoughts during our call where we were discussing @dwalluck 's changes

stuartwdouglas:
@dwalluck I have pushed a commit that allows for building both maven and gradle builds. I don't see that problem you reported, everything seemed to be working fine for me.

pkg/reconciler/dependencybuild/dependencybuild.go
gabemontero: yeah we discussed this last night after I posted this comment and determined the configmap would not be the path here

java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/build/BuildRecipeInfo.java
```diff
@@ -16,6 +16,8 @@ public class BuildRecipeInfo {
     boolean enforceVersion;
     List<String> additionalArgs = new ArrayList<>();
     List<String> ignoredArtifacts = new ArrayList<>();
+    String toolVersion;
+    String javaHome;
```
stuartwdouglas, line 20: This is not the right place for javaHome, this is something that depends on the builder image. This should just be a version and something else will resolve it into a path.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/BuildInfo.java
```diff
@@ -12,24 +12,34 @@ public class BuildInfo {
...
     List<String> ignoredArtifacts = new ArrayList<>();
 
     public Map<String, VersionRange> getTools() {
         return tools;
     }
 
+    /**
+     * Version of the build tool.
+     */
+    String toolVersion;
```
stuartwdouglas, line 36: This info is already there, in the VersionRange of the tools map.

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -247,27 +249,34 @@ func (r *ReconcileDependencyBuild) handleStateAnalyzeBuild(ctx context.Context,
...
+		if maven {
+			for _, image := range []string{"quay.io/sdouglas/hacbs-jdk11-builder:latest", "quay.io/sdouglas/hacbs-jdk8-builder:latest", "quay.io/sdouglas/hacbs-jdk17-builder:latest"} {
+				for _, command := range unmarshalled.Invocations {
+					buildRecipes = append(buildRecipes, &v1alpha1.BuildRecipe{Task: "run-maven-component-build", Image: image, CommandLine: command, EnforceVersion: unmarshalled.EnforceVersion, IgnoredArtifacts: unmarshalled.IgnoredArtifacts, ToolVersion: unmarshalled.ToolVersion, JavaHome: unmarshalled.JavaHome})
 				}
-				if gradle {
-					buildRecipes = append(buildRecipes, &v1alpha1.BuildRecipe{Task: "run-gradle-component-build", Image: image, CommandLine: command, EnforceVersion: unmarshalled.EnforceVersion, IgnoredArtifacts: unmarshalled.IgnoredArtifacts})
+			}
+		} else if gradle {
+			for _, image := range []string{"quay.io/dwalluck/gradle:latest"} {
```
stuartwdouglas, line 273: Those existing builder images also have gradle installed. Now the builder PR is in we need to look at how they are resolved.

java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/build/BuildRecipeInfo.java
dwalluck: OK, I will change `JAVA_HOME` to `JAVA_VERSION`.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/BuildInfo.java
dwalluck: I saw I could set/use the preferred version from the version range, but this variable was more about how to pass that as a parameter to the yaml file.

pkg/reconciler/dependencybuild/dependencybuild.go
stuartwdouglas: I don't think we will have all the gradle versions in a builder image, although this may change over time. The selection logic should not know about what images are available, it should just report on what it found. Whatever code we end up with that understands what is in the builder images will then attempt to map this to an image + version, trying to get as close as possible.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/BuildInfo.java
stuartwdouglas: We are going to need some logic in the reconciler that maps what was discovered to the builder images that are available. I am planning on looking at this once this PR is in.

deploy/base/gradle-v0.1.yaml
```diff
@@ -0,0 +1,352 @@
...
+        export LANG=en_US.UTF-8
+        export LC_ALL=en_US.UTF-8
+
+        if [ -n "$(params.ENFORCE_VERSION)" ]; then
+            gradle-manipulator -l "${GRADLE_HOME}" $(params.GRADLE_MANIPULATOR_ARGS) -DversionOverride=$(params.ENFORCE_VERSION) generateAlignmentMetadata || exit 1
+        else
+            gradle-manipulator -l "${GRADLE_HOME}" $(params.GRADLE_MANIPULATOR_ARGS) generateAlignmentMetadata || exit 1
+        fi
+
+        gradle -DAProxDeployUrl=file:$(workspaces.source.path)/hacbs-jvm-deployment-repo $@ || exit 1
```
stuartwdouglas, line 277: So this is supposed to be the 'user activated' task, it is basically an alternative to the tasks in https://github.com/redhat-appstudio/build-definitions. It should not be deploying the result to the sidecar, rather the real tasks will create a runnable application. I don't want to hold the PR up over this, as it is not really on the critical path at all, but if we actually want to support user level gradle builds then this will need some work and a corresponding build in the build-definitions repo.

deploy/base/kustomization.yaml
```diff
@@ -2,10 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
```
stuartwdouglas, line 7: This needs to be changed back before merge.

deploy/base/lookup-artifact-location-v0.1.yaml
```diff
@@ -17,12 +17,12 @@ spec:
       type: string
       description: >-
         The image to use for dependency analysis.
-      default: "quay.io/sdouglas/hacbs-jvm-build-request-processor:dev"
+      default: "quay.io/dwalluck/hacbs-jvm-build-request-processor:dev"
```
stuartwdouglas, line 20: This should not have been necessary, the value here is replaced by the dev-template generated overlay.

deploy/base/run-maven-component-build-v0.1.yaml
```diff
@@ -1,266 +0,0 @@
-apiVersion: tekton.dev/v1beta1
-kind: ClusterTask
```
stuartwdouglas, line 2: We can't merge with this deletion until https://github.com/redhat-appstudio/infra-deployments/pull/525 is also merged.

java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/build/BuildRecipeInfo.java
stuartwdouglas: This would allow you to explicitly specify which JDK you needed to use. At the moment we just try all three until one works, when it gets a bit smarter it will try and infer which ones to use.

pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
```diff
@@ -74,4 +74,6 @@ type BuildRecipe struct {
 	CommandLine      []string `json:"commandLine,omitempty"`
 	EnforceVersion   string   `json:"enforceVersion,omitempty"`
 	IgnoredArtifacts []string `json:"ignoredArtifacts,omitempty"`
+	ToolVersion      string   `json:"toolVersion,omitempty"`
+	JavaHome         string   `json:"javaHome,omitempty"`
```
stuartwdouglas, line 78: This should be JavaVersion.

stuartwdouglas (changes requested):
I think this is really close to being ready to go in as an initial version. Once it is in we can iterate from there.

java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/build/BuildRecipeInfo.java
dwalluck: If it's in the image name itself, as in the case of your images, then for `quay.io/sdouglas/hacbs-jdk8-builder:latest` I could do `"quay.io/sdouglas/hacbs-jdk" + javaVersion + "-builder:latest"` directly in the code and I don't even need it in the yaml. But with my current single image setup, I do.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/BuildInfo.java
dwalluck: OK to leave it in for now? It currently does set `preferredVersion` and `toolVersion` to the same value.

deploy/base/gradle-v0.1.yaml
dwalluck: You can view `gradle -DAProxDeployUrl=file:$(workspaces.source.path)/hacbs-jvm-deployment-repo publish` as equivalent to `mvn -DaltDeploymentRepository=local::file:$(workspaces.source.path)/hacbs-jvm-deployment-repo deploy`.

The equivalent line is still in the Maven version:

```
/usr/bin/mvn -s "$(workspaces.maven-settings.path)/settings.xml" $@ "-DaltDeploymentRepository=local::file:$(workspaces.source.path)/hacbs-jvm-deployment-repo" "org.apache.maven.plugins:maven-deploy-plugin:3.0.0-M2:deploy" 
```

gabemontero:
OK to add an e2e running against openshift @dwalluck @stuartwdouglas (which I would like to say ultimately needs to be part of this PR, where its passing is a requisite for merging) could be devised in such a way

- copy https://github.com/redhat-appstudio/jvm-build-service/blob/main/openshift-with-appstudio-test/e2e/jvm_build_service_test.go#L244-L410 into a new `TestExampleGradleRun` (where for now at least you bypass the contaminant test @stuartwdouglas recently added at the bottome there)
- or perhaps create functions around the existing code that are called from `TestExampleRun` and `TestExampleGradleRun` 
- change https://github.com/redhat-appstudio/jvm-build-service/blob/main/openshift-with-appstudio-test/e2e/jvm_build_service_test.go#L257 in the copy to access https://github.com/redhat-appstudio/jvm-build-service/pull/109/files#diff-c001d9b1e1e717567e68428caf28d58abe79be7230d646456a5a3b083b109fab instead of maven
- change https://github.com/redhat-appstudio/jvm-build-service/blob/main/openshift-with-appstudio-test/e2e/jvm_build_service_test.go#L293  to access https://github.com/redhat-appstudio/jvm-build-service/pull/109/files#diff-39e308d4cde004681f832ebaff6b70a1853beb44ce6ab4aef4e5b08da686430e
- change https://github.com/redhat-appstudio/jvm-build-service/blob/main/openshift-with-appstudio-test/e2e/jvm_build_service_test.go#L305 to point to https://github.com/redhat-appstudio/jvm-build-service/pull/109/files#diff-ca9c6c5a4345787ab83f5f92bac292efd87469c2290e6fc02c7ce9d1b1242a60

I'll ping you in slack in a bit @dwalluck and among other things we can talk about whether you want to take this on from the start, or if I can "bootstrap" you in one way or the other.

pkg/reconciler/dependencybuild/dependencybuild.go
dwalluck: Right now, I've _hardcoded_ the available Gradle versions from the image in the Java code, but I've added a method to map the discovered Gradle version to the nearest version (less than or equal) to the found version.

java-components/pom.xml
```diff
@@ -25,11 +25,12 @@
         <formatter-maven-plugin.version>2.18.0</formatter-maven-plugin.version>
         <impsort-maven-plugin.version>1.6.2</impsort-maven-plugin.version>
 
+        <assertj.version>3.23.1</assertj.version>
         <aws-url-connection-client.version>2.17.95</aws-url-connection-client.version>
         <aws-java-sdk-s3.version>1.12.169</aws-java-sdk-s3.version>
         <quarkus-jgit.version>1.2.0</quarkus-jgit.version>
         <quarkus-operator-sdk.version>3.0.7</quarkus-operator-sdk.version>
-        <gradle-tooling-api.version>7.4.2</gradle-tooling-api.version>
+        <gradle-tooling-api.version>7.3-20210825160000+0000</gradle-tooling-api.version>
```
dwalluck, line 33: I don't understand what repository the 7.4.2 version came from. I don't see it on Maven central.

stuartwdouglas: https://github.com/stuartwdouglas/jvm-build-service/blob/940aef09401c286e8658691c2dca6bc379a48a48/java-components/adhoc-analyser-tool/pom.xml#L169

They stopped publishing to Maven central so you have to use their repo.

dwalluck: OK, it seems that the repository just needs to be moved to `java-components/pom.xml`, then.

deploy/base/kustomization.yaml
dwalluck: Reverted back to upstream. It works now.

deploy/base/kustomization.yaml
dwalluck: Done.

deploy/base/lookup-artifact-location-v0.1.yaml
dwalluck: Done.

pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
dwalluck: Done. Switched from `JavaHome` to `JavaVersion`. It will still set `JAVA_HOME` from the yaml using that version.

java-components/pom.xml
dwalluck: Moved the declaration and updated to Gradle 7.5 in[12baa22](https://github.com/redhat-appstudio/jvm-build-service/pull/109/commits/12baa22afbea207b1a7727a3197746614c86b2fb).

deploy/crds/base/jvmbuildservice.io_dependencybuilds.yaml
```diff
@@ -159,10 +159,14 @@ spec:
                     type: array
                   image:
                     type: string
+                  javaVersion:
```
gabemontero, line 162: just making sure - you got this change by running `make generate` right @dwalluck 

thanks

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -8,23 +8,6 @@ kind: TaskRun
 metadata:
   name: run-maven-component-build
 spec:
-  params:
-  - name: URL
```
gabemontero, line 12: I don't see why you should be deleting these params

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -8,23 +8,6 @@ kind: TaskRun
...
-    value: 1.10.0
-  - name: CONTEXT_DIR
-    value: ""
-  - name: IMAGE
-    value: quay.io/sdouglas/hacbs-jdk11-builder:latest
-  - name: GOALS
-    value:
-    - build
-  workspaces:
-  - emptyDir: {}
```
gabemontero, line 24: same with the workspaces

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -33,6 +16,14 @@ spec:
       - name: maven-settings
       - name: source
     params:
+      - name: JAVA_VERSION
```
gabemontero, line 19: it would make sense to me to add these params after `GOALS` once you add those back

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -246,27 +248,34 @@ func (r *ReconcileDependencyBuild) handleStateAnalyzeBuild(ctx context.Context,
...
 			r.eventRecorder.Eventf(&db, v1.EventTypeWarning, "InvalidJson", "Failed to unmarshal build info for AB %s/%s JSON: %s", db.Namespace, db.Name, buildInfo)
 			return reconcile.Result{}, err
 		}
-		//for now we are ignoring the tool versions
-		//and just using the supplied invocations
+		// for now we are ignoring the tool versions
+		// and just using the supplied invocations
 		buildRecipes := []*v1alpha1.BuildRecipe{}
-		_, maven := unmarshalled.Tools["maven"]
+		_, maven := unmarshalled.Tools["mvn"]
```
gabemontero, line 262: can you clarify why this change is needed .... feels like it would break existing assumptions, no?

does this make sense to you @stuartwdouglas ?  ... maybe I am missing something elsewhere that necessitates this.

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -246,27 +248,34 @@ func (r *ReconcileDependencyBuild) handleStateAnalyzeBuild(ctx context.Context,
...
+
+		if maven {
+			for _, image := range []string{"quay.io/sdouglas/hacbs-jdk11-builder:latest", "quay.io/sdouglas/hacbs-jdk8-builder:latest", "quay.io/sdouglas/hacbs-jdk17-builder:latest"} {
+				for _, command := range unmarshalled.Invocations {
+					buildRecipes = append(buildRecipes, &v1alpha1.BuildRecipe{Image: image, CommandLine: command, EnforceVersion: unmarshalled.EnforceVersion, IgnoredArtifacts: unmarshalled.IgnoredArtifacts, ToolVersion: unmarshalled.ToolVersion, JavaVersion: unmarshalled.JavaVersion, Maven: true})
 				}
-				if gradle {
-					buildRecipes = append(buildRecipes, &v1alpha1.BuildRecipe{Gradle: true, Image: image, CommandLine: command, EnforceVersion: unmarshalled.EnforceVersion, IgnoredArtifacts: unmarshalled.IgnoredArtifacts})
+			}
+		} else if gradle {
```
gabemontero, line 271: yeah we don't like this `else if` stuff in golang (I had to learn this same lesson in 2015 when I left the java world and started working on openshift).

switch this and the `if maven` above to 

```
switch {
case maven:
<add the code for maven>
case gradle:
<add the code for gradle>
default:
return an error cause something is wrong with our base setup
}
```

gabemontero (changes requested):
Several things I think we minimally need to change @dwalluck @stuartwdouglas before merging.
I think a few are rebase related and we discussed on our video conf earlier.

I could not gleam @dwalluck via PR review why the unit tests and `test/e2e` / integration tests are failing for you.  Hopefully by retry and debug you can sort it out.  If not, I'll check out this PR and see if I can help.

Plus, I now finally caught up and see this PR has 25 commits at the moment.  That exceeds at least my threshold :-) .... when we finally sort out this PR and think it is ready for merge, I at least would prefer to squash these at least somewhat.

@stuartwdouglas @dwalluck do you all feel strongly against that ?

deploy/base/maven-v0.2.yaml
```diff
@@ -93,7 +93,6 @@ spec:
         runAsUser: 0
       script: |
         #!/usr/bin/env bash
-        chown 1001:1001 -R $(workspaces.source.path)
```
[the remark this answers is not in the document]
dwalluck: There were two lines that were exactly the same. This just removes one of them.

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -243,10 +245,10 @@ func streamFileYamlToTektonObj(path string, obj runtime.Object, ta *testArgs) ru
 
 func TestExampleRun(t *testing.T) {
 	ta := setup(t, nil)
-	//TODO, for now at least, keeping our test project to allow for analyzing the various CRD instances both for failure
+	// TODO, for now at least, keeping our test project to allow for analyzing the various CRD instances both for failure
```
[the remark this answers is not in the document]
dwalluck: This is some kind of automatic reformatting from IntelliJ IDEA. I am not sure if I accidentally turned something on. I can revert this by repatching ignoring whitespace.

However, in this case, it is claiming that a comment without a space after it is "illegal" in Go, even though the compiler doesn't seem to be complaining.

deploy/base/maven-v0.2.yaml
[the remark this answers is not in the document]
gabemontero: ah sorry see the one below now :-)

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -14,9 +14,10 @@ import (
 	"time"
 
 	projectv1 "github.com/openshift/api/project/v1"
-	"github.com/redhat-appstudio/jvm-build-service/pkg/apis/jvmbuildservice/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 
+	"github.com/redhat-appstudio/jvm-build-service/pkg/apis/jvmbuildservice/v1alpha1"
```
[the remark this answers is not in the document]
dwalluck: I think My IDE is automatically moving it for some reason. Reverted in 0c7bef5.

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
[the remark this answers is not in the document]
dwalluck: Reverted in c982815.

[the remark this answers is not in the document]
gabemontero: well, I would believe goland before intellij on that :-) but I'll defer on this

marking resolved

pkg/reconciler/dependencybuild/dependencybuild.go
dwalluck: It looks like this change was made by @stuartwdouglas in 0530131c. I wonder if we wanted to change "maven" to "mvn" to match the actual name of the binary? This would require changing the JSON annotation or field name in the Java code as well which generates this.

deploy/base/run-maven-component-build-v0.1.yaml
dwalluck: I would have guessed that it would have been part of whatever pull request inlined the yaml. I am not sure how it ended up here.

pkg/reconciler/dependencybuild/dependencybuild.go
dwalluck: Done in f3b832f.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
dwalluck: This should be fixed in 613a2aa.

deploy/crds/base/jvmbuildservice.io_dependencybuilds.yaml
dwalluck: Yes. Same for the `config/kcp/*` changes.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -19,7 +19,14 @@ spec:
...
     value:
-    - build
+      - -DskipTests
+      - clean
+      - install
+      - -Denforcer.skip
+  - name: JAVA_VERSION
+    value: ""
+  - name: TOOL_VERSION
+    value: ""
```
dwalluck, line 29: @stuartwdouglas Should I be adding default empty values for Maven. The file `deploy/base/maven-v0.2.yaml` currently doesn't have that added.

If I understand correctly, `TOOL_VERSION` is going away, but `JAVA_VERSION` may be useful for the Maven case, also?

pkg/reconciler/dependencybuild/buildrecipeyaml.go
dwalluck: Done.

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -246,27 +248,39 @@ func (r *ReconcileDependencyBuild) handleStateAnalyzeBuild(ctx context.Context,
...
+		// and just using the supplied invocations
 		buildRecipes := []*v1alpha1.BuildRecipe{}
+		// TODO: Change this to "mvn"? Requires more changes, to the Go test and the Java JSON model
 		_, maven := unmarshalled.Tools["maven"]
 		_, gradle := unmarshalled.Tools["gradle"]
-		for _, image := range []string{"quay.io/sdouglas/hacbs-jdk11-builder:latest", "quay.io/sdouglas/hacbs-jdk8-builder:latest", "quay.io/sdouglas/hacbs-jdk17-builder:latest"} {
-			for _, command := range unmarshalled.Invocations {
-				if maven {
-					buildRecipes = append(buildRecipes, &v1alpha1.BuildRecipe{Maven: true, Image: image, CommandLine: command, EnforceVersion: unmarshalled.EnforceVersion, IgnoredArtifacts: unmarshalled.IgnoredArtifacts})
+
```
dwalluck, line 265: @stuartwdouglas You had changed this to "mvn" in a commit. In my opinion, it may make more sense to name it that way, but it requires other changes to make it work.

deploy/base/run-maven-component-build-v0.1.yaml
gabemontero: Sorry @dwalluck from my point of view how it ended up here is a bit irrelevant at this point.

For me at least, per my earlier comment around the number of commits you now have, I think at this point it is time to 
- squash your now 32 commits locally via `git rebase main -i` 
- `git reset HEAD`
-  'git checkout deploy/base/run-gradle-component-build-v0.1.yaml`
- `git checkout deploy/base/run-maven-component-build-v0.1.yaml`
- `git commit -a -m "rebuild java dependencies that use gradle"
- `git push -f origin PLNSRVCE-346`

And then we will handle these deletes via a PR and commit message that conveys they are being deleted.

But if I'm in the minority on this opinion @dwalluck @stuartwdouglas, based on the reasoning why,  I could be convinced to defer to you two.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
gabemontero: my item from the last review was addressed @dwalluck I'll resolve this thanks

deploy/crds/base/jvmbuildservice.io_dependencybuilds.yaml
gabemontero: great thanks for confirming

marking resolved

pkg/reconciler/dependencybuild/buildrecipeyaml.go
gabemontero: aside from adding TOOL_VERSION and JAVA_VERSION, you added entries to the GOALS param array

technically speaking changing an existing param on the maven recipe would not be part of adding gradle

but I'll defer to @stuartwdouglas - are you good with that aspect of this change as well ?

pkg/reconciler/dependencybuild/dependencybuild.go
gabemontero: looks good @dwalluck thanks ... marking resolved

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -486,3 +488,249 @@ func TestExampleRun(t *testing.T) {
 		}
 	})
 }
+
+func TestExampleGradleRun(t *testing.T) {
```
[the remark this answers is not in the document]
gabemontero: so your e2e test failed @dwalluck 

If you open up the "Details" link next to `ci/prow/jvm-build-service-in-repo-e2e — Job failed. ` down at the bottom here and then click on the `Raw build-log.txt` link right under the big `Build Log` heading, that log has a print of the cluster setup followed by the invocation of the tests, and then our debug when it fails.  

In that debug is a pod/container dump that has the string `pod logs for container step-gradle-tasks in pod sample-component-build-rvgcr-gradle-run-pod` in it.

there is a really big exception there.  And then it looks like the reset of the containers in that pod are skipped.  And I see no other pods after (if you were expecting it to retry a different combo of java/gradle etc. that did not occur).

Give investigating that log a try and see what kind of progress you can make.

FYI 3 of the 4 maven tests pass.  And the one that failed, `TestExampleRun/contaminated_build_is_resolved ` is the one @stuartwdouglas just fixed overnight for us.

My guess is that if you rebase this PR against main again you'll pick up that fix.  Maybe something to try along with your squashing of the commit I asked for, assuming you agree it is time for that.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
dwalluck: @gabemontero I believe it was your original commit when `pkg/reconciler/dependencybuild/buildrecipeyaml.go` was created that had Maven `GOALS` set to `- build` which can't be right since that's a Gradle task. Correct me if I'm wrong.

I tried updating it in my pull request to match the existing maven goals. I don't know if my change is correct, but the original upstream was not correct either.

I think a couple of these maven changes probably belong in their own pull request anyway.

gabemontero: ah you are right @dwalluck now that I look at main brainch

but it shows up as a diff in the PR view ... again, you branch is a bit behind main and that is confusing me.  4 commits behind based on what https://github.com/dwalluck/jvm-build-service/tree/PLNSRVCE-346 tells me.

yeah per my recent comments elsewhere, I think some commit squashing, and another rebase, are in order.

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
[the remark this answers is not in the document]
gabemontero: small clarification, one of the gradle tests passed, but then the second one failed.

```
--- FAIL: TestExampleGradleRun (603.15s)
    --- PASS: TestExampleGradleRun/pipelinerun_completes_successfully (0.07s)
    --- FAIL: TestExampleGradleRun/artifactbuilds_and_dependencybuilds_generated (601.14s)
```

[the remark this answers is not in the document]
gabemontero: and now that I see that no artifactbuilds or dependencybuilds were generated, that exception I was referring to in https://github.com/redhat-appstudio/jvm-build-service/pull/109#discussion_r933553687 I think makes more sense about being the "root cause"

vs. artifactbuilds and  dependencybuilds getting generated and certain recipe attempts failing

deploy/base/run-maven-component-build-v0.1.yaml
gabemontero: and just remembered / rediscoverd - @stuartwdouglas already has a yaml cleanup PR that is taking care of this file:  https://github.com/redhat-appstudio/jvm-build-service/pull/136

it got hit by some of the bugs that between he and I we diagnosed / fixed

I will see if I can rebase that, getting it passing e2e's and merged

dwalluck:
> @stuartwdouglas @dwalluck do you all feel strongly against that ?

I am OK with squashing. Or it can also be squashed on merge, too. @stuartwdouglas?

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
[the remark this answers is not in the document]
dwalluck: I saw one of the build logs it fails to download/find one of the plugins. The repository used is `gradlePluginsPortal()`. It should have it.

 * Where:
        Build file '/workspace/source/subproject3/build.gradle' line: 2
        
        * What went wrong:
        Plugin [id: 'com.github.johnrengelman.shadow', version: '5.2.0'] was not found in any of the following sources:
        
        - Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
        - Plugin Repositories (could not resolve plugin artifact 'com.github.johnrengelman.shadow:com.github.johnrengelman.shadow.gradle.plugin:5.2.0')
          Searched in the following repositories:
            MavenLocal(file:/root/.m2/repository/)
            hacbs-mvn(http://localhost:2000/maven2)
            Gradle Central Plugin Repository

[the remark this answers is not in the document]
dwalluck: https://plugins.gradle.org/plugin/com.github.johnrengelman.shadow/5.2.0
url "https://plugins.gradle.org/m2/"

pkg/reconciler/dependencybuild/buildrecipeyaml.go
stuartwdouglas: I am fine with these for now in terms of just getting this feature in, but I do want to revisit this in the future. I need to capture all this in a JIRA but basically what I am thinking is:
- General purpose builder images with both maven and gradle
- Programatically assemble these tasks keeping them as close as possible (so we can re-use code between both)
- System wide config as to which images have which versions, and we use this to map builder images to requests
- The actual tool and command line to be selected programatically and set as parameters, rather than something hard coded into the image.

We are not really in a position to start on this yet though, we need this in and my config PR in and then we can begin to look at how this will all fit together.

pkg/reconciler/dependencybuild/dependencybuild.go
stuartwdouglas: As I said above lets just get this in and refine it later.

stuartwdouglas:
@dwalluck @gabemontero the new gradle tests failed, but IMHO they are testing something we are not really interested in. They are using the gradle sample task to build a user application using gradle (https://github.com/redhat-appstudio/jvm-build-service/pull/109/files#diff-c001d9b1e1e717567e68428caf28d58abe79be7230d646456a5a3b083b109fabR4) which is not really something we care about. At this state we only care about rebuilding dependencies built with gradle, so we should just be make sure our sample project has a gradle based dependency, and verify that is is complete. 

We actually already have a gradle dependency in our sample application (https://github.com/stuartwdouglas/shaded-java-app/blob/main/pom.xml#L23), so initially all we really need to do is update the existing test to make sure that the ASM build completes successfully. 

Note that the corresponding 'maven-v0.2.yaml' is not actually used by HACBS, only for our local development and testing. HACBS uses a different build defined in 'build-definitions' (https://github.com/redhat-appstudio/build-definitions/blob/main/tasks/patches/add-sbom-and-push.yaml#L156)

gabemontero:
> @dwalluck @gabemontero the new gradle tests failed, but IMHO they are testing something we are not really interested in. They are using the gradle sample task to build a user application using gradle (https://github.com/redhat-appstudio/jvm-build-service/pull/109/files#diff-c001d9b1e1e717567e68428caf28d58abe79be7230d646456a5a3b083b109fabR4) which is not really something we care about. At this state we only care about rebuilding dependencies built with gradle, so we should just be make sure our sample project has a gradle based dependency, and verify that is is complete.
> 
> We actually already have a gradle dependency in our sample application (https://github.com/stuartwdouglas/shaded-java-app/blob/main/pom.xml#L23), so initially all we really need to do is update the existing test to make sure that the ASM build completes successfully.

Sorry, this is probably obvious to you @stuartwdouglas and @dwalluck but can you clarify which thing exactly is the "ASM build"?

Otherwise, based on your explanation, sure, let's pivot in that direction and ditch the gradle test I steered @dwalluck towards.

> 
> Note that the corresponding 'maven-v0.2.yaml' is not actually used by HACBS, only for our local development and testing. HACBS uses a different build defined in 'build-definitions' (https://github.com/redhat-appstudio/build-definitions/blob/main/tasks/patches/add-sbom-and-push.yaml#L156)

yep and @psturc is working the story I opened last week, https://issues.redhat.com/browse/PLNSRVCE-533, to take the next step and use the build defined in build-definitions.  Full transparency, I was uncertain when I talked to @psturc if it made sense to have both forms of builds, and in fact steered him toward verifying that a component/build-definition based build used dependencies stored in the maven cache that resulted from running `maven-v0.2.yaml`.  

Perhaps we should alter that path?

stuartwdouglas:
ASM is a Java library build with Gradle that is in the e2e tests: https://github.com/stuartwdouglas/shaded-java-app/blob/main/pom.xml#L24

The only issue I see with using the HACBS build is that is has a lot of 'extra' stuff related to building an application that is not really related to the JVM build service (e.g. security scanning, deploying the resulting images etc). 
The test already attempts to build it, but it just verifies that at least one build passes, we can probably not check that all of them pass and that would verify the basics.

gabemontero:
rebase / squash looks OK to me @dwalluck 

and per that last couple of comments here that addition gradle test is removed

let's see how the existing e2e does 

then based on that, we can see whether or not to update the existing e2e to either confirm all dependencybuilds complete, or if at least the `asm` one does

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -404,12 +404,14 @@ func TestExampleRun(t *testing.T) {
...
 				}
 			}
-			dbComplete := false
 			dbList, err := jvmClient.JvmbuildserviceV1alpha1().DependencyBuilds(ta.ns).List(context.TODO(), metav1.ListOptions{})
 			ta.Logf(fmt.Sprintf("number of dependencybuilds: %d", len(dbList.Items)))
+			dbComplete := true
 			for _, db := range dbList.Items {
-				if db.Status.State == v1alpha1.DependencyBuildStateComplete {
-					dbComplete = true
+				if db.Status.State != v1alpha1.DependencyBuildStateComplete {
```
gabemontero, line 411: cool deal @dwalluck 

let's see how the in flight e2e test does

@stuartwdouglas FYI

gabemontero:
app studio bootstrap fail in jvm-build-service-e2e https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/109/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-e2e/1554166743352479744

/test jvm-build-service-e2e

stuartwdouglas:
I did not realise that tcnative was being brought in as a transitive dependency, it won't build until we have our new builder images. I have pushed an exclusion to the git repo.

gabemontero:
aws cluster install error, let's try again

/retest

PR #110, opened 2022-07-19
PLNSRVCE-455: Use hard coded task spec

stuartwdouglas:
This changed the lookup build info task to use a hard coded spec.

If image overrides are required these are now done as environment
variables in the operator.

### DIFF
--- a/deploy/operator/overlays/dev-template/kustomization.yaml
+++ b/deploy/operator/overlays/dev-template/kustomization.yaml
@@ -8,3 +8,6 @@ images:
   - name: hacbs-jvm-operator
     newName: quay.io/QUAY_USERNAME/hacbs-jvm-controller
     newTag: dev
+
+patchesStrategicMerge:
+  - operator-images.yaml
--- a/deploy/operator/base/deployment.yaml
+++ b/deploy/operator/base/deployment.yaml
@@ -18,14 +18,8 @@ spec:
           image: hacbs-jvm-operator
           imagePullPolicy: Always
           env:
-            - name: QUARKUS_LOG_LEVEL
-              value: DEBUG
-          # livenessProbe:
-          #   httpGet:
-          #     path: /q/health/live
-          #     port: 8080
-          #   initialDelaySeconds: 10
-          #   periodSeconds: 10
+            - name: RECIPE_DATABASE
+              value: "https://github.com/stuartwdouglas/tempdata2.git"
           resources:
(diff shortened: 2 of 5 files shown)


deploy/operator/overlays/dev-template/kustomization.yaml
```diff
@@ -8,3 +8,14 @@ images:
   - name: hacbs-jvm-operator
     newName: quay.io/QUAY_USERNAME/hacbs-jvm-controller
     newTag: dev
+
+patches:
+  - patch: |-
+      - op: add
+        path: /spec/template/spec/containers/0/env/-
+        value:
+          name: BUILD_REQUEST_PROCESSOR_IMAGE
```
gabemontero, line 17: So changing the name to `JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE` would allow you to already tie in with was our openshift CI integration is doing on PR e2e invocations @stuartwdouglas per https://github.com/redhat-appstudio/jvm-build-service/blob/main/.ci/common-ci-settings-and-functions.sh#L52

deploy/base/lookup-artifact-location-v0.1.yaml
```diff
@@ -1,59 +0,0 @@
-apiVersion: tekton.dev/v1beta1
```
gabemontero, line 1: I don't think we can delete this out of the gate because of https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build/jvm-build-service/kustomization.yaml#L7

I think you have to do the hardcoding of the task like you have in the operator code, but leave this file, merge this PR, and then create a infra-deployments PR that removes https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build/jvm-build-service/kustomization.yaml#L7 and https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/preview.sh#L98-L99

You would also want to add logic in preview.sh that updates the operator deployment to add the env var setting on the deployment based on the presence of `JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE` etc.

After those merge, you can come back to this repo and delete this file.

gabemontero:
With https://github.com/redhat-appstudio/infra-deployments/pull/506#issuecomment-1190718614 I'm good merging this.  It will make additional verification of https://github.com/redhat-appstudio/infra-deployments/pull/506 after its rebase / possible pull in of suggestions.

PR #112, opened 2022-07-19
timeout increases, more debug, based on openshift/release rehearsal results

gabemontero:
see https://github.com/openshift/release/pull/30600


gabemontero:
@stuartwdouglas FYI

PR #114, opened 2022-07-20
Move build info lookup to artifact location file

stuartwdouglas:
Both these files are going to be deleted anyway, but
the lookup-build-info-v0.1.yaml has not been added to
infra deployments yet. By just putting both tasks in
the same file it saves an additional update in infra
deployments of a file that is going away anyway.


stuartwdouglas:
The other approach would be to manually update this PR before it is merged to include the new file, but as it is going away anyway I think this approach is quicker/simpler. 

Alternativly if both https://github.com/redhat-appstudio/infra-deployments/pull/506 and https://github.com/redhat-appstudio/jvm-build-service/pull/110 go in then this file is no longer needed and can be safely deleted.

gabemontero:
yep this will make getting the e2e's passing on the latest commits of this repo a few steps easier thanks @stuartwdouglas :-)

/approve
/lgtm

PR #115, opened 2022-07-20
If an update fails attempt to requeue it

stuartwdouglas:


pkg/reconciler/artifactbuild/artifactbuild.go
```diff
@@ -374,7 +374,7 @@ func (r *ReconcileArtifactBuild) handleStateComplete(ctx context.Context, abr *v
 				db.Status.State = v1alpha1.DependencyBuildStateNew
 			}
 			if err := r.client.Status().Update(ctx, &db); err != nil {
-				return reconcile.Result{}, err
+				return reconcile.Result{Requeue: true}, err
```
gabemontero, line 377: gotta admit I always have to revisit this with controller runtime ... I thought returning an error resulted in a retry of the event anyway ... you are seeing that is not the case @stuartwdouglas ?

stuartwdouglas: Hmm, ok, maybe something else is causing it then, I am just seeing artifacts sometimes get stuck in contaminated and this being fixed once the operator restarts. I will do some more debugging.

gabemontero: ok sounds good

gabemontero: it may also be that just returning the error means controller runtime retries up to its configured limit, and then gives up ... have vague recollections of that but I could be wrong (would have to dive into the controller runtime code to confirm - if I have time for that near term I'll try to look).

perhaps you were running into a situation where you got enough update conflicts that you hit this .... anyway perhaps some useful food for thought before you dive into the next round of debug

PR #118, opened 2022-07-20
PLNSRVCE-455: convert maven/gradle cluster tasks into hard coded task specs

gabemontero:
ok dinner time came upon me @stuartwdouglas 

I can certainly pick up on this on Thursday.  But if you get the urge to take this in progress commit and run with it, go for it, but by all means, with all the other stuff we need to dive into, certainly feel free to go elsewhere.

### DIFF
--- a/pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
+++ b/pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
@@ -66,7 +66,10 @@ type DependencyBuildList struct {
 }
 
 type BuildRecipe struct {
+	//deprecated task string
 	Task             string   `json:"task,omitempty"`
+	Maven            bool     `json:"maven,omitempty"`
+	Gradle           bool     `json:"gradle,omitempty"`
 	Image            string   `json:"image,omitempty"`
 	CommandLine      []string `json:"commandLine,omitempty"`
 	EnforceVersion   string   `json:"enforceVersion,omitempty"`
--- a/pkg/reconciler/dependencybuild/buildrecipeyaml.go
+++ b/pkg/reconciler/dependencybuild/buildrecipeyaml.go
@@ -1,0 +1,481 @@
+package dependencybuild
+
+var (
+	// since taskspec does not implment runtime.Object we have to mimic the TaskRun created in dependencybuilds
+	maven = `
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: run-maven-component-build
+spec:
+  params:
+  - name: URL
+    value: https://github.com/smallrye/smallrye-common.git
+  - name: TAG
+    value: 1.10.0
+  - name: CONTEXT_DIR
+    value: ""
+  - name: IMAGE
+    value: quay.io/sdouglas/hacbs-jdk11-builder:latest
+  - name: GOALS
+    value:
+    - build
+  workspaces:
+  - emptyDir: {}
+    name: maven-settings
+  - emptyDir: {}
+    name: source
+  taskSpec:
+    results:
+      - name: contaminants
+        description: Any community GAVs that have ended up in the final output.
+    workspaces:
+      - name: maven-settings
+        mountPath: $(workspaces.maven-settings.path)
+      - name: source
+        mountPath: $(workspaces.source.path)
+    params:
+      - name: URL
+        type: string
+      - name: TAG
+        type: string
+      - name: IMAGE
+        type: string
+        description: Build image
+      - name: GOALS
+        description: maven goals to run
+        type: array
+        default:
+          - -DskipTests
+          - clean
+          - install
+          - -Denforcer.skip
+      - name: MAVEN_MIRROR_URL
+        description: The Maven repository mirror url
+        type: string
+        default: http://localhost:2000/maven2
+      - name: SERVER_USER
+        description: The username for the server
+        type: string
+        default: ""
+      - name: SERVER_PASSWORD
+        description: The password for the server
+        type: string
+        default: ""
+      - name: PROXY_USER
+        description: The username for the proxy server
+        type: string
+        default: ""
+      - name: PROXY_PASSWORD
+        description: The password for the proxy server
+        type: string
+        default: ""
+      - name: PROXY_PORT
+        description: Port number for the proxy server
+        type: string
+        default: ""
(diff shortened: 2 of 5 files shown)


pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
```diff
@@ -66,7 +66,10 @@ type DependencyBuildList struct {
 }
 
 type BuildRecipe struct {
+	//deprecated task string
```
gabemontero, line 69: will remove this field when we come back and remove the maven/gradlecluster task files

gabemontero:
@dwalluck FYI this PR or immediate follow ons will also in-line / replace the gradle cluster task as well.

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -309,6 +313,262 @@ func (r *ReconcileDependencyBuild) handleStateBuilding(ctx context.Context, db *
 	tr.Labels = map[string]string{artifactbuild.DependencyBuildIdLabel: db.Labels[artifactbuild.DependencyBuildIdLabel], artifactbuild.TaskRunLabel: "", TaskType: TaskTypeBuild}
 
 	tr.Spec.TaskRef = &pipelinev1beta1.TaskRef{Name: db.Status.CurrentBuildRecipe.Task, Kind: pipelinev1beta1.ClusterTaskKind}
+	switch {
+	case db.Status.CurrentBuildRecipe.Maven:
+		tr.Spec.TaskSpec = &pipelinev1beta1.TaskSpec{
```
Michkov, line 318: Idea for discussion: would not be easier to have it as yaml file which will be [embed](https://pkg.go.dev/embed)?
Pros:
-  Easy to read/edit
- Easy to test standalone

Cons:
- No validation at build time

gabemontero:
Also I need to update the integration test as part of my change here:

```
1.658356633808018e+09	DEBUG	events	Warning	{"object": {"kind":"DependencyBuild","namespace":"default","name":"acmedep1","uid":"848bc9d2-8457-4729-a759-6f470bea9b77","apiVersion":"jvmbuildservice.io/v1alpha1","resourceVersion":"273"}, "reason": "MissingRecipeType", "message": "recipe for DependencyBuild default:acmedep1 neither maven or gradle"}
1.6583566338888743e+09	ERROR	controller.dependencybuild	Reconciler error	{"reconciler group": "jvmbuildservice.io", "reconciler kind": "DependencyBuild", "name": "acmedep1", "namespace": "default", "error": "recipe for DependencyBuild default:acmedep1 neither maven or gradle"}
```

pkg/reconciler/dependencybuild/dependencybuild.go
gabemontero: ok the solution I'm about to push avoids creating yaml files that we do not want included in the infra-deployment etc. processing set, but uses yaml to construct these huge tasks.

Let's see what you all think.

gabemontero:
frustrated by maven and java formatting, I realized / convinced myself updating the java BuildRecipe to add the maven/gradle booleans was not necessary, that only the controller needed that anyway

If someone want to explain what `mvn` command or plugin I need to install where to sort this out in the future, I'm all ears ;-)

stuartwdouglas:
If you just run mvn install it will fix the formatting for you.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -0,0 +1,481 @@
...
+          runAsUser: 0
+        resources:
+          requests:
+            memory: "128Mi"
+            cpu: "10m"
+          limits:
+            memory: "512Mi"
+            cpu: "300m"
+        script: |
+          #!/usr/bin/env bash
```
stuartwdouglas, line 133: I don't really like this approach long term, but I am fine with it for now (we now have bash embedded in yaml embedded in golang, ugh). Once this is in, the gradle work is in, and we have better builder images I will revisit this. 

There is likely to be a lot of common code between Maven/Gradle, and also things that are not necessary any more that can be removed.

gabemontero: Certainly if I'm in the minority wrt not wanting to add/move more yaml files laying around that confuse people, I'll pivot.  But at least for me it is easy to traverse between the consuming code in dependencybuild.go and the variables defined in the new go file in my editor.  And as you noted, as this is suppose to be temporary, this is how I chose to balance the competing concerns, in particular "isolating" future throw away code in golang file we know nobody will mistakingly start referencing in infra-deployments or build-definitions.  I'm also good with pivoting the a golang only approach, though as @Michkov noted that chunk of golang code was getting long and tedious given what is involved with seeding the sidecar.

Certainly better builder images and consolidation of code could help going back to say a golang only approach.

gabemontero: I mean, you said you were fine with it ... I guess I'm just trying to say I get not being super enamored with it :-).  At least for me, some of the other choices bothered me more than this one.

stuartwdouglas: Yea, there does not seem to be any super good choices here, although given that we now have 4 of these I am hoping that we can share enough code/add utility methods so the code based solution does not end up being too bad. 

This is something to deal with later though.

BTW did you consider using https://pkg.go.dev/embed ? As a Java programmer my first impulse would have been to embed this in the file and load it from the class path.

gabemontero:
The dev flow is failing for me

`0s          Warning   TaskRunCreationFailed    dependencybuild/6140a73fc7248bf9bbf9b242ddb6f421                          The DependencyBuild jvm-build-service/6140a73fc7248bf9bbf9b242ddb6f421 failed to create its build pipeline run`

Calling it a night, will debug on Friday.  Hopefully we can get some of these pending e2e PRs merged and I can focus on this.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
gabemontero: I did but this was at the crux in my "I want to reduce the number of yaml files" motivation.

stuartwdouglas:
@gabemontero I pushed a minor fix and now it is working for me locally

stuartwdouglas (approved):

PR #119, opened 2022-07-21
Display the sidecar logs if a task fails

stuartwdouglas:
Tekton removes sidecar logs, which makes debugging errors incredibly difficult. This will output the logs even if the deploy or maven commands fail.

@gabemontero FYI when you are migrating to hard coded tasks.


Michkov (approved):

stuartwdouglas:
Eventually, but I think we should wait to not conflict with David's work.

PR #120, opened 2022-07-21
Minor improvements

stuartwdouglas:


java-components/resource-model/src/main/java/com/redhat/hacbs/resources/model/v1alpha1/BuildRecipe.java
```diff
@@ -10,6 +10,8 @@ public class BuildRecipe {
 
     private String enforceVersion;
 
+    private List<String> ignoredArtifacts;
```
gabemontero, line 13: I should make a note to update the java side of BuildRecipe when I finish https://github.com/redhat-appstudio/jvm-build-service/pull/118

gabemontero:
ok I need to figure out why this repo's prow config is not resulting in PRs merging when all e2e's are green and a PR has the lgtm and approve labels.

PR #128, opened 2022-07-25
Added Container Registry storage

phillip-kruger:
This PR change the Deployer in hacbs-sidecar to load the deployer from a factory and then have the ability to add more deployer.
The current default is still S3 for now, but this PR also adds a Container Registry Storage option.

Signed-off-by: Phillip Kruger <phillip.kruger@gmail.com>

### DIFF
--- a/java-components/sidecar/src/main/java/com/redhat/hacbs/sidecar/resources/deploy/containerregistry/Gav.java
+++ b/java-components/sidecar/src/main/java/com/redhat/hacbs/sidecar/resources/deploy/containerregistry/Gav.java
@@ -2,14 +2,11 @@ package com.redhat.hacbs.sidecar.resources.deploy.containerregistry;
 
 import java.util.Objects;
 
-public class Gav {
-    private String groupId;
-    private String artifactId;
-    private String version;
-    private String tag;
-
-    public Gav() {
-    }
+public final class Gav {
+    private final String groupId;
+    private final String artifactId;
+    private final String version;
+    private final String tag;
 
     public Gav(String groupId, String artifactId, String version, String tag) {
         this.groupId = groupId;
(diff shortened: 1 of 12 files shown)


java-components/sidecar/src/main/java/com/redhat/hacbs/sidecar/resources/deploy/containerregistry/Gav.java
```diff
@@ -0,0 +1,91 @@
+package com.redhat.hacbs.sidecar.resources.deploy.containerregistry;
+
+import java.util.Objects;
+
+public class Gav {
+    private String groupId;
```
stuartwdouglas, line 6: I know it does not really matter here but these should really be immutable. In general mutability + equals/hashCode can result in lots of problems, as the hash code would change if you modify the object.

java-components/sidecar/src/main/java/com/redhat/hacbs/sidecar/resources/deploy/containerregistry/ImageData.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.nio.file.Path;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ImageData {
+
+    private Path artifactsPath;
+    private Set<Gav> gavs;
```
stuartwdouglas, line 11: Same here.

stuartwdouglas (changes requested):
Looks good, just a few minor things. I will hold of on merging it though until we have stability in the e2e tests.

gabemontero:
with https://github.com/redhat-appstudio/jvm-build-service/pull/129 in repo e2e's are green

/retest

gabemontero:
though the compile is failing now at least in https://github.com/redhat-appstudio/jvm-build-service/runs/7493797597?check_suite_focus=true

phillip-kruger:
@stuartwdouglas seems like the error has something to do with testcontainers that needs ryuk to performs fail-safe cleanup of containers. see https://www.testcontainers.org/features/configuration/

How does other builds that needs testcontainers handle this ?

stuartwdouglas:
What makes you say that? It looked like docker.io failed to service an image to me.

stuartwdouglas (approved):

PR #122, opened 2022-07-22
Minor fix

stuartwdouglas:

PR #121, opened 2022-07-21
Handle failure if no build info

stuartwdouglas:

PR #129, opened 2022-07-25
PLNSRVCE-452: switch back to TR pruning, remove TR related tests; test improvements; some code fixes

gabemontero:
pulled in https://github.com/redhat-appstudio/jvm-build-service/pull/121 in case it would get me a fully green e2e (it was not the deciding factor)

pulled in https://github.com/redhat-appstudio/jvm-build-service/pull/122 in case it would get me a fully gree e2e (it was not the deciding factor)

Aside from switching back to TaskRun (instead of Pod) pruning and removing the associated TaskRun verifications in the in repo e2e test that could fail as as result, this PR also augments the tests some, as well as fixes two bugs I hit in local testing
- an empty context provided to the build request process would break it (my workaround is possibly not the ideal sollution, where I listed at least some alternatives in https://coreos.slack.com/archives/C03KN2B47GU/p1658764804972509)
- the side care inlining change had a bug with the cache service URL, which was a stratight forward fix;

To turn off pruning for the in repo e2e job (to allow for debug) but leave it on for eventual e2e's from redhat-appstudio/e2e-tests (to validate that default path) I've cleared out the settings in this repo and have https://github.com/openshift/release/pull/30790 up

As I noted in slack, for the `mvn-goals` and `gradle-goals` failures I am currently seeing, which I think prevent at least some of the artifactbuilds and dependencybuilds to get to complete, I think it might be best to merge this with the failing in repro e2e run, and then let @stuartwdouglas / @dwalluck look at the results, either via local testing or via subsequent PRs they bring up to address.

Ideally, I'd like to not back off from the current stake in the ground of at least having some artifactbuilds and dependencybuilds successfully complete as signs of a validated PR.  But if we collectively agree that is where we are at, so be it.

Lastly, @stuartwdouglas also noted that he was going to come up with a more inexpensive repo to build, given our cluster size constraints in openshift CI.  Curious if the problems I am currently seeing with the 2 goals go away with a "simpler" repo to build.  I'm thinking "no" but did not want to presume either.

@phillip-kruger @brunoapimentel FYI for your respective active PRs.

### DIFF
--- a/pkg/reconciler/dependencybuild/dependencybuild.go
+++ b/pkg/reconciler/dependencybuild/dependencybuild.go
@@ -444,13 +444,9 @@ func (r *ReconcileDependencyBuild) handleTaskRunReceived(ctx context.Context, tr
 			db.Status.State = v1alpha1.DependencyBuildStateSubmitBuild
 		}
 		if os.Getenv(artifactbuild.DeleteTaskRunPodsEnv) == "1" {
-			pod := v1.Pod{}
-			poderr := r.client.Get(ctx, types.NamespacedName{Namespace: tr.Namespace, Name: tr.Status.PodName}, &pod)
-			if poderr == nil {
-				err := r.client.Delete(ctx, &pod)
-				if err != nil {
-					return reconcile.Result{}, err
-				}
+			delerr := r.client.Delete(ctx, tr)
+			if delerr != nil {
+				return reconcile.Result{}, delerr
 			}
 		}
 		return reconcile.Result{}, r.client.Status().Update(ctx, &db)
--- a/pkg/reconciler/dependencybuild/dependencybuild.go
+++ b/pkg/reconciler/dependencybuild/dependencybuild.go
@@ -233,7 +233,7 @@ func (r *ReconcileDependencyBuild) handleStateAnalyzeBuild(ctx context.Context, 
 		}
 	}
 	success := tr.Status.GetCondition(apis.ConditionSucceeded).IsTrue()
-	if !success {
+	if !success || len(buildInfo) == 0 {
 		db.Status.State = v1alpha1.DependencyBuildStateFailed
 		db.Status.Message = message
 	} else {
--- a/pkg/reconciler/dependencybuild/dependencybuild.go
+++ b/pkg/reconciler/dependencybuild/dependencybuild.go
@@ -503,6 +503,11 @@ func (r *ReconcileDependencyBuild) handleStateContaminated(ctx context.Context, 
 func createLookupBuildInfoTask(build *v1alpha1.DependencyBuildSpec) *pipelinev1beta1.TaskSpec {
 	image := os.Getenv("JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE")
 	recipes := os.Getenv("RECIPE_DATABASE")
+	path := build.ScmInfo.Path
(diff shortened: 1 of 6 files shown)


pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -507,6 +503,11 @@ func (r *ReconcileDependencyBuild) handleStateContaminated(ctx context.Context,
 func createLookupBuildInfoTask(build *v1alpha1.DependencyBuildSpec) *pipelinev1beta1.TaskSpec {
 	image := os.Getenv("JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE")
 	recipes := os.Getenv("RECIPE_DATABASE")
+	path := build.ScmInfo.Path
+	//TODO should the buidl request process require context to be set ?
```
gabemontero, line 507: possibly better ways to address an empty context (in the java code or elsewhere) give me pause from merging this even with failing in-repo e2e's in order to facilitate debug /resolution by the rest of the team wrt not getting at least some artifactbuilds or dependencybuilds to complete state.

gabemontero: but if need be let's go with this and the correct it in a follow up PR that ideally addresses the remaining issues preventing at least some of the artifactbuilds and dependencybuilds from reaching complete state

gabemontero:
openshift cluster bring up failure in jvm-build-service-e2e

note/reminder, as of Friday, there are no jvm-build-service tests in jvm-build-service-e2e ... we will add them back in once we have in-repo-e2e working and a smaller repo to build (so we do not get resource starved when the different ginkgo tests run concurrently).

stuartwdouglas:
Oops, looks like I forgot to actually hit send on the simpler repo yesterday: https://github.com/stuartwdouglas/shaded-java-app

stuartwdouglas (approved):

gabemontero:
> Oops, looks like I forgot to actually hit send on the simpler repo yesterday: https://github.com/stuartwdouglas/shaded-java-app

cool I'll try that when I re activate our e2e's in https://github.com/redhat-appstudio/e2e-tests/blob/main/tests/build/jvm-build.go

I'm thinking we keep them same one in the in repo e2e for now as there is no contention with other tests running in parallel

gabemontero:
and we are all green

gabemontero:
> > Oops, looks like I forgot to actually hit send on the simpler repo yesterday: https://github.com/stuartwdouglas/shaded-java-app
> 
> cool I'll try that when I re activate our e2e's in https://github.com/redhat-appstudio/e2e-tests/blob/main/tests/build/jvm-build.go
> 
> I'm thinking we keep them same one in the in repo e2e for now as there is no contention with other tests running in parallel

Hmmm ...https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/103/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1551728649214889984 looks like proof that we can hit constraints / timing issues even with the in-repo e2e's, where we are the only test running, and openshift CI with the quarkus app @stuartwdouglas 

I'll start looking at runs of https://github.com/stuartwdouglas/shaded-java-app on Tuesday and perhaps switch to that

PR #131, opened 2022-07-26
PLNSRVCE-454 move to pipelines

stuartwdouglas:


gabemontero:
You need to rebase your branch @stuartwdouglas  ... I see a reference to an old task run related test that has since been removed because I reactivated pruning:

`    --- FAIL: TestExampleRun/taskruns_related_to_artifactbuilds_and_dependencybuilds_generated (601.71s)`

pkg/reconciler/artifactbuild/artifactbuild.go
```diff
@@ -140,39 +141,39 @@ func (r *ReconcileArtifactBuild) handleTaskRunReceived(ctx context.Context, tr *
...
 			abr.Status.SCMInfo.SCMType = res.Value
-		case TaskResultMessage:
+		case PipelineResultMessage:
 			abr.Status.Message = res.Value
-		case TaskResultContextPath:
+		case PipelineResultContextPath:
 			abr.Status.SCMInfo.Path = res.Value
 		}
 	}
 	if os.Getenv(DeleteTaskRunPodsEnv) == "1" {
```
gabemontero, line 175: yeah our choice of env var name is now unfortunate

I'm assuming we would want to clean that up after the fact wrt this PRs change

Unfortunately it will be a staging exercise:
- add the new PR env in infra deployments after https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build/jvm-build-service/operator-images.yaml#L16-L17 and https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/preview.sh#L100
- update the controller code here to add the new env var, but have the checks look at both env vars (if either is off, don't prune)
- update after https://github.com/openshift/release/blob/master/ci-operator/config/redhat-appstudio/jvm-build-service/redhat-appstudio-jvm-build-service-main.yaml#L117 to add the new env var
- then start deleting refs to the tr env var, starting with infra-deployments, then jvm-build-service, then openshift/release

gabemontero:
> You need to rebase your branch @stuartwdouglas ... I see a reference to an old task run related test that has since been removed because I reactivated pruning:
> 
> ` --- FAIL: TestExampleRun/taskruns_related_to_artifactbuilds_and_dependencybuilds_generated (601.71s)`
> 
> And of course with your move to pipelines there are additional reasons to no longer execute this test

forgot we have a button available to do the rebase  ... just hit it

gabemontero:
OK you may want to force push a rebase from your end @stuartwdouglas (I think I hit the wrong rebase button), but the "in-repo" e2e test failure still seems like there could be a failure reaching complete state (and the old, no longer valid test, is gone).

The pod logs were not super interesting, but the events noted a weird error in various discovery tasks:

`        	* &{0 } (*models.Error) is not supported by the TextConsumer, can be resolved by supporting TextUnmarshaler interface`

stuartwdouglas:
I have rebased, lets see how it goes. Those errors in the test look odd, googling them makes it sound HTTP related.

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -374,6 +375,44 @@ func TestExampleRun(t *testing.T) {
 		}
 	})
 
+	ta.t.Run("taskruns related to artifactbuilds and dependencybuilds generated", func(t *testing.T) {
```
gabemontero, line 378: did you add this on purpose @stuartwdouglas or did this reappear as part of my attempt to provide rebase help?

this test may work in CI because we turn off pruning but can fail for folks testing locally if they do not turn off pruning

I would suggest just keeping this removed, unless you felt strongly about it

stuartwdouglas: I think that is a rebase issue. I will fix.

gabemontero (approved):
let's merge assuming in-repo-e2e passes

PR #135, opened 2022-07-27
revert run.yaml back to 'code-with-quarkus'

gabemontero:
I think this was another unintended after effect of my wrong button'ed attempt to help @stuartwdouglas rebase his https://github.com/redhat-appstudio/jvm-build-service/pull/131

So to run with 'code-with-quarkus' we have
- https://github.com/redhat-appstudio/jvm-build-service/blob/main/hack/examples/run-sample-pipeline.sh
- https://github.com/redhat-appstudio/jvm-build-service/blob/main/hack/examples/run.yaml after this fix

To run 'shaded-java-app' we have
- https://github.com/redhat-appstudio/jvm-build-service/blob/main/hack/examples/run-e2e-shaded-app.sh
- https://github.com/redhat-appstudio/jvm-build-service/blob/main/hack/examples/run-e2e-shaded-app.yaml

PR #136, opened 2022-07-28
YAML cleanup

stuartwdouglas:
- Delete 'staging' and 'production', these were never used and
  non-functional.
- Delete all ClusterTasks and patches targetting them, as they are no
  longer used.


gabemontero:
@stuartwdouglas so I'm seeing the `TestExampleRun/contaminated_build_is_resolved` fail in both this PR and my PR https://github.com/redhat-appstudio/jvm-build-service/pull/135, where at first blush neither of those changes seem like they would affect that aspect of the system.

I see from the test's debug data that netty ends up failed:

`    jvm_build_service_test.go:51: time: 2022-07-28 03:29:50.096689751 +0000 UTC m=+2718.630813518: Netty State: DependencyBuildStateFailed`

where the test is always expecting it to complete:  https://github.com/redhat-appstudio/jvm-build-service/blob/main/openshift-with-appstudio-test/e2e/jvm_build_service_test.go#L482

Does the netty dependency build state have to be completed, or is failed valid?

If failed is not valid, the test is currently dumping logs for 38 pods.  Unfortunately, I had a little trouble figuring how which pod was for the netty dependency build.  Can you help with that?  

I'll make another stab at trying to deciphering that and see what I can find after I'm off all these morning calls and can focus.

Otherwise, I saw some nasty exception traces in a sidecar-proxy pod, but I'm not sure if those are benign or orthogonal.

gabemontero:
so I'm going to add some debug in the test in my PR to make it a little easier to correlate the netty dependency build with the various pods, but my best guess is this `mvn-goals` is at the root of `netty` failing instead of completing:  https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/136/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1552478208459280384/artifacts/jvm-build-service-in-repo-e2e/gather-extra/artifacts/pods/jvm-build-service-test-namespace-brxfx_d1f3a6d34e891ddc731c33f45f8e05de-build-5-task-pod_step-mvn-goals.log

gabemontero:
all green e2e's ... cross ref'ed with https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build/jvm-build-service/kustomization.yaml

merging

PR #140, opened 2022-07-29
Fix operator segfault

stuartwdouglas:


gabemontero:
in repo test passes
kcp think addressed by https://github.com/redhat-appstudio/jvm-build-service/commit/ab1d935d68a0a57674d95b02f9e0b3fae3fbd367 though if kcp has to be address this often I may comment out / remove this github action check until we are closer to being "kcp aware"

There are several pipeline as code issues going on (Pavel and I found one that is still active and Michal mentioned some more stuff today) that I think are affecting our calls here so I am merging even though they have not completed yet

PR #148, opened 2022-07-29
another attempt at checking for whether JVM_BUILD_SERVICE_PR_SHA is set in override tekton script

gabemontero:
chime in with bash expertise @stuartwdouglas or if you think this bash is OK merge this and we can retest https://github.com/openshift/release/pull/30919


stuartwdouglas:
I don't really see how this changes things? They appear to be functionally identical, and testing it out locally I see the exact same results.

stuartwdouglas:
Actually never mind, I see the difference now, this should work

gabemontero:
and it did work ... your openshift/release PR e2e's got past this and are now hitting the bug fixed last Friday ... hopefully with getting the infra deployments bump of our commits merged, that will be able to test with those fixes and be green too

stuartwdouglas:
I think this is a different problem, the builder image we were using for our sample runs is no longer there, I just force merged a fix. 

Basically I was using a Quarkus builder image for my initial experiments and we just never updated it.

PR #149, opened 2022-08-01
Use newer builder image for sample runs

stuartwdouglas:
The image we were using for our sample/CI runs is no longer there. This
gets things working again, but we should change to using a proper
builder image ASAP.


stuartwdouglas:
Merging, as the current situation is completly broken.

PR #150, opened 2022-08-01
PLNSRVCE-533: fix: use hacbs bundle, override sidecar image

psturc:
### What

First part of this ticket https://issues.redhat.com/browse/PLNSRVCE-533:

* update tasks in hacbs bundle instead of default (appstudio) one
* also replace sidecar image reference with the one built by openshift-ci

### How to verify this?

```bash
$ APPSTUDIO_QE_REPO="quay.io/<your-quay-user>/test-images" \
JVM_BUILD_SERVICE_PR_SHA=my-dummy-sha \
JVM_BUILD_SERVICE_ANALYZER_IMAGE=my-analyzer-image \
JVM_BUILD_SERVICE_SIDECAR_IMAGE=my-sidecar-image \
./override-tekton-bundle.sh
# You should see the log message mentioning the hacbs bundle reference:
New build bundle will be created based on quay.io/redhat-appstudio/hacbs-templates-bundle
# then you should see the updated image references in generated task.yaml locally
$ grep "my-" /tmp/bundle-override/task.yaml
      image: my-sidecar-image
    - image: my-analyzer-image
# also note that the tmp/bundle-override/pipelines.yaml now contains hacbs pipeline specific tasks
```

### DIFF
--- a/.ci/override-tekton-bundle.sh
+++ b/.ci/override-tekton-bundle.sh
@@ -19,7 +19,8 @@ function getCurrentBuildBundle() {
     BUILD_BUNDLE=$(yq e ".configMapGenerator[].literals[]" \
     "$TEMP_FOLDER"/build-kustomization.yaml \
     | grep default_build_bundle \
-    | sed "s/default_build_bundle=//")
+    | sed "s/default_build_bundle=//" \
+    | sed "s/build-templates-bundle/hacbs-templates-bundle/")
 
     echo "New build bundle will be created based on $BUILD_BUNDLE"
 }
@@ -51,6 +52,13 @@ function updateAnalyzerImage() {
     |= \"$JVM_BUILD_SERVICE_ANALYZER_IMAGE\"" "$TEMP_FOLDER"/task.yaml
 }
 
+function updateSidecarImage() {
+    echo "jvm-build-service sidecar image set to $JVM_BUILD_SERVICE_SIDECAR_IMAGE"
+    yq e -i "select(.metadata.name == \"s2i-java\") \
+    | (.spec.sidecars[] | select(.image == \"*hacbs-jvm-sidecar*\").image) \
+    |= \"$JVM_BUILD_SERVICE_SIDECAR_IMAGE\"" "$TEMP_FOLDER"/task.yaml
+}
+
 function updatePipelineRef() {
     yq e -i "(.spec.tasks[].taskRef | select (.name == \"s2i-java\").bundle) \
     |= \"$TASK_BUNDLE_IMG\"" "$TEMP_FOLDER"/pipelines.yaml
@@ -77,6 +85,7 @@ getCurrentBuildBundle
 createPipelinesFile
 createTaskFile
 updateAnalyzerImage
+updateSidecarImage
 updatePipelineRef
 createDockerConfig
 createAndPushTaskBundle


gabemontero:
this PR will need https://github.com/redhat-appstudio/jvm-build-service/pull/151

gabemontero:
OK @psturc I see you changes now firing with #151 merged.

However, the e2e failed, and not from an error I've seen before and hence might classify as a flake.  And given the results in the other PRs today being relatively clean, it is possible this is related to your change, though the reason is not obvious to me:

```
     jvm_build_service_test.go:51: time: 2022-08-01 20:44:40.185576124 +0000 UTC m=+602.566064083: non-normal event reason Failed about obj PipelineRun:sample-component-build-cn8z9 message Pipeline jvm-build-service-test-namespace-d77fn/sample-component-build can't be Run; it has an invalid spec: "" type does not match default value's type: "string": finally.params[app-path].default.type, finally.params[app-path].type, finally.params[context].default.type, finally.params[context].type, tasks.params[app-path].default.type, tasks.params[app-path].type, tasks.params[context].default.type, tasks.params[context].type
        invalid value: : finally.params[app-path].type, finally.params[context].type, finally.params[url].type, tasks.params[app-path].type, tasks.params[context].type, tasks.params[url].type
    jvm_build_service_test.go:51: time: 2022-08-01 20:44:40.185586304 +0000 UTC m=+602.566074253: non-normal event reason InternalError about obj PipelineRun:sample-component-build-cn8z9 message 1 error occurred:
        	* "" type does not match default value's type: "string": finally.params[app-path].default.type, finally.params[app-path].type, finally.params[context].default.type, finally.params[context].type, tasks.params[app-path].default.type, tasks.params[app-path].type, tasks.params[context].default.type, tasks.params[context].type
        invalid value: : finally.params[app-path].type, finally.params[context].type, finally.params[url].type, tasks.params[app-path].type, tasks.params[context].type, tasks.params[url].type 
```

Perhaps the way we are building the sidecar in CI is incorrect ?

in CI - https://github.com/redhat-appstudio/jvm-build-service/blob/main/Dockerfile.sidecar-in-ci
how quay.io/redhat-appstudio gets updated:  https://github.com/redhat-appstudio/jvm-build-service/blob/main/.tekton/sidecar-push.yaml#L19-L20

the `mvn` invocations are slightly different, but I didn't *think* enough to break something

WDYT @stuartwdouglas ?

In the interim, 

/retest

gabemontero:
I created https://github.com/redhat-appstudio/jvm-build-service/pull/152 for comparison in case something unexpected has landed and broken us

gabemontero:
ok #152 was clean and so now is the e2e in repo here

gabemontero:
must have been a weird flake earlier

I'm good merging this @stuartwdouglas , or do you recognize what https://github.com/redhat-appstudio/jvm-build-service/pull/150#issuecomment-1201727620 was about and hence does that give you pause

stuartwdouglas:
I think I have seen that before, if a pipeline is created immediately after Tekton is installed you can sometimes get a similar nonsensical error. Locally you can fix it by deleting and re-creating the pipeline.

stuartwdouglas:
Looking at the tekton code it looks like this might be a race that can occur if a taskrun is reconciled before the corresponding task. The task reconcile should set the type of the field based on the default, however if the taskrun reconcile happens first the type has not been written back to the API server and so validation fails.

stuartwdouglas (approved):

gabemontero:
very, very interesting @stuartwdouglas  - and so glad you happen to have seen that before 

in hindsight it makes sense, though that path certainly did not occur to me

my next thought goes to 
- if this occurs with enough frequency
- should we / can we identify this error and delete / recreate as you noted

I'll take a second pass at the failed run during my work day today and see if any answer to ^^ comes from that

gabemontero:
> very, very interesting @stuartwdouglas - and so glad you happen to have seen that before
> 
> in hindsight it makes sense, though that path certainly did not occur to me
> 
> my next thought goes to
> 
>     * if this occurs with enough frequency
> 
>     * should we / can we identify this error and delete / recreate as you noted
> 
> 
> I'll take a second pass at the failed run during my work day today and see if any answer to ^^ comes from that

ok after a second look of a few hours, along with running the tests a few times (not to reproduce, but to see what metadata the tekton controller provides that we can rely on to see that reconciliation is complete, of which I found none), I have landed on 

should NOT do anything for this in the controller (or test for that matter, where the event ^^ stemmed from setting up our test artifacts) at this time

With respect test artifacts, as our e2e's will soon pivot with @psturc 's pending changes to a component based run that leverages the app studio bundles, the precise test creation flow that led to this error goes away (i.e. the test creating the tasks, pipeline, and pipelinerun in quick succession like our dev flow does).  It is entirely possible that even infrequently, this situation is specific to how our developer flow.  

Glass half full - a quasi interesting intellectual exercise.

PR #154, opened 2022-08-02
PLNSRVCE-534: have sidecar image pull policy conform with production grade k8s conventions

gabemontero:
So yeah in production k8s clusters, the general convention is to not use Always for the image pull policy.  That policy best serves inner loop development.

And sure enough, "stress testing", i.e. building service registry, I did instances of 

` Failed to pull image .... pull QPS exceeded`

So a low hanging fruit improvement.

I do have a hack, to go along with our, as I understand it, short term hard coding of task refs in the pipeline runs, to allow for switching from IfNotPresent.  If hard coding becomes permanent, and in conjunction we stabilize our yaml file set and locations, this could make me more amenable to moving off of buildrecipes.go

### DIFF
--- a/pkg/reconciler/dependencybuild/buildrecipeyaml.go
+++ b/pkg/reconciler/dependencybuild/buildrecipeyaml.go
@@ -252,7 +252,7 @@ spec:
       - image: hacbs-jvm-sidecar
         securityContext:
           runAsUser: 0
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
         env:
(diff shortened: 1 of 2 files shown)


gabemontero:
cluster set up fail jvm-build-service-e2e

/test jvm-build-service-e2e

gabemontero:
FYI I'm also running a service registry build with this change ... will give a summary of that as well, comparing with the run that demonstrated the QPS exceeded events noted in this PR's description

gabemontero:
this is also my first service registry run with gradle rebuilds

stuartwdouglas (approved):

PR #155, opened 2022-08-03
Fix sidecar memory issue

stuartwdouglas:
There is a bug in Resteasy Reactive REST client where the response is
being buffered in memory. When running in constrained environments and
downloading a large jar this can result in the sidecar running out of
memory.


sm43:
sorry for spamming, just trying to debug Pipelines As Code failure to report status

/retest jvm-build-cache-on-pull-request

PR #156, opened 2022-08-03
PLNSRVCE-534: depend on IfNotPresent being default (consistent with other steps) and use of specific sha's

gabemontero:
see https://github.com/redhat-appstudio/build-definitions/pull/186#issue-1327280582


gabemontero:
e2e's clean

PR #159, opened 2022-08-03
Add .editorconfig

dwalluck:


stuartwdouglas (approved):

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+end_of_line = lf
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[{*.bash,*.sh}]
+indent_size = 4
+
+[*.go]
+indent_style = tab
```
gabemontero, line 14: so the tab indent style matches what "golang exclusive" editors do, so cool there

when you tried this out @dwalluck doees the `indent_size` default to 4?
If not, I would ask you add `indent_size =4` like you hae for the other file types

thanks

now, the alternative is to just store this in your clone, and we add this to .gitignore
but if you and @stuartwdouglas prefer this approach, I think it will be OK for me.

thanks

dwalluck: It seems that without this file the IDE is not smart enough to know to keep tabs for files like `*.go` and `Makefile`.

gabemontero: OK seems like we'll be OK

I'll merge before EOB, but I'm trying to get https://github.com/redhat-appstudio/infra-deployments/pull/558 and we finally got a clean e2e run, so hopefully I can find an approver soon.

PR #161, opened 2022-08-04
PLNSRVCE-533: test: update in-repo tests, use component pipeline

psturc:
https://issues.redhat.com/browse/PLNSRVCE-533

### List of changes
* added env vars `JVM_BUILD_SERVICE_TEST_REPO_URL` and `JVM_BUILD_SERVICE_TEST_REPO_REVISION` that are propagated to the [jvm-build-service test in e2e-tests repo](https://github.com/redhat-appstudio/e2e-tests/blob/e4506d71d34652258684655ce5303027c272046d/tests/build/jvm-build.go#L31-L32) so that PR created in jvm-build-service repo and any change to the test repo revision is tested with component-type build e2e test (alongside with the "in-repo pipelinerun-based" test located in this repository)
* added a readme file with instructions on how to run in-repo e2e test
* small improvements for `openshift-with-appstudio-test/e2e/jvm_build_service_test.go`


gabemontero:
@psturc - can you put the go mod vendor changes in a one commit, and then the changes to this repo's files in another.

That pattern (which we have to use in openshift extensively) greatly simplifies PR review, as we an just click on the in repo code changes commit and get a smaller file set.

thanks

openshift-with-appstudio-test/e2e/client.go
```diff
@@ -10,6 +10,8 @@ import (
 	projectset "github.com/openshift/client-go/project/clientset/versioned"
 	jvmclientset "github.com/redhat-appstudio/jvm-build-service/pkg/client/clientset/versioned"
 	pipelineclientset "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	"k8s.io/apimachinery/pkg/runtime"
```
gabemontero, line 13: I would prefer this k8s import be coupled with the other k8s imports below (i.e right before the `kubeset` import )

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -6,331 +6,129 @@ package e2e
...
-	"time"
 
-	projectv1 "github.com/openshift/api/project/v1"
 	"github.com/redhat-appstudio/jvm-build-service/pkg/apis/jvmbuildservice/v1alpha1"
-	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 
-	corev1 "k8s.io/api/core/v1"
+	applicationservice "github.com/redhat-appstudio/application-service/api/v1alpha1"
+
+	"k8s.io/apimachinery/pkg/labels"
```
gabemontero, line 18: similar import ordering comment

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -6,331 +6,129 @@ package e2e
...
-func setup(t *testing.T, ta *testArgs) *testArgs {
-	if ta == nil {
-		ta = &testArgs{
-			t:        t,
-			timeout:  time.Minute * 10,
-			interval: time.Second * 15,
+		err = wait.PollImmediate(ta.interval, ta.timeout, func() (done bool, err error) {
+			prs, err := tektonClient.TektonV1beta1().PipelineRuns(ta.ns).List(context.TODO(), listOptions)
+			if err != nil {
+				return false, nil
```
gabemontero, line 53: please log the error before returning with ta.Logf
invariably when we have some weird apiservier issue, that log will help

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -6,331 +6,129 @@ package e2e
...
-	if len(sidecarImage) > 0 {
-		ta.Logf(fmt.Sprintf("PR sidecar image: %s", sidecarImage))
-		for _, sidecar := range ta.maven.Spec.Sidecars {
-			if sidecar.Name != "proxy" {
-				continue
+
+			for _, tr := range pr.Status.TaskRuns {
+				if tr.PipelineTaskName == "build-container" && tr.Status != nil && tr.Status.TaskSpec != nil && tr.Status.TaskSpec.Steps != nil {
+					for _, step := range tr.Status.TaskSpec.Steps {
+						if step.Name == "analyse-dependencies-java-sbom" {
```
gabemontero, line 116: fyi @brunoapimentel here is the other side the equation, the check we talked about, where we verify the bundle override you built is getting used :-)

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -340,26 +138,25 @@ func TestExampleRun(t *testing.T) {
...
 				}
-				ta.Logf(fmt.Sprintf("not yet successful pipeline run: %s", string(prBytes)))
-
+				return false, fmt.Errorf("pipeline run did not succeed: %s", string(prBytes))
 			}
 			return true, nil
 		})
 		if err != nil {
-			debugAndFailTest(ta, "timed out when waiting for the pipeline run to complete")
+			debugAndFailTest(ta, fmt.Sprintf("failure occured when waiting for the pipeline run to complete: %v", err))
```
gabemontero, line 159: the error is going to be a "wait expried / timeout error" anyway

IMO this change does not provide any value

gabemontero (changes requested):
just some minor items @psturc - looks good !!

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -6,331 +6,129 @@ package e2e
...
-		ta.ns = generateName(testNamespace)
-		_, err := projectClient.ProjectV1().ProjectRequests().Create(context.Background(), &projectv1.ProjectRequest{
-			ObjectMeta: metav1.ObjectMeta{Name: ta.ns},
-		}, metav1.CreateOptions{})
+	})
 
-		if err != nil {
-			debugAndFailTest(ta, fmt.Sprintf("%#v", err))
+	ta.t.Run("build-container task from component pipelinerun references a correct sidecar image", func(t *testing.T) {
+		ciSidecarImage := os.Getenv("JVM_BUILD_SERVICE_SIDECAR_IMAGE")
```
gabemontero, line 67: So given the last failed test these were no far off @psturc 
`     utils.go:131: the build-container task from component pipelinerun doesn't contain correct sidecar image. expected: registry.build01.ci.openshift.org/ci-op-6tldjcng/pipeline@sha256:43ee9d375e2ad881c63b85a214c91d5d57b4f6fa1969b5991c602b88c404f5cc, actual: registry.build01.ci.openshift.org/ci-op-6tldjcng/pipeline `

the same because the SHA is stripped from the openshift/release env var at https://github.com/redhat-appstudio/jvm-build-service/blob/85aa504011cc5a7e458f65dc35ca317312a4eb00/.ci/common-ci-settings-and-functions.sh#L47

So you need to use https://github.com/redhat-appstudio/jvm-build-service/blob/85aa504011cc5a7e458f65dc35ca317312a4eb00/.ci/common-ci-settings-and-functions.sh#L48-L49

to build the string to compare with @psturc

gabemontero: not sure why the analyzer check test did not run @psturc but I would imagine it needs the same adjustment

gabemontero: oh ... too much multitasking ... is it the bundle image that is missing the SHA at the end .. if so, then maybe the override is incorrect ?

psturc: @gabemontero so as we discussed I updated the [common-ci....sh](https://github.com/redhat-appstudio/jvm-build-service/pull/161/commits/607e71dddc36a443b05b53e9bc5323bd1dd5a04e) script a bit to avoid stripping the original env vars coming from openshift/release. This should fix the issue. I hope I didn't miss a place where we rely on the former values of those env vars and it doesn't result in some other failure. We'll see

.ci/common-ci-settings-and-functions.sh
```diff
@@ -33,28 +33,38 @@ if [ "$PR_RUN" != "notpr" ]; then
...
+  JVM_BUILD_SERVICE_IMAGE_REPO=${JVM_BUILD_SERVICE_IMAGE_REPO%@*}
   # Tag, i.e. ImageStreamTag, defined at: https://github.com/openshift/release/blob/master/ci-operator/config/redhat-appstudio/jvm-build-service/redhat-appstudio-jvm-build-service-main.yaml
   # as the default always reconciles in CI.  Override this for local testing.
-  export JVM_BUILD_SERVICE_IMAGE_TAG=${JVM_BUILD_SERVICE_IMAGE_TAG:-"redhat-appstudio-jvm-build-service-image"}
+  JVM_BUILD_SERVICE_IMAGE_TAG=${JVM_BUILD_SERVICE_IMAGE_TAG:-"redhat-appstudio-jvm-build-service-image"}
+
+  export JVM_BUILD_SERVICE_CACHE_IMAGE_REPO JVM_BUILD_SERVICE_CACHE_IMAGE_TAG
+
+  JVM_BUILD_SERVICE_CACHE_IMAGE_REPO=${JVM_BUILD_SERVICE_CACHE_IMAGE:-"quay.io/redhat-appstudio/hacbs-jvm-cache"}
+  JVM_BUILD_SERVICE_CACHE_IMAGE_REPO=${JVM_BUILD_SERVICE_CACHE_IMAGE_REPO%@*}
```
psturc, line 48: I originally intended to do this change on a single line, but the shellcheck was screaming at me that I should not nest param expansions (although it worked perfectly fine in my terminal 😜  )

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
gabemontero: those env vars are referenced in the controller golang code @psturc 
```
./pkg/reconciler/dependencybuild/dependencybuild.go:	image := os.Getenv("JVM_BUILD_SERVICE_SIDECAR_IMAGE")
./pkg/reconciler/dependencybuild/dependencybuild.go:	image := os.Getenv("JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE")
./pkg/reconciler/artifactbuild/artifactbuild.go:	image := os.Getenv("JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE")
```

but of course they are not affected by the machinations in https://github.com/redhat-appstudio/jvm-build-service/pull/161/commits/607e71dddc36a443b05b53e9bc5323bd1dd5a04e

similar, some of the overlay / `make dev` yaml files reference them, but again, the should be unaffected by https://github.com/redhat-appstudio/jvm-build-service/pull/161/commits/607e71dddc36a443b05b53e9bc5323bd1dd5a04e

.ci/common-ci-settings-and-functions.sh
gabemontero: shellcheck in particular / pedantic ;-)

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
gabemontero: otherwise the proof should be in these tests here ... the image is right in the sidecar / analyzer or it is not right :-)

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
psturc: Hmm I thought pollImmediate works in a way that it returns an error in cases
* of a timeout
* conditionFunc returns an error

so I was assuming the timeout is not always a cause of a failure and was thinking that in case the pipelinerun fails (not because of a timeout), it prints the error [from here](https://github.com/redhat-appstudio/jvm-build-service/blob/607e71dddc36a443b05b53e9bc5323bd1dd5a04e/openshift-with-appstudio-test/e2e/jvm_build_service_test.go#L154)

But I don't really mind to switch it back (especially if it doesn't work like that :)

openshift-with-appstudio-test/e2e/const.go
```diff
@@ -0,0 +1,12 @@
...
+
+const (
+	testNamespace          = "jvm-build-service-test-namespace-"
+	maxNameLength          = 63
+	randomLength           = 5
+	maxGeneratedNameLength = maxNameLength - randomLength
+
+	testProjectGitUrl = "https://github.com/stuartwdouglas/hacbs-test-project"
+	// TODO update this
+	testProjectDevfileUrl = "https://raw.githubusercontent.com/psturc/shaded-java-app/main/devfile.yaml"
```
psturc, line 11: @gabemontero @stuartwdouglas we need to point an AppStudio Component to a devfile (that is then parsed by build-service which then decides which pipeline is going to be used for building the component). I used a devfile from my other repo, which I guess is fine now, but I was wondering if we could make it a part of that `hacbs-test-project` repo, so it lives together with the source code - wdyt?

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
gabemontero: ah ... I missed that you are returning an error if the pipelinerun failed and is done

so nevermind leave it as is  @psturc

openshift-with-appstudio-test/e2e/const.go
gabemontero: @stuartwdouglas owns the test repos, so I'll defer to him (though I am betting it is OK to add a devfile)

.ci/common-ci-settings-and-functions.sh
gabemontero: though I like using it when I'm not sure if I am getting too crazy with bash

psturc:
@gabemontero good news is that all tests updated in this PR have passed. But this `some_artfactbuilds_and_dependencybuilds_complete` test has failed and the output from the test is [really long](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/161/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1555263524807118848/artifacts/jvm-build-service-in-repo-e2e/jvm-build-service-e2e-in-repo-e2e/build-log.txt). Any idea what went wrong there? It seems that none of the buildArtifacts completed.

gabemontero:
> @gabemontero good news is that all tests updated in this PR have passed. But this `some_artfactbuilds_and_dependencybuilds_complete` test has failed and the output from the test is [really long](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/161/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1555263524807118848/artifacts/jvm-build-service-in-repo-e2e/jvm-build-service-e2e-in-repo-e2e/build-log.txt). Any idea what went wrong there? It seems that none of the buildArtifacts completed.

hey @psturc - first, yes, there are a lot of pod logs to consider when diagnosing a failure, so our output is log, and some of the in flight logging has been increased recently to help with debug, so that increased it further.  And it gets trickly because errors are expected in many of the pod logs as different JVM versions, maven versions, gradle versions, are tried.

you'll have the opportunity the limit the logging as you see fit when we get this test passing and you mimic this test in the e2e-test repo :-) 

Now - why it is failing?  Well, one thing I see that I at least typically do not see is that several of the `deploy-and-check-for-contaminates` containers are getting a bunch of nasty looking aws SDK exception and stack traces I normally do not see.  That step failing most likely breaks the expected chain of events.

Why those exceptions are happening?  Well, the underlying change with your test pivot is using the bundle's analyzer and sidecar setting vs. the dev ones the prior form of the test used, and it is affecting communication with the localstack pod.  Maybe the exception is totally orthogonal.  But I also see a nasty looking python stack track in the localstack pod.  So maybe it got resource starved, died, and that is why those contaiminate checks got those AWS SDK exceptions.

So we'll 

to see if it fails consistently.  If it does not fail again, then it was an AWS flake.  But if it does, and I am correct that is root cause, we'll need @stuartwdouglas to take a look.  I have no experience with that AWS S3 local stack stuff currently used.

Quick FYI - I am out Friday.  My guess is @stuartwdouglas will be able to look at this while past business hours.

psturc:
ok, thank you for explaining it in great detail, looks like you were right @gabemontero, it was an AWS flake, because the in-repo-test has passed this time.
the other e2e-tests have failed on cluster connection error
```
Unable to connect to the server: dial tcp 52.52.253.187:6443: i/o timeout
```

There are some conflicts in the test file I need to resolve, but otherwise I think I need to just babysit this PR and it should be good to go if you're ok with changes proposed here @gabemontero @stuartwdouglas

gabemontero:
/hold

see https://coreos.slack.com/archives/C03KN2B47GU/p1659958940328259 for the tl;dr

and to confirm our understanding, let's 

again now that #105 has merged; I would expect that until https://github.com/redhat-appstudio/build-definitions/pull/179 has merged that cache URL is incorrect, and assuming that nasty amazon stack trace we are seeing in the cache instance in the `jvm-build-service` namespace if fact means it is broke, this will fail

gabemontero:
well the in repo e2e passed .... is it because the the cache service still exists in the jvm-build-service namespace (and it is functionaly) ?!?!?

I think I'm going to have to check this PR out and confirm one way or the other.  Also, the artifact-cache pods do not have that nasty looking amazon exception.

gabemontero:
ok ... this passed without the build-definitions PR because #105 had this short term work around - https://github.com/redhat-appstudio/jvm-build-service/blob/650846d2011d745b82c72cd59cf1e2398daeddfa/pkg/reconciler/dependencybuild/dependencybuild.go#L405-L408

with that, at some point, we'll need to test the build-defintions change merged, with that code ^^ commented out, and with this change to prove it works.

We'll just need to decide @psturc @stuartwdouglas the staging for all that

openshift-with-appstudio-test/e2e/const.go
psturc: created a PR for that: https://github.com/stuartwdouglas/hacbs-test-project/pull/1

psturc:
Appstudio bootstrap failed (cluster API connection issue)

psturc:
@gabemontero all CI checks are green, I think this PR is ready for re-review

psturc:
Oops, I forgot to update the devfile URL - I did that in the last commit. The content of a devfile basically the same, so it shouldn't affect the test functionality.

gabemontero:
Next, from the in repo e2e, confirmation that the images built from CI are being used, greps from the pod yaml collected from must gather:

```
                        "image": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline:redhat-appstudio-jvm-build-service-jdk8-builder-image",
                        "image": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline:redhat-appstudio-jvm-build-service-jdk17-builder-image",
                        "image": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline:redhat-appstudio-jvm-build-service-jdk11-builder-image",
                        "image": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline@sha256:e9298e05f542a38a341779c3c97c618d9b706b8bb1e4214889ea4af209b61339",
                        "imagePullPolicy": "IfNotPresent",
                        "name": "step-process-build-requests",
                        "image": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline:redhat-appstudio-jvm-build-service-cache-image",
                        "image": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline@sha256:e9298e05f542a38a341779c3c97c618d9b706b8bb1e4214889ea4af209b61339",
                        "imagePullPolicy": "IfNotPresent",
                        "name": "step-lookup-artifact-location",
                        "env": [
                            {
                                "name": "JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE",
                                "value": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline:redhat-appstudio-jvm-build-service-reqprocessor-image"
                            },
                            {
                                "name": "JVM_BUILD_SERVICE_SIDECAR_IMAGE",
                                "value": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline:redhat-appstudio-jvm-build-service-sidecar-image"
                            },
                        "image": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline:redhat-appstudio-jvm-build-service-image",
                        "image": "registry.build01.ci.openshift.org/ci-op-prbrigm0/pipeline:redhat-appstudio-jvm-build-service-cache-image",
```

That ^^ includes the bundle updates as well.

Also, there were no hits on `quay.io/redhat-appstudio/hacbs`, so good news there, as  PR from this repo should be overriding all the default jvm-build-service images.

openshift-with-appstudio-test/e2e/util.go
```diff
@@ -276,3 +278,58 @@ func streamFileYamlToTektonObj(path string, obj runtime.Object, ta *testArgs) ru
...
+	}
+
+	ta.appstudioApp = app
+
+}
+
+func createAppstudioComponent(ta *testArgs) {
+	name := generateName("test-component-")
+
+	quayOrg := os.Getenv("QUAY_E2E_ORGANIZATION")
```
gabemontero, line 305: should we update the README at https://github.com/redhat-appstudio/jvm-build-service/blob/main/.ci/README.md to describe this env var ?

gabemontero (commented):
just a minor item on if we should update the test README @psturc 

if you agree, let's get that updated and then we can merge

or if there is a good reason not to update the README (or if my scan of it missed that it was already there), we'll merge this as is then

thanks

openshift-with-appstudio-test/e2e/util.go
psturc: Good point.

This env var is required if your quay.io account doesn't have permissions to push to `quay.io/redhat-appstudio/test-images` repo - in that case the test would be failing on the `build-container` taskrun if you'd try to run the test locally.

I think it's not needed to document it in the .ci folder (since CI will always work with `redhat-appstudio-qe` quay and github orgs by default), but I think it'd make sense to add a note about it to the `openshift-with-appstudio-test/e2e/`. Or perhaps in the root readme file? wdyt?

gabemontero: yeah sure create a README in https://github.com/redhat-appstudio/jvm-build-service/tree/main/openshift-with-appstudio-test/e2e and add this bit there, along with the basic need an openshift cluster with appstudio bootstrapped, kubeconfig set or cluster login

thanks

psturc:
aws s3 `NoSuchBucketException`:
```
[com.red.hac.art.ser.LocalCache] (executor-thread-9) Failed to download commons-cli/commons-cli/1.0/commons-cli-1.0.jar from s3://Optional[artifact-store][default]: java.lang.RuntimeException: software.amazon.awssdk.services.s3.model.NoSuchBucketException:
The specified bucket does not exist (Service: S3, Status Code: 404, Request ID: 7a62c49f-347e-4fc4-9331-6e8eEXAMPLE, Extended Request ID: 
```

openshift-with-appstudio-test/e2e/util.go
psturc: created a readme file: https://github.com/redhat-appstudio/jvm-build-service/pull/161/commits/724fc86081912ebcacedeea7697eab8f81f13e03

psturc:
appstudio bootstrap failure

psturc:
same issue as in previous run (aws s3 NoSuchBucketException)

@gabemontero do you know if that could be a flake or something we should look into?

gabemontero:
looks like @stuartwdouglas has a change up with https://github.com/redhat-appstudio/jvm-build-service/pull/194 @psturc to hopefully help with this

gabemontero:
ok @psturc per the discussion thread where I just posted the comment https://github.com/redhat-appstudio/e2e-tests/pull/176#discussion_r949185910 I think I am changing my opinion today from stand up about having both a component based test as well as the existing pipelinerun based test.

I think we need both now, but it is not because of an "ease of use" thing.  We need both to properly cover changes to the test repo *BEFORE* we merge those, so that we have done sufficient coverage and don't break the e2e-tests with unvetted changes in the test repro.

Of course ping me in slack if my explanation here is not clear, and we can hope on a video conf and hopefully clear things up.

gabemontero:
> ok @psturc per the discussion thread where I just posted the comment [redhat-appstudio/e2e-tests#176 (comment)](https://github.com/redhat-appstudio/e2e-tests/pull/176#discussion_r949185910) I think I am changing my opinion today from stand up about having both a component based test as well as the existing pipelinerun based test.
> 
> I think we need both now, but it is not because of an "ease of use" thing. We need both to properly cover changes to the test repo _BEFORE_ we merge those, so that we have done sufficient coverage and don't break the e2e-tests with unvetted changes in the test repro.
> 
> Of course ping me in slack if my explanation here is not clear, and we can hope on a video conf and hopefully clear things up.

OK @psturc and I had a videoconf; he understands the need here (a jvm build service PR for testing a specific revision of the test repo) and will look at first augmenting the component based approach here and in https://github.com/redhat-appstudio/e2e-tests/pull/176 to facilitate that before bringing the dev flow / pipelinerun induced flow back

psturc:
passing the test repo-related env vars to e2e works as expected - ran this command against the CI cluster used for the `jvm-build-service-e2e` CI job:
```bash
➜  /tmp oc get components --all-namespaces -o json | jq -r '.items[].spec.source.git'
{
  "revision": "233724555ee2aff17a97b0ddf2f143308209c6a3",
  "url": "https://github.com/stuartwdouglas/hacbs-test-project"
}
```

psturc:
All e2e checks are green. Now I will rebase and push a small change (adding new lines between different import sources), which will trigger another round of e2e tests - hopefully successful ones, so we can finally merge this one if you're happy with the changes @gabemontero @stuartwdouglas

gabemontero:
so I think this is good to go

PR #165, opened 2022-08-08
Add service account for local testing

stuartwdouglas:


gabemontero (approved):
yeah these files are not accessed by https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build/jvm-build-service/kustomization.yaml#L3-L5

PR #166, opened 2022-08-08
Use a specific revision for the tests

stuartwdouglas:
This will prevent our tests being broken by changes to the test
repository (e.g. adding a gradle test).


gabemontero:
the jvm build service e2e failed with our pruning timing issue
that is getting tracked in @stuartwdouglas #171 with testing soon to come with https://github.com/openshift/release/pull/31184

merging

PR #167, opened 2022-08-08
Use K8n ConfigMap as ConfigSource

phillip-kruger:
Signed-off-by: Phillip Kruger <phillip.kruger@gmail.com>

### DIFF
--- a/java-components/sidecar/pom.xml
+++ b/java-components/sidecar/pom.xml
@@ -59,6 +59,10 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-kubernetes-config</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
             <artifactId>quarkus-junit5</artifactId>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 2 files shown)


stuartwdouglas:
@gabemontero @Michkov this approach allows us to configure the sidecar directly from the config map, the`quarkus-kubernetes-config` extension directly access the config from the kube API. 

The other way to do it would be to mount the config in the sidecar but this approach seems a lot simpler as it does not require any YAML changes. Do either of you have any concerns with this approach?

gabemontero:
> @gabemontero @Michkov this approach allows us to configure the sidecar directly from the config map, the`quarkus-kubernetes-config` extension directly access the config from the kube API.
> 
> The other way to do it would be to mount the config in the sidecar but this approach seems a lot simpler as it does not require any YAML changes. Do either of you have any concerns with this approach?

I like the simplicity here vs. mounting the configmap.  Neat quarkus extension.  

Questions (trying to perform due diligence PR review) to make sure we are covered:
- is there retry semantics / config with this extension to deal with weird / infrequent things like momentary API server throttling?
- if something weird happens like the config map is deleted right after the sidecar is launched, and before the quarkus extention can fetch it, do we think it will fail in a way that a customer could decipher?

Both of those items are covered by core k8s if we went down the mount the config map approach.

And just out of curiosity since this is a draft PR

phillip-kruger:
Closing here, will include this in the fallback PR

PR #175, opened 2022-08-09
correct periodic db count check

gabemontero:


gabemontero:
s3 hiccup with e2e - unrelated

PR #179, opened 2022-08-11
fix: set recreate strategy for cache deployment

Michkov:
Cache is using Read-write-once PVC which does not work with default
Rolloing update strategy, changing to Recreate.


gabemontero:
unhealthy openshift api server with pruning run

an ok *this* was the root cause wrt what I reported in slack a couple of days ago in the jvm-build-service namespace, right?

good find @Michkov 

/approve
/lgtm

PR #187, opened 2022-08-16
Use latest version of tests

stuartwdouglas:


gabemontero:
@psturc FYI this is related to the devfile PR in the test repo that @stuartwdouglas just merged

gabemontero:
apparent tool install flake still with the pruning e2e ... will 

/retest

one more time and if it still occurs, must be some diff in the different jobs where something broke?

PR #188, opened 2022-08-16
Initial work on local container based dev

stuartwdouglas:
This will now enumerate the available artifacts on the server, which makes for much much faster builds (way faster than localstack).

At present this just enables local development flow, next step is to integrate into e2e and staging.

### DIFF
--- a/Dockerfile.cache-in-ci
+++ b/Dockerfile.cache-in-ci
@@ -2,7 +2,7 @@ FROM registry.access.redhat.com/ubi8/openjdk-17:1.13 AS builder
 WORKDIR /work
 COPY . .
 USER 0
-RUN cd java-components/cache && mvn clean package -DskipTests
+RUN cd java-components && mvn clean package -DskipTests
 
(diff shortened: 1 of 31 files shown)


stuartwdouglas:
The lack of sidecar logs is a real problem here, and it looks like it is crashing before logging is established.

stuartwdouglas:
In repo tests are now passing

stuartwdouglas:
@gabemontero this is green now, it would be great to get this in so we can start looking at using this for our e2e tests.

gabemontero:
got a merge conflict now @stuartwdouglas

gabemontero (approved):

gabemontero:
feel free to merge after your rebase and the tests getting green again ... just github approved

it is a US holiday on Monday so don't want for me to hit the merge button once ^^ occurs

stuartwdouglas:
The JDK 17 build does appear to have actually completed (i.e. image was pushed), and the build controller only failed on 'show summary'.

I think this is as close to a green build we are going to get at the moment.

PR #191, opened 2022-08-17
Added a way to prepend the SHA tags for container registry

phillip-kruger:
So we can run multiple tests


stuartwdouglas (approved):

gabemontero:
S3 / AWS SDK / localstack pain during the in repo e2e; presumably with the pruning job passing this is just an env / constrain flake, but let's see if we can get clean

gabemontero:
all green ... merging

PR #192, opened 2022-08-18
DO NOT MERGE

stuartwdouglas:


openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -4,265 +4,17 @@
...
-	if !ok {
-		debugAndFailTest(ta, fmt.Sprintf("file %s did not produce a pipelinerun: %#v", runYamlPath, obj))
-	}
-	ta.run, err = tektonClient.TektonV1beta1().PipelineRuns(ta.ns).Create(context.TODO(), ta.run, metav1.CreateOptions{})
-	if err != nil {
-		debugAndFailTest(ta, err.Error())
-	}
+	//test to make sure prow artifacts work the way I think the docs say they do
+	d1 := []byte("hello\ntest  artifacts\n")
+	_ = os.WriteFile(os.Getenv("ARTIFACTS")+"/TEST_ARTIFACT.txt", d1, 0644)
```
stuartwdouglas, line 14: @gabemontero based on the prow job authors guide (https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md) I was expecting this to result in a 'TEST_ARTIFACT.txt' file I could download at the end. Any idea if this is possible?

I want to create a structured zip file of the results (build logs etc) with a HTML report that makes it easy to examine the current status.

gabemontero: sorry @stuartwdouglas I have no experience with that specifically; though I *suspect* openshift CI is already overriding that to create the artifacts link you after you hit one of the Details links in the e2e status.

their docs on how they construct all that is at https://docs.ci.openshift.org/docs/how-tos/artifacts/

otherwise, their slack channel is #forum-testplatform on coreos slack

A couple of thoughts:
- there are ways to have specific tests results to show up in the junit section (I have seen it with the ginkgo based tests in openshift/origin), but I'm not sure how that gets wired up, or if it is even possible with our "in repo" form of testing.  I don't see me diving into trying to sort that out any time soon, but of course, if you are motivated, go for it.

- you can download an archive of all the gathered data today; for example, if you go to https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/192/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-with-pruning-e2e/1560093510080663552/ and click on the download widget, you are provided with a gsutil invocation; I have been able to use that recently; check out https://cloud.google.com/storage/docs/gsutil_install.  You can also grab sub sections if you do down the dir structures

stuartwdouglas: I wanted to do something slightly different to both of those, basically take all the related failure information and group it in the same directory of a zip file, and then allow you to download that so you can easily peruse the failures.

I might have to take a different approach, and maybe look at making the way we dump logs attempt to group everything related.

stuartwdouglas: Ah, I see you have already done that at: https://github.com/redhat-appstudio/jvm-build-service/pull/196

gabemontero:
@stuartwdouglas - looks like @psturc knew how to do this artifact gathering you were trying to do here.

See https://github.com/redhat-appstudio/e2e-tests/pull/181/files

In particular, at least in that repo, `ARTIFACT_DIR` was the env var name.  And I see that env var defined / described in the openshift CI team's docs (they manage the openshift prow instance and have written the system on top of prow that we engage with via https://github.com/openshift/release ) over in  https://docs.ci.openshift.org/docs/how-tos/artifacts/#ci-operator-artifacts

PR #193, opened 2022-08-18
Allow sidecar to read configmap

stuartwdouglas:
This affects local dev

### DIFF
--- a/deploy/base/rbac.yaml
+++ b/deploy/base/rbac.yaml
@@ -11,6 +11,16 @@ rules:
       - artifactbuilds
     verbs:
       - create
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+      - secrets
+    resourceNames:
+      - jvm-build-config
+      - jvm-build-secrets
+    verbs:
+      - get
 ---
 apiVersion: v1
 kind: ServiceAccount
--- a/pkg/reconciler/dependencybuild/dependencybuild.go
+++ b/pkg/reconciler/dependencybuild/dependencybuild.go
@@ -437,6 +437,7 @@ func (r *ReconcileDependencyBuild) handleStateBuilding(ctx context.Context, db *
 		// work around for developer mode while we are hard coding the task spec in the controller
 		pr.Spec.PipelineSpec.Tasks[0].TaskSpec.Sidecars[0].ImagePullPolicy = v1.PullAlways
 	}
+	pr.Spec.ServiceAccountName = "pipeline"
 	//TODO: this is all going away, but for now we have lost the ability to confiugure this via YAML
 	//It's not worth adding a heap of env var overrides for something that will likely be gone next week
 	//the actual solution will involve loading deployment config from a ConfigMap


deploy/base/rbac.yaml
```diff
@@ -11,6 +11,16 @@ rules:
       - artifactbuilds
     verbs:
       - create
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
```
gabemontero, line 17: I guess technically this allows the get on a configmap named jvm-build-secrets and a secret named jbm-build-config as well (not sure if that was intended)

No big deal ... this is still sufficiently narrow in scope

gabemontero (commented):
/lgtm

deploy/base/rbac.yaml
stuartwdouglas: I thought about this overnight and I think that this is actually wrong. Because of the way permissions work in Tekton AFAIK this would also allow user code to read the secret, and compromise the quay credentials.

We need to change this to actually mount the config into the sidecar I think, which is a problem for the secret as it may not exist in all cases. We probably need to have the operator create an empty secret when provisioning the workspace.

stuartwdouglas: @phillip-kruger I don't think we can use the quarkus-kubernetes-config extension, as it requires permissions that would then be propagated to the user code in the pipeline run.

stuartwdouglas: @gabemontero I am not actually sure how the original PR that introduced the need for this passed, I would have expected the sidecar to crash being unable to read the config. Is there a difference in how the user is setup on e2e vs in the staging cluster itself?

gabemontero: well there are layers of indirection here so the answer to your question @stuartwdouglas , so my answer is layered / a little tl;dr :-) :
1) for creating any pod in k8s, the permissions of the requesting user and the service account are unioned, and if either of those have permission, the pod is allowed
2) with tekton, the tekton controller's SA is always the "requesting user" for the pod, and in our case, our `pipeline` SA is the SA always used 

those 2 ^^ do not vary whether it is an e2e cluster or the appstudio staging cluster

Now, are there differences in the permissions applied either the Tekton controller SA set up by openshift pipeliens or each namespace's pipeline SA in the staging cluster vs. our e2e's?  That should not be the case AFAIK.

Now there is an `oc` command that helps you resolve these questions. Unfortunately, I do not have permission to run this on the staging cluster, so I cannot answer that side of it.

But if you run

`oc adm policy who-can get configmap jvm-build-config -n <test namespace>`

You'll find out who can do that.  So today, for my testing, I have an openshift cluster set up for our "dev" flow, where I
- installed tekton via the operatorhub at the openshift console
- ran `make dev`

And sure enough that `oc` command listed:
```
system:serviceaccount:openshift-pipelines:tekton-pipelines-controller
```

so *THAT* ^^ is why it woks on my test cluster.  By the way, I get the same result with the `jvm-build-secrets` Secret.

If you want to find out how it looks on the staging cluster or our openshift CI, where appstudio sets things up, launch a openshift cluster via clusterbot, do a preview boostrap from your local clone of infra-deployments, create a test namespace, and run that `oc adm policy who-can get` command for each item.  If I can remember, I'll try it next time a create a cluster that way.  My guess though is that the tekton controller SA has permissions.

Now, is it a good thing that the tekton controller can do that ?!?!?!?  

A question for the folks in #forum-tektoncd-pipelines, `make dev` is doing that.  Let's checkpoint next week, and then see about raising a big storm with the tekton guys if need be.

Finally, back to this PR's description, and the fact that this PR was for our developer / local dev flow.  Now, I can absolutely believe that the `pipeline` SA created by OpenShift PIpelines / RH downstream tekton (where that SA is what is used in both staging cluster and e2e) are different than the permissions for the `pipeline` SA manually set up as part of our dev flow.

And hence why you needed this change for that.

gabemontero: OK it is an upstream tekton rbac def that allows the tekton controller to read any secret:  https://github.com/tektoncd/pipeline/blob/main/config/200-clusterrole.yaml#L58-L61

I suppose it has been vetted then.

PR #196, opened 2022-08-18
PLNSRVCE-585: add focused debug for service registry periodic to accelerate failures to target

gabemontero:
@dwalluck @stuartwdouglas here is the dump data for failed db's and ab's regardless change to the service registry test I started talking about last week

got to admin, when I was running this locally, it was not obvious why some of the db's / ab's were marked as failed based on the pod logs's for their pipelineruns

presumably it will be obvious to you guys when you start seeing results

I may merge this sooner rather than later in order to test with https://github.com/openshift/release/pull/31424; then, once we have the optional PR level test job which anyone on the team can choose to run from a PR, if you all see ways to improve / tweak what this is doing, we can adjust

once obvious thing - there are functions here which should go into utils.go, but I'm deferring on that to no further disrupt @psturc 's https://github.com/redhat-appstudio/jvm-build-service/pull/161

### DIFF
--- a/openshift-with-appstudio-test/e2e/periodic_test.go
+++ b/openshift-with-appstudio-test/e2e/periodic_test.go
@@ -7,14 +7,18 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


gabemontero:
all green

PR #200, opened 2022-08-26
feat: added codecov, golangci-lint, security scan

psturc:
https://issues.redhat.com/browse/HACBS-1055


.github/workflows/codecov-main.yaml
```diff
@@ -0,0 +1,19 @@
+name: Codecov report on push to the "main" branch
```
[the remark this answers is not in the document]
psturc: codecov is used in many [openshift org's repositories](https://app.codecov.io/gh/openshift) and also in [redhat-appstudio's](https://app.codecov.io/gh/redhat-appstudio).

I do not know whether there's a strict requirement to use only this tool.
However the codecov suggestions/warning do not force developers to address all of them, it's up to an individual person to decide whether it makes sense to take some actions on them.

The main purpose of using this tool is to provide an information about a test coverage and suggest how it can be improved. Also there are ways to configure this tool, i.e. possibly we might be able to remove the codecov code suggestions in case they prove to be too much annoying :)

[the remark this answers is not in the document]
psturc: golanci-lint - I will try to address it in this PR

gabemontero:
the gosec complaints are "won't fix"as I see it, as what @stuartwdouglas is doing is not encrypting data, but rather just encoding 

the test result page seemed to provide a way of indicating that, though I'm curious of that selection persists across PRs @psturc

psturc:
> the gosec complaints are "won't fix"as I see it, as what @stuartwdouglas is doing is not encrypting data, but rather just encoding
> 
> the test result page seemed to provide a way of indicating that, though I'm curious of that selection persists across PRs @psturc

I think we can remove it by [annotating those pieces of code](https://github.com/securego/gosec#annotating-code) that we know that aren't a security issue

gabemontero:
> > the gosec complaints are "won't fix"as I see it, as what @stuartwdouglas is doing is not encrypting data, but rather just encoding
> > the test result page seemed to provide a way of indicating that, though I'm curious of that selection persists across PRs @psturc
> 
> I think we can remove it by [annotating those pieces of code](https://github.com/securego/gosec#annotating-code) that we know that aren't a security issue

cool ... so as I read https://github.com/redhat-appstudio/jvm-build-service/pull/200/checks?check_run_id=8093189957:
- for dependencybuild.go, that is currently line 5, 155, 593
- for artifacdtbuild.go, that is currently line 5, 6, 376, 499
can you handle those as well with this PR @psturc ?  or would you rather defer to @stuartwdouglas or I for that ?

psturc:
> > > the gosec complaints are "won't fix"as I see it, as what @stuartwdouglas is doing is not encrypting data, but rather just encoding
> > > the test result page seemed to provide a way of indicating that, though I'm curious of that selection persists across PRs @psturc
> > 
> > 
> > I think we can remove it by [annotating those pieces of code](https://github.com/securego/gosec#annotating-code) that we know that aren't a security issue
> 
> cool ... so as I read https://github.com/redhat-appstudio/jvm-build-service/pull/200/checks?check_run_id=8093189957:
> 
> * for dependencybuild.go, that is currently line 5, 155, 593
> * for artifacdtbuild.go, that is currently line 5, 6, 376, 499
>   can you handle those as well with this PR @psturc ?  or would you rather defer to @stuartwdouglas or I for that ?

I'm on it - will add a comment here once it's ready for a review

psturc:
issues with CI
```
Connecting to github.com (github.com)|140.82.112.3|:443... connected.
HTTP request sent, awaiting response... 500 Internal Server Error
2022-08-30 15:18:16 ERROR 500: Internal Server Error.
```

/test jvm-build-service-e2e

psturc:
all tests passed 😮 

@stuartwdouglas @gabemontero  ready for another round of reviews

openshift-with-appstudio-test/e2e/types.go
```diff
@@ -1,3 +1,6 @@
+//go:build normal && periodic
```
gabemontero: interesting; I didn't need this to build

any idea why you needed it @psturc ?

openshift-with-appstudio-test/e2e/util.go
```diff
@@ -49,7 +52,7 @@ func dumpPods(ta *testArgs, namespace string) {
 				ta.Logf(fmt.Sprintf("error getting pod logs for container %s: %s", container.Name, err.Error()))
 				continue
 			}
-			b, err := ioutil.ReadAll(readCloser)
+			b, err := io.ReadAll(readCloser)
```
gabemontero, line 55: ioutil was flagged for some reason then?

pkg/reconciler/artifactbuild/artifactbuild.go
```diff
@@ -496,15 +496,15 @@ func (r *ReconcileArtifactBuild) handleRebuild(ctx context.Context, abr *v1alpha
...
+	hashedBytes := sha1.Sum([]byte(gav)) //#nosec
 	hash := hex.EncodeToString(hashedBytes[:])[0:8]
 	namePart := gav[strings.Index(gav, ":")+1:]
 
 	//generate names based on the artifact name + version, and part of a hash
 	//we only use the first 8 characters from the hash to make the name small
 	var newName = strings.Builder{}
 	lastDot := false
-	for _, i := range []rune(namePart) {
+	for _, i := range namePart {
```
gabemontero, line 507: interesting ... this give you any pause @stuartwdouglas ?  I think it is OK but this name generation is yours

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -588,9 +588,10 @@ func (r *ReconcileDependencyBuild) handlePipelineRunReceived(ctx context.Context
 		// even though we filter out artifactbuild pipelineruns, let's check the kind and make sure
 		// we use a dependencybuild ownerref
 		var ownerRef *v12.OwnerReference
-		for _, or := range ownerRefs {
+		for i, or := range ownerRefs {
 			if strings.EqualFold(or.Kind, "dependencybuild") || strings.EqualFold(or.Kind, "dependencybuilds") {
-				ownerRef = &or
+				ownerRef = &ownerRefs[i]
+				break
```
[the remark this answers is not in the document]
gabemontero: yeah that is "slice 101" ... good catch on gosec's part

gabemontero (commented):
I couple of minor questions @psturc but seems good to me

Aside from answering those, let's get a yay/nay from @stuartwdouglas before merging

thanks!!

openshift-with-appstudio-test/e2e/util.go
psturc: yes, [deprecated from v1.16](https://pkg.go.dev/io/ioutil)

stuartwdouglas (approved):

openshift-with-appstudio-test/e2e/types.go
psturc: apologies, I should have mentioned that golangci-lint wasn't originally including [jvm_build_service_test.go](https://github.com/redhat-appstudio/jvm-build-service/blob/main/openshift-with-appstudio-test/e2e/jvm_build_service_test.go) and [periodic_test.go](https://github.com/redhat-appstudio/jvm-build-service/blob/main/openshift-with-appstudio-test/e2e/periodic_test.go) (because of the build tags), hence it evaluated types and functions in go files in this folder as unused. So I updated these files with build tags and they mean that they're going to be included into go build/run only if we specify `normal` or `periodic` tag (which is already set in [Makefile](https://github.com/redhat-appstudio/jvm-build-service/blob/c5469a6c6037922b9a2dfc28b69b856ebd0c46f6/Makefile#L33-L37))

psturc: in fact the way I configured golangci-lint and gosec jobs is wrong. I created a follow-up PR with fixes and small improvements: https://github.com/redhat-appstudio/jvm-build-service/pull/207

PR #203, opened 2022-08-30
PLNSRVCE-654: Allow checking out commits in addition to branches

dwalluck:


gabemontero:
@dwalluck see my comment in our slack thread today about how to launch the service registry build.  I'm going to let you do it as an educational exercise.

The default e2e's are with simpler repos that do not produced failures.

stuartwdouglas (approved):

PR #211, opened 2022-08-31
Change some sidecar memory settings

stuartwdouglas:


gabemontero:
in particular those maven localhost connection refused errors

stuartwdouglas:
This will also need an infra deployments update, otherwise the sidecar there might have memory issues so 

/hold

gabemontero:
> This will also need an infra deployments update, otherwise the sidecar there might have memory issues so
> 
> /hold

I will see if I can find cycles for ^^ today

stuartwdouglas:
I would not update infra deployments until we know for sure what is going on. Something very weird is happening, and most of the actual exceptions that are displayed are after the build has finished (when the sidecar is shutting down).

stuartwdouglas:
Let's just close this, and work on removing the sidecar altogether.

PR #212, opened 2022-08-31
Use official builder images for gradle

stuartwdouglas:
This changes the way that builder images are produced, and how gradle builds are performed. In particular:

- Builder images now have multiple gradle versions. The versions that are including are driven by the config map information so the config and builders will stay aligned.
- The build discovery tool now has no knowledge of available versions, it just performs the discovery and reports what it has found
- dependencybuilds.go is now responsible for mapping the discovered information into builder image invocations.

PR #215, opened 2022-09-01
shift dev flow toward CI flow (possible prep for kcp disruption)

gabemontero:
so this PR shifts our developer flow toward the openshift CI flow in few ways
- it maintains the `QUAY_USERNAME` "entry point", but converts that into the env vars openshift/release sets for all of our generated images
- it allows for replacement of the entire image vs. the repository portion of a quay.io image spec; this allows for use of images built in CI which are not stored in quay.io, but openshift's internal image registry
- it installs tekton via the openshift pipelines operator, but does it via cli/yaml vs. the web console path
- it does not require the building of the builder images everytime you run `make dev` but it does require you to do it at least once; this will mean your local developer flow testing will be a bit closer to what openshift CI gives you for the builder images, vs. the current fallback of using the builder images @stuartwdouglas maintains in his quay.io repository.

What motivated me to do this:
- the aforementioned "official builder image" element to local testing
- it makes my "developer flow" testing on openshift easier, as I no longer have to log onto the openshift console and install openshift pipelines to get tekton
- *most impacting* ... with the impending shift in infra-deployments to kcp, and with that, how appstudio will bootstraps will be in flux, and it is very possible that all the e2e's we have added here to test our changes can be rendored useless.  In stand up today, Michal said he had not heard officially if e2e-tests would immediately get ported to the pre-kcp branch of infra-deployments.  *IF THERE IS A NOTICEABLE DELAY*, I should be able to pivot with relative ease from our e2e's  using the e2e-tests infrastructure for bootstrapping appstudio using our `make dev` flow to install tekton and our controller.
- Longer term, when the KCP transition is "far enough along", including our controller, which I've only started working on this week, presumably a new test "framework" will exist that we need to transition to.  Again, details are fuzzy still on when that will be, what that will look like, how hard is it to consume, etc.

One caveat: it occurred to me that our dev flow overlay system config map is missing the newer tags that @stuartwdouglas is adding with https://github.com/redhat-appstudio/infra-deployments/pull/681.  But since both @dwalluck and myself asked questions / requested changes in that PR, I did not try to replicate those changes in https://github.com/redhat-appstudio/jvm-build-service/compare/main...gabemontero:openshift-install-openshift-pipelines-dev-flow?expand=1#diff-d8945ba487c78b7bfeac39f2220ab59c50015ccb794351cbd9096778ea6032dd ..... presumably once everything is addressed in that PR, we would want to keep these in sync, either in this PR, or a follow up one.

With these changes, I was able to run `make dev` and then `make appstudio-installed-on-openshift-e2e` against my personal openshift cluster for today.

@phillip-kruger FYI


stuartwdouglas (approved):

stuartwdouglas:
With regard to the new tags those infra-deployments changes are temporary, https://github.com/redhat-appstudio/jvm-build-service/pull/212/files#diff-8944c42fd2615afea6412ddd9d9d14643e29ee78ae8d5019364e396dcd70871d will add them into our repo, and then infra deployments can be changed to reference this patch to keep things in sync.

gabemontero:
> With regard to the new tags those infra-deployments changes are temporary, https://github.com/redhat-appstudio/jvm-build-service/pull/212/files#diff-8944c42fd2615afea6412ddd9d9d14643e29ee78ae8d5019364e396dcd70871d will add them into our repo, and then infra deployments can be changed to reference this patch to keep things in sync.

awesome / perfect

tests all green / merging

PR #216, opened 2022-09-02
Disable color in Gradle Manipulator console output

dwalluck:


pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -584,9 +584,9 @@ spec:
           echo ADDITIONAL_ARGS="${ADDITIONAL_ARGS}"
 
           if [ -n "$(params.ENFORCE_VERSION)" ]; then
-              gradle-manipulator --info --stacktrace -l "${GRADLE_HOME}" $(params.GRADLE_MANIPULATOR_ARGS) -DversionOverride=$(params.ENFORCE_VERSION) "${ADDITIONAL_ARGS}" generateAlignmentMetadata || exit 1
+              gradle-manipulator --no-colour' --info --stacktrace -l "${GRADLE_HOME}" $(params.GRADLE_MANIPULATOR_ARGS) -DversionOverride=$(params.ENFORCE_VERSION) "${ADDITIONAL_ARGS}" generateAlignmentMetadata || exit 1
```
stuartwdouglas, line 587: looks like there is an extra comma here that is breaking the gradle builds

dwalluck: Fixed.

PR #220, opened 2022-09-02
PLNSRVCE-606 Multiple maven installations

stuartwdouglas:


stuartwdouglas:
@dwalluck @gabemontero so it seems like the 'standard' java images might not really be what we want, as the runtime image does not provide tools we need like 'javadoc', and the non-runtime one that we are using at the moment packages extra stuff (i.e. maven) that we don't really need (which is causing the linked issue about two different maven's being installed). 

I wonder if we should just base these images on UBI and install Java ourselves? It seems like it would give us a lot more flexibility.

dwalluck:
> I wonder if we should just base these images on UBI and install Java ourselves? It seems like it would give us a lot more flexibility.

I think that's a good idea. We can just install java-*-openjdk-devel manually on top. Note that if using ubi-minimal, builds also seemed to require glibc-langpack-en.

dwalluck:
> does not provide tools we need like 'javadoc', and the non-runtime one that we are using at the moment packages extra stuff (i.e. maven) that we don't really need (which is causing the linked issue about two different maven's being installed).

You should be able to do `dnf -y remove maven maven-lib`, and it's cleaner than just doing `rm`, before installing the new maven because the files are in /usr/share/maven/lib.

stuartwdouglas:
closing for now, as this is moving to a new repo anyway

PR #223, opened 2022-09-02
General Cleanup

stuartwdouglas:
- Remove Embedded YAML
- Remove unused defaults
- Remove references to personal repos


gabemontero (approved):
For some reason I remember trying tasks instead of taskruns immediately and running into some sort of problem, but I don't remember the specifics, and if I did run into that issue, I don't see where I documented it in the original PR (though I documented other issues I ran into).

At this point, if  tests are passing, cool :-) 

Not merging yet cause I don't want to disrupt the infra dep bump and its current e2e run.

stuartwdouglas:
@gabemontero I have changed this significantly to remove all the embedded YAML, and to have the corresponding shell scripts as actual .sh files.

I think this increases the maintainability significantly, as now the tasks are unified and can share constant values with the rest of the code.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -1,541 +1,192 @@
...
+	var build string
+	if maven {
+		settings = mavenSettings
+		build = mavenBuild
+	} else {
+		settings = gradleSettings
+		build = gradleBuild
+	}
+	zero := int64(0)
+	buildSetup := pipelinev1beta1.TaskSpec{
```
gabemontero, line 46: I originally tried to initialize via golang struct formatting but kept running into presumably syntax errors that lead to runtime errors (vs. compile errors) instantiating objects, and then fell back to yaml based creation, and worrying about the yaml file sprawl.

I'm happy to see you were able to sort this out.

gabemontero: this removes the needs around "verification" that I mentioned in our slack conversation, but said I would clarify within the PR.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -1,541 +1,192 @@
...
+			{Name: PipelineIgnoredArtifacts, Type: pipelinev1beta1.ParamTypeString},
+		},
+		Sidecars: []pipelinev1beta1.Sidecar{createSidecar(sidecarImage, namespace)},
+		Steps: []pipelinev1beta1.Step{
+			{
+				Container: v1.Container{
+					Name:  "git-clone",
+					Image: "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.21.0", //TODO: should not be hard coded
+					Resources: v1.ResourceRequirements{
+						Requests: v1.ResourceList{"memory": resource.MustParse("128Mi"), "cpu": resource.MustParse("10m")},
```
gabemontero, line 67: minimally add TODO comments around all these resource mem/cpu settings, unless you want to take it on in this PR

long term like the image ref todo's you have, these need to be configurable

pkg/reconciler/dependencybuild/dependencybuild_test.go
```diff
@@ -249,7 +251,7 @@ func TestStateBuilding(t *testing.T) {
 		db.Namespace = metav1.NamespaceDefault
 		db.Name = "test"
 		db.Status.State = v1alpha1.DependencyBuildStateBuilding
-		db.Status.CurrentBuildRecipe = &v1alpha1.BuildRecipe{Image: "quay.io/sdouglas/hacbs-jdk11-builder:latest"}
+		db.Status.CurrentBuildRecipe = &v1alpha1.BuildRecipe{Image: "quay.io/redhat-appstudio/hacbs-jdk11-builder:latest"}
```
gabemontero, line 254: should we not be using the `JVM_BUILD_SERVICE_JDK11_BUILDER_IMAGE` env var here, with perhaps a sha specific version of `quay.io/redhat-appstudio/hacbs-jdk11-builder` as a default?

gabemontero (changes requested):
some minor requests/questions but an improvement for sure @stuartwdouglas as you mentioned in slack

pkg/reconciler/dependencybuild/dependencybuild_test.go
stuartwdouglas: This is a unit test, it does not actually use the image but just verifies that it uses the expected one. This could be any string and it would be ok.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
stuartwdouglas: I have started this for the cache at https://github.com/redhat-appstudio/jvm-build-service/pull/233, am planning on going through and making all the limits configurable at some point.

Configuring the image is a bit trickier, it probably should be using the task from the build definitions bundle.

stuartwdouglas:
I have added the todos, lets see if CI happens to be working at the moment.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
gabemontero: sounds good

now, in #211 you bumped this from 128Mi to 256Mi; I know #211 was a bit of a guesstimate, but I just wanted to confirm what you think the default should be here

I'm also thinking we should move whatever changes are in #211 that we want to keep to this PR, and then close #211, right?

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -1,541 +1,200 @@
...
+				PeriodSeconds:       3,
+			},
+			ReadinessProbe: &v1.Probe{
+				ProbeHandler:        v1.ProbeHandler{HTTPGet: &v1.HTTPGetAction{Path: "/q/health/ready", Port: intstr.IntOrString{IntVal: 2000}}},
+				InitialDelaySeconds: 1,
+				PeriodSeconds:       3,
+			},
+			Resources: v1.ResourceRequirements{
+				//TODO: make configurable
+				Requests: map[v1.ResourceName]resource.Quantity{"memory": resource.MustParse("256Mi"), "cpu": resource.MustParse("10m")},
```
gabemontero, line 190: ok this lines up with the change you have in #211

gabemontero (commented):
did some cross referencing with #211 below

I do realize that the moving of function from sidecar to cache can affect the default values here, so I suppose we have some staging decisions to make based on when you get time for that.

There are also the dockerfile changes in #211, but I'll comment there about those.

gabemontero:
with throttling merged, going to update this branch, try the jvm-build-service-e2e test again

gabemontero:
ok with this last run, the original component build failed per https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/223/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-e2e/1567540681721778176/artifacts/jvm-build-service-e2e/jvm-build-service-e2e/artifacts/jvm-build-service-test/jvm-build-htyb-pr-jvm-build-suite-component-nhwe-r822z.log and is sidecar/cache constraint ... from the super long exception stack trace:

```
Caused by: org.apache.http.conn.HttpHostConnectException: Connect to localhost:2000 [localhost/127.0.0.1] failed: Connection refused
```

let's try again and see if we at least get to ab's and db's getting generated

/test jvm-build-service-e2e

pkg/reconciler/dependencybuild/dependencybuild_test.go
gabemontero: duh right

gabemontero:
OK @stuartwdouglas we have pipelinerun logs to look at now that the less stringent form of pruning is in play with the throttling PR merged.

The collection of stuff is at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/223/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-e2e/1567577332888113152/artifacts/jvm-build-service-e2e/jvm-build-service-e2e/artifacts/jvm-build-service-test/

But at least for me, and to be clear, I don't think this PR introduced them, but the errors at 
- https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/223/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-e2e/1567577332888113152/artifacts/jvm-build-service-e2e/jvm-build-service-e2e/artifacts/jvm-build-service-test/jvm-build-jaby-pr-55df1004d27dabdc12e4db533fd222ed-build-0.log
- https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/223/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-e2e/1567577332888113152/artifacts/jvm-build-service-e2e/jvm-build-service-e2e/artifacts/jvm-build-service-test/jvm-build-jaby-pr-8b31ad91766ce8c95fda3b3fb485203c-build-0.log
- https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/223/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-e2e/1567577332888113152/artifacts/jvm-build-service-e2e/jvm-build-service-e2e/artifacts/jvm-build-service-test/jvm-build-jaby-pr-ea268de30de6bab363b6ccea90e4628c-build-1.log


are interesting.  They are one of the types I noticed while you were on PTO that I tried to describe last week.  I think it speaks to a disconnect between what is in build-definitions/infra-deployment vs. jvm-build-service only (i.e. our dev flow) of some sort.  The unknown host exception makes me think the localstack was not launched.  I also don't see any localstack pods in the generic must gather directory.

stuartwdouglas:
It looks like 'enable-rebuilds' has not been set to true, which results in localstack not being created.

Unfortunately though it seems like this still results in the creation of ABR's, which should be disabled for a non-rebuild namespace, I have filed https://issues.redhat.com/browse/PLNSRVCE-695 to track this.

That said, it begs the question as to how the hell this ever worked? My guess is that it didn't, and the -e flag I added to the scripts has shown this up.

stuartwdouglas:
So this PR is needed for this to pass: https://github.com/redhat-appstudio/e2e-tests/pull/193

For now I think I will remove the '-e' and add it in a seperate PR. This needs to go in ASAP as it has a high chance of causing conflicts.

stuartwdouglas:
I have addressed the requested changes and we have a green run, so I am merging.

PR #224, opened 2022-09-02
Search Gradle build files using regex

dwalluck:
This is more flexible than a plain string.

Also, since Gradle does not seem to have an API to build a model from a build file, it may just be easier to look for various constructs by regex right now.

### DIFF
--- a/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/gradle/GradleUtils.java
+++ b/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/gradle/GradleUtils.java
@@ -25,7 +25,7 @@ public final class GradleUtils {
     /**
      * Identifier for the plugin {@code com.github.sherter.google-java-format}.
      */
-    public static final String GOOGLE_JAVA_FORMAT_PLUGIN = "com.github.sherter.google-java-format";
+    public static final String GOOGLE_JAVA_FORMAT_PLUGIN = ".*" + Pattern.quote("com.github.sherter.google-java-format") + ".*";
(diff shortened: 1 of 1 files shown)


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/gradle/GradleUtils.java
```diff
@@ -200,7 +200,7 @@ public static boolean isInBuildGradle(Path path, CharSequence csq) {
                         } catch (IOException e) {
                             return Stream.empty();
                         }
-                    }).anyMatch(s1 -> s1.contains(csq));
+                    }).anyMatch(s1 -> s1.matches(regex));
```
stuartwdouglas, line 203: Shouldn't this be using `Matcher.find()` instead of matches, to remove the need for a `.*` at the start and end? Also I would suggest using `java.util.regex.Pattern#MULTILINE` rather than iterating over the lines, so you can match multi line constructs.

stuartwdouglas (changes requested):

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/gradle/GradleUtils.java
dwalluck: How's 3f7f8b8?

stuartwdouglas (approved):

dwalluck:
> New changes are detected. LGTM label has been removed.

I just updated the javadoc, but all tests are passing now.

PR #226, opened 2022-09-02
Add gradle plugins repo to test

stuartwdouglas:


stuartwdouglas:
This needs the redirect config from https://github.com/redhat-appstudio/jvm-build-service/pull/212, I might just merge it into the same PR

PR #227, opened 2022-09-05
Include support for minikube based dev

stuartwdouglas:
This splits the dev mode between openshift and minikube. You can now
run make dev-openshift or make dev-minikube depending on your target
environment.


gabemontero:
the openshift version of this worked for me after checking out this PR locally

merging

PR #235, opened 2022-09-06
PLNSRVCE-688: Add jsweet maven repository for org.jsweet.ext:typescri…

dwalluck:
…pt.java-ts.core:jar:1.4.0.1

Required by apicurio-data-models-1.1.26.

### DIFF
--- a/deploy/base/config.yaml
+++ b/deploy/base/config.yaml
@@ -12,6 +12,7 @@ data:
   maven-repository-304-eclipselink: "https://download.eclipse.org/rt/eclipselink/maven.repo"
   maven-repository-305-redhat: "https://maven.repository.redhat.com/ga"
   maven-repository-306-gradleplugins: "https://plugins.gradle.org/m2"
+  maven-repository-307-jsweet: "https://repository.jsweet.org/artifactory/libs-release-local"
 ---
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


dwalluck:
I never updated the `BUILD_POLICY_DEFAULT_STORE_LIST` value for an repositories I added. I can do that in a separate PR.

deploy/cache/base/deployment.yaml
```diff
@@ -51,6 +51,10 @@ spec:
               value: https://maven.repository.redhat.com/ga
             - name: STORE_REDHAT_TYPE
               value: maven2
+            - name: STORE_JSWEET_URL
```
stuartwdouglas, line 54: This file has no effect any more, it was still around for the transtion  but has just been deleted.

stuartwdouglas (approved):
LGTM  once the conflict is fixed.

deploy/cache/base/deployment.yaml
dwalluck: Removed from this PR.

stuartwdouglas:
Note that BUILD_POLICY_DEFAULT_STORE_LIST is not really needed any more, this is automatically populated by the operator when it sets up the cache.

PR #236, opened 2022-09-06
PLNSRVCE-658: kcp rbac, api export permission, api binding, logicalclusterv2, some kcp startup based testing

gabemontero:
so this is a placeholder jvm-build-service draft PR that around 

`Watch a virtual workspace and define permission claims in its APIExport - [ASC-164](https://issues.redhat.com/browse/ASC-164)`

that is partially based on some active PR work I am seeing in the application-service (though I am not pulling everything yet I'm seeing there), as well as some initial experimentation with being registering my openshift cluster for the day running this version of the jvm-build-service while logged into CPS per https://docs.google.com/document/d/1ygHQOBDMMnnsEDQoXSLSxaLUrplhOC42R3_kHCr25TI/edit#heading=h.dwe8luhxmneh and creating a workspace and syncher with my openshift cluster, as well as create the kcp api* items and see them recognized both via the CLI as well as the syncer deployment running in my backend cluster.

I'm putting this draft PR up @mmorhun @Michkov @stuartwdouglas for awareness and a possible sanity check, if what I have here jumps out at you enough to warrant comment.

Otherwise, my next steps / hopes are to next register a cluster that has
- https://github.com/redhat-appstudio/infra-deployments/pull/709 bootstrap / preview changes from @Michkov after that has merged, plus additions to add jvm-build-service back to https://github.com/redhat-appstudio/infra-deployments/tree/main/components  in an infra-deployements test branch, pointing to his branch
- hopefully have enough created / running to instantiate our pipeline/piplinerun based dev flow and see how far along in the sync'ing process we get before things break down


deploy/operator/base/rbac.yaml
```diff
@@ -1,130 +1,6 @@
-apiVersion: rbac.authorization.k8s.io/v1
```
stuartwdouglas, line 1: Just something to be aware of is that this change will mean we need to also do a corresponding update in `pre-kcp` infra-deployments to make sure the operator still has permission.

Depending on when this goes in this may or may not be an issue.

deploy/kcp-manager/kustomization.yaml
```diff
@@ -0,0 +1,39 @@
...
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+#namePrefix: application-service-
+
+# Labels to add to all resources and selectors.
+#commonLabels:
+#  someName: someValue
+
+bases:
+  - ../operator/base
```
gabemontero, line 16: after #258 merges, I'll add `../operator/config` after `../operator/base`

deploy/operator/base/rbac.yaml
gabemontero: yeah generally speaking (and by that I mean the best practice we adopted in openshift) it is better to have more files with single instances of api obj instances vs. one of these yaml files with a bunch of objects, and so I was trying to handle that here; however, also at the time (about 2 weeks ago now) I first did this break up, I think I was getting weird errors trying to execute the new kcp kustomize paths when access this global file (I don't have those errors now however 2 weeks later) - complaints about creating a file twice or something, don't remember for sure now.

however, the amount of rbac changes appears active (I see you have more changes in your sidecar / cache refactor).  It may make sense to break up the rbac.yaml file in a separate PR when things have settled down (assuming I do not hit that same error I saw 2 weeks ago with the kcp kustomzie apply).

I hope to circle back to this PR today, and we'll try to go back to just an rbac.yaml file, and see what issues I do or do not hit.  If I don't hit any issues, then I'll stay with rbac.yaml and we can consider breaking it up to "adhere to best practices" in a separate PR.

gabemontero: ok I was able to sort this out / get `rbac.yaml` working

will be pushing shortly

gabemontero:
ok rebases were going well with regression testing passing until I rebased on #260 .... since that rebase the baseline e2e is failling for me, and I'm not sure exactly why @stuartwdouglas 

I may push the changes up here in a moment anyway, where either 
a) the rebase mistake is obvious from the code diffs, if not to me, to you @stuartwdouglas :-) 
b) the off chance my local env is messed up with all the recycles, perhaps a clean run in the PR looks OK

gabemontero:
> ok rebases were going well with regression testing passing until I rebased on #260 .... since that rebase the baseline e2e is failling for me, and I'm not sure exactly why @stuartwdouglas
> 
> I may push the changes up here in a moment anyway, where either a) the rebase mistake is obvious from the code diffs, if not to me, to you @stuartwdouglas :-) b) the off chance my local env is messed up with all the recycles, perhaps a clean run in the PR looks OK

OK was lucky and it was option b) ... where things are passing in CI ... whew :-)

And actually am managing to do some kcp related tesing using https://github.com/openshift-pipelines/pipeline-service/blob/main/ckcp/README.md ... will have some updates from that hopefully before EOB today.

gabemontero:
so @stuartwdouglas @Michkov - my initial testing on kcp today has uncovered our first blocker / decision point

Specifically, the processing of various config setting in our "system level" configmap which we traditionally define in the `jvm-build-service` namespace.

So currently, after creating our deployment under the kcp kubeconfig, the resulting pod in the backend cluster (which is not in the `jvm-build-service` namespace) fails finding the system config.

Now, I went as far as making a local change in the operator to determine its on namespace by accessing the file system:

```
	// because of KCP, we cannot assume we are running in configmap.SystemConfigMapNamespace so let's get the namespace from the pod
	// via /run/secrets/kubernetes.io/serviceaccount/namespace
	namespace := configmap.SystemConfigMapNamespace
	if kcp {
		fb, ferr := os.ReadFile("/run/secrets/kubernetes.io/serviceaccount/namespace")
		if ferr != nil {
			controllerLog.Info(fmt.Sprintf("error reading pod namespace file: %s", ferr.Error()))
			return nil, ferr
		}
		namespace = string(fb)
		controllerLog.Info(fmt.Sprintf("controller namespace is %s", namespace))
	}
```

However, that value still thinks it is in `jvm-build-servivce` !!!  kcp is doing some *tricky* stuff here :-) 

```
gmontero ~/go/src/github.com/redhat-appstudio/jvm-build-service  (kcp-rbac)$ oc logs hacbs-jvm-operator-9f666d64b-vnkwn  -n kcp-7eiabsy3j22v
1.6632741365745795e+09	INFO	main	Looking up virtual workspace URL
1.6632741365839887e+09	INFO	main	Using virtual workspace URL	{"url": "https://ckcp-ckcp.apps.gmontero410.devcluster.openshift.com:443/services/apiexport/root:default:pipeline-service-compute/jvmbuildservice"}
1.6632741365924003e+09	INFO	controller-runtime.metrics	Metrics server is starting to listen	{"addr": ":8080"}
1.6632741365946953e+09	INFO	controller	controller namespace is jvm-build-service
1.6632741365966756e+09	ERROR	controller	timed out waiting for system ConfigMap jvm-build-service/jvm-build-system-config to be created	{"error": "no matches for kind \"ConfigMap\" in version \"v1\""}
github.com/redhat-appstudio/jvm-build-service/pkg/controller.NewManager
	/opt/app-root/src/pkg/controller/controller.go:120
main.main
	/opt/app-root/src/cmd/controller/main.go:81
runtime.main
	/usr/lib/golang/src/runtime/proc.go:255
1.6632741365967298e+09	ERROR	main	unable to start cluster aware manager	{"error": "no matches for kind \"ConfigMap\" in version \"v1\""}
main.main
	/opt/app-root/src/cmd/controller/main.go:83
runtime.main
	/usr/lib/golang/src/runtime/proc.go:255
gmontero ~/go/src/github.com/redhat-appstudio/jvm-build-service  (kcp-rbac)$ oc get cm jvm-build-system-config -n kcp-7eiabsy3j22v
NAME                      DATA   AGE
jvm-build-system-config   7      3m57s
```

Note, the config map I created while under the kcp kubeconfig is in fact present ^^.

I *have* to think there is some kcp related wiring we can do that reconciles this (a controller accessing a configmap in its namespace is a very common pattern), but I don't know what that solution is yet.

I'll start pinging folks in slack my Friday AM / Europe's PM, unless one of you two already happen to know the answer.

If by some chance there is not a work around for this, we are going to have to pivot and define these items as annotations or env vars on the controller's deployment.

gabemontero:
Or we remove that check in the beginning, but none of the reconcilers process anything / requeue events until the configmap reconciler receives the event for our "system config" config map.

A more complicated change, but perhaps/probably the "philosophically correct" one @stuartwdouglas @Michkov

deploy/kcp/apiexport_jvmbuildservice.yaml
```diff
@@ -0,0 +1,27 @@
...
+  name: jvmbuildservice
+spec:
+  permissionClaims:
+  - group: ""
+    resource: "secrets"
+  - group: ""
+    resource: "configmaps"
+  - group: "jvmbuildservice.io"
+    identityHash: 2aa7813102d123ead4faa4e5cdf4b9d87b21eb9d9e0424475c9cb3cfd1963e50
+    resource: "artifacdtbuilds"
```
Michkov, line 16: typo, and I think that you do not need to have permissionClaim on resources which are in latestResourceSchemas.

gabemontero: thanks, both for the typo, and then, yeah, I was uncertain about that particular .... now that I'm actually testing, I'll be sure and try a permutation with these removed and confirm one way or the other.

gabemontero:
actually, after testing today, I've convinced myself we have to move off configmaps and create 2 new CRs, one each for
- system config
- per namespace "user" config

Building on the aforementioned namespace conundrum in kcp, since we cannot assume which namespace it comes from, there is no viable way now of preventing as user from creating their own "system" config map in their namespace and come in conflict with ours.  

A new CR, with appropirate RBAC, would negate this.

Then, with the user config map, a for this, I should have known better, as I had to deal with this with shared resources, especially on openshift, there is a **TON** of confgmap update activity across all the namespaces in the cluster (including openshift component namespaces).  We are getting too much event activity.  Our controller will quickly get earmarked as "hot" by SREs monitoring "real" clusters.

Certainly I intend to talk to @stuartwdouglas about his "voice to voice" next week, but in the interim, I'm going to start looking at this conversion as part of my kcp story here, given the system config implication.

gabemontero:
OK I've pushed an update that switches the system config from a config map to a crd and is working for me locally

For kcp at least, I think this is the meets in, though for performance reasons, I do still want to convert the user config map.  But we could do that separately.  This also minimizes the surface area of this change that @stuartwdouglas and I still need to discuss in our team sync up next week.

gabemontero:
I also pushed a new fix for this perm issue I am now seeing in the artifact cache that I believe is unrelated to the other changes here:
```
        2022-09-16 17:32:46,017 ERROR [org.jbo.res.rea.com.cor.AbstractResteasyReactiveContext] (executor-thread-8) Request failed: com.google.cloud.tools.jib.api.CacheDirectoryCreationException: Could not create cache directory
        	at com.google.cloud.tools.jib.cache.Cache.withDirectory(Cache.java:58)
        	at com.google.cloud.tools.jib.configuration.BuildContext$Builder.build(BuildContext.java:329)
        	at com.google.cloud.tools.jib.api.JibContainerBuilder.toBuildContext(JibContainerBuilder.java:726)
        	at com.google.cloud.tools.jib.api.JibContainerBuilder.containerize(JibContainerBuilder.java:591)
        	at com.redhat.hacbs.artifactcache.deploy.containerregistry.ContainerRegistryDeployer.createImages(ContainerRegistryDeployer.java:170)
        	at com.redhat.hacbs.artifactcache.deploy.containerregistry.ContainerRegistryDeployer.deployArchive(ContainerRegistryDeployer.java:130)
        	at com.redhat.hacbs.artifactcache.deploy.containerregistry.ContainerRegistryDeployer_ClientProxy.deployArchive(Unknown Source)
        	at com.redhat.hacbs.artifactcache.resources.DeployResource.deployArchive(DeployResource.java:223)
        	at com.redhat.hacbs.artifactcache.resources.DeployResource$quarkusrestinvoker$deployArchive_91c6a91f4462db62dda378bdb05b70742226a722.invoke(Unknown Source)
        	at org.jboss.resteasy.reactive.server.handlers.InvocationHandler.handle(InvocationHandler.java:29)
        	at io.quarkus.resteasy.reactive.server.runtime.QuarkusResteasyReactiveRequestContext.invokeHandler(QuarkusResteasyReactiveRequestContext.java:108)
        	at org.jboss.resteasy.reactive.common.core.AbstractResteasyReactiveContext.run(AbstractResteasyReactiveContext.java:140)
        	at io.quarkus.vertx.core.runtime.VertxCoreRecorder$14.runWith(VertxCoreRecorder.java:557)
        	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2449)
        	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1478)
        	at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:29)
        	at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:29)
        	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        	at java.base/java.lang.Thread.run(Thread.java:833)
        Caused by: java.nio.file.AccessDeniedException: /home/jboss/.cache
        	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:90)
        	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
        	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        	at java.base/sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:397)
        	at java.base/java.nio.file.Files.createDirectory(Files.java:700)
        	at java.base/java.nio.file.Files.createAndCheckIsDirectory(Files.java:807)
        	at java.base/java.nio.file.Files.createDirectories(Files.java:793)
        	at com.google.cloud.tools.jib.cache.Cache.withDirectory(Cache.java:56)
        	... 18 more
```

gabemontero:
Lastly, I was having various issues in my ckcp testing setting up watchers for various CRDs (both pipeline run and this repo's CRDs)

Hopefully some newbie user error someone from one of the 3 pillars can help me with

gabemontero:
ok I've reverted the run as user 0 change for the cache re https://github.com/redhat-appstudio/jvm-build-service/pull/236#issuecomment-1249679640 with @stuartwdouglas 's fix in #263

gabemontero:
@stuartwdouglas with my removal of runAsUser: 0 for the artifact cache (in favor of your fix in ) the artifact cache is no longer coming up in this PR:

`2022-09-19 06:04:47,962 ERROR [io.qua.run.Application] (main) Failed to start application (with profile prod): java.nio.file.AccessDeniedException: /cache/hacbs`

no need to retest this one until your #263 merges

deploy/kcp/apiexport_jvmbuildservice.yaml
gabemontero: fyi I've made this adjustment, along with the associated change needed apibinding that we discussed with @mmorhun in standup today and was confirmed during the cabal with the infra team

also been able to verify the receipt of events for our CRDs locally

gabemontero:
was able to make some progress today (see the Jira for details)

currently blocked on the creation of deployments for the artifact cache; awaiting a response to my initial query (this one to the infra team)

gabemontero:
OK the artifact cache is getting permission denied, even after @stuartwdouglas 's permission fixes that accompanied the removal of the side card merged

```
        2022-09-20 19:44:32,312 ERROR [io.qua.run.Application] (main) Failed to start application (with profile prod): java.nio.file.AccessDeniedException: /cache/hacbs
        	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:90)
        	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
        	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        	at java.base/sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:397)
        	at java.base/java.nio.file.Files.createDirectory(Files.java:700)
        	at java.base/java.nio.file.Files.createAndCheckIsDirectory(Files.java:807)
        	at java.base/java.nio.file.Files.createDirectories(Files.java:793)
        	at com.redhat.hacbs.artifactcache.services.client.ociregistry.OCIRegistryRepositoryClient.<init>(OCIRegistryRepositoryClient.java:78)
```

quite possibly my latest rebase missed something ??? .... I'll start looking at the code diffs, but if that stack trace ^^ gives you any clues @stuartwdouglas please do advise, thanks.

deploy/base/pipeline.yaml
```diff
@@ -16,4 +15,4 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: pipeline
-    namespace: test-jvm-namespace
+    namespace: jvm-build-service
```
gabemontero, line 18: so if everybody starts testing on openshift, the need for defining pipeline SAs goes away, as the openshift pipelines operator does this for you

WDYT @stuartwdouglas @dwalluck @phillip-kruger ?

in any event, this changes seems redundant with https://github.com/redhat-appstudio/jvm-build-service/pull/236/files#diff-1ff731e4f73618e50718e56e5a434022f7664b1f507d7c3579ddc4f8b44f65a4

deploy/operator/base/rbac.yaml
```diff
@@ -5,6 +5,10 @@ metadata:
   labels:
     rbac.authorization.k8s.io/aggregate-to-edit: "true"
 rules:
+  - apiGroups: ["security.openshift.io"]
```
gabemontero, line 8: I think this is a remnant of my attempt to fix the artifact cache perm issue

deploy/operator/base/rbac.yaml
```diff
@@ -179,6 +187,10 @@ metadata:
...
```
gabemontero, line 190: I think this is a remnant of my attempt to fix the artifact cache perm issue

should not be causing current issues I'm seeing in this PR

gabemontero:
yeah if anything, I'm accidentally increasing the privilege with these items ^^   though perhaps not to the cache service account

any ideas @stuartwdouglas around this file access denied to `/cache/hacbs` ?

I believe you said https://github.com/redhat-appstudio/jvm-build-service/pull/263/files#diff-1b6ca4cbe6cefa7cf43b86548eff0e6f292e36beb1abefb6fdb86397ef585952R160 was the fix to the permission issue before.  I don't see how this PR is affecting it.

thanks

deploy/operator/config/kustomization.yaml
```diff
@@ -6,6 +6,6 @@ commonLabels:
 resources:
   - system-config.yaml
 
-patchesStrategicMerge:
-  #note that when updating the ref here you must also update system-config.yaml
-  - https://raw.githubusercontent.com/redhat-appstudio/jvm-build-service-builder-images/fbe0f42ebe3d50b16d454c95207dd15be6b5008f/image-config.yaml
+#patchesStrategicMerge:
+#  #note that when updating the ref here you must also update system-config.yaml
+#  - https://raw.githubusercontent.com/redhat-appstudio/jvm-build-service-builder-images/fbe0f42ebe3d50b16d454c95207dd15be6b5008f/image-config.yaml
```
stuartwdouglas, line 11: Why this change? The intention here was that the tags would always match what was actually in the image with no possibility for user error when updating the images.

deploy/base/pipeline.yaml
stuartwdouglas: This is really minikube specific. If we still want to support it it should be part of the minikube scripts.

deploy/crds/base/jvmbuildservice.io_systemconfigs.yaml
```diff
@@ -0,0 +1,65 @@
...
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            properties:
+              jdk8image:
```
stuartwdouglas, line 38: It is highly likely we will have more images than these three in future. We should have something like a list of 'tag + image' based structs, they should not be hard coded.

deploy/operator/base/rbac.yaml
stuartwdouglas: This appears to be the only thing that could actually be affecting the cache, although I don't see how it would affect it.

deploy/operator/config/kustomization.yaml
stuartwdouglas: Uh, never mind, I see that it is temporary

deploy/crds/base/jvmbuildservice.io_systemconfigs.yaml
gabemontero: I had considered a map a la labels or annotations before "making it simpler"

I'll try that out in the next pass.

deploy/operator/base/rbac.yaml
gabemontero: unable to test locally (still in kcp debug) but I removed this remnant just to eliminate it as a variable 

let's see what CI says

deploy/operator/base/rbac.yaml
gabemontero: unable to test locally (still in kcp debug) but I removed this remnant just to eliminate it as a variable

let's see what CI says

deploy/base/pipeline.yaml
gabemontero: make sense @stuartwdouglas 

one follow up question - with minikube, why does the `pipeline` SA in the `test-jvm-namespace` need the permissions in the `pipeline` clusterrole, namely

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: pipeline
  labels:
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
rules:
  - apiGroups:
      - jvmbuildservice.io
    resources:
      - artifactbuilds
    verbs:
      - create
  - apiGroups:
      - ""
    resources:
      - configmaps
    resourceNames:
      - jvm-build-config
    verbs:
      - get
```

We are not applying that clusterrole in our openshift ci script.  

Is it because the aggregate-to-edit usage in the operator's rbac.yaml has no effect on minikube?

Assuming that is the case, I think moving to the minikube script entails
- creating the pipeline SA in the test ns from the script
- creating the role and role binding from the script as well, where I delete the files from ./deploy/base and use `cat <<EOF | kubectl apply -f - ..... EOF`

deploy/operator/base/rbac.yaml
gabemontero: sure enough, the e2e is passing again

openshift must have been performing some voodoo under the covers with that partial application of anyuid, but exactly what, who knows .... a mystery unfortunately I just do not have time to dive into

deploy/base/pipeline.yaml
gabemontero: this update has been pushed in the latest commit for review

gabemontero:
> looking at https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/236/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1572599833112350720 looks like the openshift pipeline install is not coming up (similar to Stuart's PR today).
> 
> I'm going to push my refactor for [#236 (comment)](https://github.com/redhat-appstudio/jvm-build-service/pull/236#discussion_r976645173) and see if that run has better luck. If it is a perm fail, I'll see about debugging that locally when I have a free env to do so.

all good, including the fact that the partial application of anyuid inexplicably appears to have been causing the perm issue with the artifact cache

deploy/crds/base/jvmbuildservice.io_systemconfigs.yaml
gabemontero: ok @stuartwdouglas I've updated this to be map based ... ptal

gabemontero:
So if I get clean e2e's (they passsed locally) and if you are good with the systemconfig related changes, @Michkov is interested in merging items like this, even if only partially functional / in progress, so we can make progress on the infra deployment pieces and test there as well.

We can then take on converting the user configmaps to `UserConfig`, working with infra team to get deployment creation working, and the "how we create and sync" discussions we started last night @stuartwdouglas, in subsequent PRs.

gabemontero:
ok all green @stuartwdouglas 

I do have to push a couple of small kcp yam file changes for infra team

pkg/reconciler/dependencybuild/dependencybuild_test.go
```diff
@@ -99,7 +104,7 @@ func TestStateNew(t *testing.T) {
 			Name:      "test",
 		}, &db))
 		g.Expect(db.Status.CurrentBuildRecipe).ShouldNot(BeNil())
-		g.Expect(db.Status.CurrentBuildRecipe.Image).Should(Equal("quay.io/redhat-appstudio/hacbs-jdk11-builder:latest"))
+		g.Expect(db.Status.CurrentBuildRecipe.Image).Should(HavePrefix("quay.io/redhat-appstudio/hacbs-jdk"))
```
gabemontero, line 107: this is a tweak I need to make to deal with the fact that the build recipe ordering is variable

pkg/reconciler/dependencybuild/dependencybuild_test.go
```diff
@@ -196,7 +201,7 @@ func TestStateDetect(t *testing.T) {
 				case PipelineScmUrl:
 					g.Expect(param.Value.StringVal).Should(Equal("some-url"))
 				case PipelineImage:
-					g.Expect(param.Value.StringVal).Should(Equal("quay.io/redhat-appstudio/hacbs-jdk11-builder:latest"))
+					g.Expect(param.Value.StringVal).Should(HavePrefix("quay.io/redhat-appstudio/hacbs-jdk"))
```
gabemontero, line 204: again ordering

pkg/reconciler/dependencybuild/dependencybuild_test.go
```diff
@@ -397,8 +402,18 @@ func TestStateDependencyBuildStateAnalyzeBuild(t *testing.T) {
 		db := getBuild(client, g)
 		g.Expect(db.Status.State).Should(Equal(v1alpha1.DependencyBuildStateSubmitBuild))
 		g.Expect(len(db.Status.PotentialBuildRecipes)).Should(Equal(2))
-		g.Expect(db.Status.PotentialBuildRecipes[0].Image).Should(Equal("quay.io/redhat-appstudio/hacbs-jdk11-builder:latest"))
-		g.Expect(db.Status.PotentialBuildRecipes[1].Image).Should(Equal("quay.io/redhat-appstudio/hacbs-jdk8-builder:latest"))
+		find11 := false
```
gabemontero, line 405: again ordering

pkg/reconciler/dependencybuild/dependencybuild_test.go
```diff
@@ -420,7 +435,17 @@ func TestStateDependencyBuildStateAnalyzeBuild(t *testing.T) {
...
```
gabemontero, line 438: again ordering

gabemontero (commented):
@stuartwdouglas PTAL at the in-line comments below

pkg/reconciler/dependencybuild/dependencybuild_test.go
stuartwdouglas: This is a bit of a regression, previously the ordering was defined by the 'builder-image.names' config map key. We probably need to add some kind of priority concept here.

stuartwdouglas:
I think we can merge this for now, and worry about the order later. It's definitely not a priority at this stage, and is relatively easy to fix later.

PR #239, opened 2022-09-07
Fix service registry test

stuartwdouglas:


gabemontero:
if the service registry rbac problem is timing related, maybe we get luck and 

will get somewhere

PR #245, opened 2022-09-08
tweaks/workarounds to get dev flow working on CI

gabemontero:
see https://github.com/openshift/release/pull/32127#issuecomment-1241162414

### DIFF
--- a/deploy/openshift-ci.sh
+++ b/deploy/openshift-ci.sh
@@ -7,3 +7,5 @@ echo "Running out of ${DIR}"
 $DIR/install-openshift-pipelines.sh
 oc apply -f $DIR/namespace.yaml
 $DIR/patch-yaml.sh
+oc apply -k $DIR/overlays/development
+
--- a/deploy/base-development.sh
+++ b/deploy/base-development.sh
@@ -19,3 +19,4 @@ JVM_BUILD_SERVICE_JDK8_BUILDER_IMAGE=quay.io/$QUAY_USERNAME/hacbs-jdk8-builder:d
 JVM_BUILD_SERVICE_JDK11_BUILDER_IMAGE=quay.io/$QUAY_USERNAME/hacbs-jdk11-builder:dev \
 JVM_BUILD_SERVICE_JDK17_BUILDER_IMAGE=quay.io/$QUAY_USERNAME/hacbs-jdk17-builder:dev \
 $DIR/patch-yaml.sh
+kubectl apply -k $DIR/overlays/development
--- a/deploy/base/config.yaml
+++ b/deploy/base/config.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: jvm-build-config
+  namespace: test-jvm-namespace
 data:
   enable-rebuilds: "true"
   maven-repository-300-jboss: "https://repository.jboss.org/nexus/content/groups/public/"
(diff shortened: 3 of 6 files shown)


deploy/overlays/dev-template/system-config.yaml
```diff
@@ -10,3 +10,7 @@ data:
   builder-image.jdk8.image: jdk8-builder
   builder-image.jdk11.image: jdk11-builder
   builder-image.jdk17.image: jdk17-builder
+  # pulled the tags from https://github.com/redhat-appstudio/infra-deployments/blob/pre-kcp/components/build/jvm-build-service/system-config.yaml#L13-L15
+  builder-image.jdk8.tags: jdk:8,maven:3.8,gradle:7.4.2;6.9.2;5.6.4;4.10.3
```
stuartwdouglas, line 14: These should be coming from https://github.com/stuartwdouglas/jvm-build-service/blob/4a78dab7188f16f1a175de169f55590cb4236508/deploy/base/system-config.yaml#L11 , and once the new builder repo is live it will reference the config file at https://github.com/redhat-appstudio/jvm-build-service-builder-images/blob/main/image-config.yaml

Why is this needed here?

stuartwdouglas (changes requested):

deploy/overlays/dev-template/system-config.yaml
gabemontero: ah cause I didn't realize that other copy was there

and sure, now that I see it there, I get where the kustomize merge comes in

and I'm still use to having *one* @#$%@#$% copy of a api obj yaml file :-)

thanks will remove

deploy/openshift-ci.sh
```diff
@@ -7,3 +7,5 @@ echo "Running out of ${DIR}"
 $DIR/install-openshift-pipelines.sh
 oc apply -f $DIR/namespace.yaml
 $DIR/patch-yaml.sh
+oc apply -k $DIR/overlays/development
```
stuartwdouglas, line 10: Are you sure you want this bit? A lot of the 'dev' overlay stuff is also setup in the test itself. e.g. this config map creation: https://github.com/redhat-appstudio/jvm-build-service/blob/857389d1fd39350509801fc67751211e9dd00d8e/openshift-with-appstudio-test/e2e/util.go#L202

The dev overlay also includes its own version, so if you are using the same namespace the test will fail.

gabemontero: yeah the creation of an extra config map in the `test-jvm-namespace` namespace the e2e's will not use seemed like a small cost compared to being able to reuse the dev flow as much as possible and get the image substitution 

Not sure what you mean by the dev overlay using its own version, but the e2e's will not use `test-jvm-namespace` at all, and use the namesapce that configmap you referenced is in.

Now, if you think that calling 

`oc apply -k $DIR/operator/overlays/development` is sufficient for our means, where the overrides for the controller, sidecar, and req processor get overridden (I think I can see where that is the case, as I think the other images are referenced by the controller, right?), then sure, I'll test that out on Friday and switch to that.

stuartwdouglas: I think you would still need the patches on the maven task. I guess this is fine.

gabemontero: forgot also add the crds and base/rbac as well via oc apply -f

PR #246, opened 2022-09-08
Switch to using builder images repo

stuartwdouglas:
This now references the images generated from the builder images repo. The Jobs to build the images in PAC have been removed, however the dockerfiles for openshift remain, as we need this merged before we can remove the openshift builds.


stuartwdouglas:
Lets see what CI says.

PR #253, opened 2022-09-11
Add all ABs for service registry

stuartwdouglas:
As a temporary measure to aid with debugging the service registry build this adds all the AB's generated by the build so they can be easily tested locally.


gabemontero:
ooh nifty / cool deal @stuartwdouglas 

@dwalluck check it out

PR #261, opened 2022-09-14
Add support for Gradle builds that use the legacy maven plugin

dwalluck:


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/gradle/GradleUtils.java
```diff
@@ -18,14 +18,14 @@
...
      */
-    public static final List<String> DEFAULT_GRADLE_ARGS = List.of("build", "publish");
+    public static final String GOOGLE_JAVA_FORMAT_PLUGIN = Pattern.quote("com.github.sherter.google-java-format");
 
     /**
-     * Identifier for the plugin {@code com.github.sherter.google-java-format}.
+     * Code for applying the plugin {@code maven}.
      */
-    public static final String GOOGLE_JAVA_FORMAT_PLUGIN = Pattern.quote("com.github.sherter.google-java-format");
+    public static final String MAVEN_PLUGIN = Pattern.quote("apply(plugin: \"maven\");");
```
stuartwdouglas, line 28: Shouldn't this account for possible whitespace differences? Something like:
`apply\\s*\\(\\s*plugin:\\s*\\"maven\\"`

stuartwdouglas (changes requested):

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/gradle/GradleUtils.java
dwalluck: It turns out that this matches all the builds we have for now.

Technically, it needs to be something like

```
apply\\s*\\[(]?\\s*plugin:\\s*["']maven["'][)]?
```

since I think either `"` or `'` can be used and parentheses are optional. But even that won't match everything, because there are several valid ways to import plugins and at two languages (Groovy and Kotlin).

stuartwdouglas:
Which builds are we expecting this to help with?

gabemontero:
> Which builds are we expecting this to help with?

yeah the counts look similar in the service registry build:  `types.go:25: time: 2022-09-14 21:53:54.081497292 +0000 UTC m=+3762.655224371: dependencybuild FINAL created count: 126 complete count: 93, failed count: 28, contaminated count: 4`

that said, the fact that it is passing here makes me wonder what happened with my local main branch run yesterday and my PR (though @stuartwdouglas did comment there on an interesting error I had missed that I need to look into) :-)

dwalluck:
> Which builds are we expecting this to help with?

There are three builds using the legacy maven plugin: com.github.java-json-tools:jackson-coreutils:2.0, com.github.java-json-tools:msg-simple:1.2, and com.github.java-json-tools:btf:1.3.

Unfortunately, though the `build` now happens, they are all still failing at `uploadArchives` because it looks like Gradle manipulator didn't modify that section which is something I need to look into.

PR #262, opened 2022-09-15
Fix logic error in Java version detection

stuartwdouglas:


dwalluck (approved):
I see that there was a logic error in the case of `-1`.

PR #264, opened 2022-09-19
Give wrappers the same owners as the pipeline

stuartwdouglas:
This is breaking the 'rebuild' functionality is the wrappers are not deleted when the DependencyBuild is deleted, so subsequent attempts to create the same pipeline run fail due to it already existing.

### DIFF
--- a/pkg/reconciler/tektonwrapper/util.go
+++ b/pkg/reconciler/tektonwrapper/util.go
@@ -27,6 +27,9 @@ func (b *BatchedCreate) CreateWrapperForPipelineRun(ctx context.Context, client 
 	tw.Namespace = run.Namespace
 	tw.Name = run.Name
 	tw.GenerateName = run.GenerateName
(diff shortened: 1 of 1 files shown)


gabemontero:
tekton never came up in the e2e

let's try again

gabemontero:
also curious how fixing rebuild helps with the service registry build

gabemontero:
openshift install AWS VPC fail with the apicurio run ... I'll check the openshift slack channels if this is a known situation or if this is just a flake

that said, the normal e2e passed; so I'll merge this before EOB today, but will try an apicurio if it makes sense at least one more time

gabemontero:
about the same I guess, give or take

`    types.go:25: time: 2022-09-19 18:30:14.743890445 +0000 UTC m=+3876.823370730: dependencybuild FINAL created count: 117 complete count: 93, failed count: 22, contaminated count: 1`

merging

PR #265, opened 2022-09-20
Merge maven-metadata.xml versions

stuartwdouglas:
If this is from multiple sources it needs to be merged.

Also add a Jenkins sample build that demonstrates this.

### DIFF
--- a/deploy/base/config.yaml
+++ b/deploy/base/config.yaml
@@ -14,4 +14,5 @@ data:
   maven-repository-305-redhat: "https://maven.repository.redhat.com/ga"
   maven-repository-306-gradleplugins: "https://plugins.gradle.org/m2"
   maven-repository-307-jsweet: "https://repository.jsweet.org/artifactory/libs-release-local"
+  maven-repository-308-jenkins: "https://repo.jenkins-ci.org/public/"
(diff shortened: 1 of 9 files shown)


stuartwdouglas:
@gabemontero it needed a few minor changes but the initial jenkins build completes now.

gabemontero (commented):
cool deal / thanks @stuartwdouglas 

I'll build on this after it merges and add the 3 openshift jenkins plugins 

Also, I'll try it soon enough, but after the initial build, what was the rough estimate of succeeded vs. failed vs. contaminated dependencies ?

/lgtm

gabemontero:
> cool deal / thanks @stuartwdouglas
> 
> I'll build on this after it merges and add the 3 openshift jenkins plugins
> 
> Also, I'll try it soon enough, but after the initial build, what was the rough estimate of succeeded vs. failed vs. contaminated dependencies ?
> 
> /lgtm

so yeah I see it compile as well ... took way longer than service registry.

I then see 97 artifactbuilds created, all either failed or missing

17 dependency builds, 2 contaminated, the rest failed.

So less of those than service registry, but starting at ground 0.

PR #266, opened 2022-09-20
PLNSRVCE-311 Remove dependency analyser image

stuartwdouglas:
- Move dependency analysis into the build-request-processor
- Move responsibility for creating ArtifactBuild objects to the cache
- Don't mount service account tokens for our pipelines


stuartwdouglas:
@gabemontero this PR will also require build-definitions changes and openshift CI changes in a similar manner to the sidecar one. If we can get this in as well it might be nice to handle both of them at once.

openshift-with-appstudio-test/e2e/jvm_build_service_test.go
```diff
@@ -45,15 +45,15 @@ func TestExampleRun(t *testing.T) {
...
 	if len(analyserImage) > 0 {
 		ta.Logf(fmt.Sprintf("PR analyzer image: %s", analyserImage))
-		for _, step := range ta.maven.Spec.Steps {
+		for i, step := range ta.maven.Spec.Steps {
 			if step.Name != "analyse-dependencies" {
 				continue
 			}
 			ta.Logf(fmt.Sprintf("Updating analyse-dependencies step with image %s", analyserImage))
-			step.Image = analyserImage
+			ta.maven.Spec.Steps[i].Image = analyserImage
```
stuartwdouglas, line 56: @gabemontero looks like we were not actually using the dependency analyser built in CI, but because it does not change much we did not pick up on it.

gabemontero: ah another forgot how golang slices work mistake (either in coding or review of another's PR) :-/

PR #268, opened 2022-09-21
Rename recipes directory to scm-info

stuartwdouglas:
The name no longer makes sense now that there are different directory structures.


gabemontero:
tekton never came up

PR #270, opened 2022-09-26
PLNSRVCE-658: fix deployment get/create, failed attempt to fix service get/create

gabemontero:
So with an assist from @Mo3m3n (I think I have that git ID correct, apologies if it was not), we were able to adjust the way jvm-build-service made its controller-runtime manager "kcp aware", which ended up fixing the HTTP URL used on the POST to create the deployments for the artifact cache and localstack.

With log level 6 still, you will see for example:

```
I0926 18:26:54.317774       1 round_trippers.go:553] GET https://ckcp-ckcp.apps.gmontero410.devcluster.openshift.com:443/services/apiexport/root:default:pipeline-service-compute/jvmbuildservice/clusters/%2A/apis/apps/v1/deployments?limit=500&resourceVersion=0 200 OK in 4 milliseconds
I0926 18:26:54.323793       1 round_trippers.go:553] GET https://ckcp-ckcp.apps.gmontero410.devcluster.openshift.com:443/services/apiexport/root:default:pipeline-service-compute/jvmbuildservice/clusters/%2A/apis/apps/v1/deployments?allowWatchBookmarks=true&resourceVersion=2590&timeoutSeconds=590&watch=true 200 OK in 5 milliseconds
I0926 18:26:54.413683       1 shared_informer.go:285] caches populated
1.6642168144138787e+09	INFO	userconfig	GetImageName controller image quay.io/gabemontero/hacbs-jvm-controller:dev	{"request": "jvm-build-service/jvm-build-config", "cluster": "root:default:pipeline-service-compute"}
1.664216814413896e+09	INFO	userconfig	GetImageName using quay.io/gabemontero/hacbs-jvm-cache:dev for hacbs-jvm-cache	{"request": "jvm-build-service/jvm-build-config", "cluster": "root:default:pipeline-service-compute"}
I0926 18:26:54.425015       1 round_trippers.go:553] POST https://ckcp-ckcp.apps.gmontero410.devcluster.openshift.com:443/services/apiexport/root:default:pipeline-service-compute/jvmbuildservice/clusters/root:default:pipeline-service-compute/apis/apps/v1/namespaces/jvm-build-service/deployments 201 Created in 10 milliseconds
```

I am even able to get/create some core k8s types, like serviceaccounts:

```
[...]

PR #272, opened 2022-09-27
KCP Fixes around results

stuartwdouglas:
This makes some fairly significant changes to the overall flow to enable KCP workspaces:

- Work around TR results limits by updating the results field directly via the API server
- Move deployment out of the cache to an task
- Move API updates into the controller all driven by pipeline run results

This is not 100% ready yet, but I am posting it here for early review.


hack/examples/pipeline-gradle.yaml
```diff
@@ -53,7 +53,7 @@ spec:
         - fetch-repository
       params:
         - name: MAVEN_MIRROR_URL
-          value: http://jvm-build-workspace-artifact-cache.$(context.taskRun.namespace).svc.cluster.local/v1/cache/default/0/
+          value: http://jvm-build-workspace-artifact-cache.$(context.taskRun.namespace).svc.cluster.local
```
gabemontero, line 56: I cannot say with certainty, but at least with the artifact cache referencing localstack, this reliable pattern of service level http URLs no longer held up, because the name of the namespace in kcp is mapped to a `kcp-<random id>` namespace on the workload cluster.

There is I'm sure (though I have not had a chance to look for / discover / analyze it yet, so between you and I we will need to sort that out) a solution in kcp for this.

hack/examples/pipeline.yaml
```diff
@@ -31,32 +34,14 @@ spec:
...
-    #       workspace: shared-workspace
     - name: maven-run
       taskRef:
         name: maven
       runAfter:
         - fetch-repository
       params:
         - name: MAVEN_MIRROR_URL
-          value: http://jvm-build-workspace-artifact-cache.$(context.taskRun.namespace).svc.cluster.local/v1/cache/default/0/
+          value: http://jvm-build-workspace-artifact-cache.$(context.taskRun.namespace).svc.cluster.local
```
gabemontero, line 44: same svc url in kcp work/investigation needed comment

pkg/controller/controller.go
```diff
@@ -72,7 +71,7 @@ func NewManager(cfg *rest.Config, options ctrl.Options, kcp bool) (ctrl.Manager,
 	c := rest.CopyConfig(cfg)
 	cacheOptions := cache.Options{
 		SelectorsByObject: cache.SelectorsByObject{
-			&pipelinev1beta1.PipelineRun{}: {Label: labels.SelectorFromSet(map[string]string{artifactbuild.PipelineRunLabel: ""})},
+			&pipelinev1beta1.PipelineRun{}: {},
```
gabemontero, line 74: fwiw, per our discussion my Monday evening / your Tuesday AM @stuartwdouglas, even with the removal of this label, the fact that we are using selectors by object at all means I had to go down the path of solving "atypical" integration between controller runtime and kcp, so that pain/effort was no wasted :-)

pkg/reconciler/dependencybuild/scripts/deploy.sh
```diff
@@ -1,5 +0,0 @@
-#!/bin/sh
```
gabemontero, line 1: so I'm guessing that the `Move deployment out of the cache to an task` highlight in this PR's description includes this deletion, along with the addition of these tar's to the specific gradle / mavne scripts below

is that correct @stuartwdouglas ?

gabemontero (commented):
Some preliminary comments and questions @stuartwdouglas 

Obviously with this still WIP, I did not call out for changes.  Just tried to understand, as well as highlight possible concerns down the road as this progresses.

Aside from the in-line comments, with the addition of the create artifactbuilds permission to the cache, it made me wonder if since it is already watching resources, were you thinking of making the step of converting it to another controller-runtime based controller?

Among possible advantages of that, we know now how to add the kcp layer on top of controller runtime setup so that the creates of the artifactbuilds from directly from the cache code will be wired as such that they are visible from kcp.

pkg/controller/controller.go
stuartwdouglas: This is only because we need some way of looking at the 'initial' build that kicks off the whole process. I need to look at its results so I can use them to create the ArtifactRequest objects.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -41,6 +40,29 @@ func createPipelineSpec(maven bool, namespace string, commitTime int64) *pipelin
 		build = gradleBuild
 	}
 	zero := int64(0)
+	deployArgs := []string{
```
gabemontero, line 43: another education question ... why do we need this config in the pipelinespec no?  was this an overisght when I moved off of configmaps?

if not, I'm missing how the kcp results pivot necessitates this.  As with some of my other question, I'm not saying this is wrong, I just need help understanding.  Thanks.

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -559,9 +565,37 @@ func (r *ReconcileDependencyBuild) handleBuildPipelineRunReceived(ctx context.Co
 
 		success := pr.Status.GetCondition(apis.ConditionSucceeded).IsTrue()
 		if success {
+			for _, i := range pr.Status.PipelineResults {
```
gabemontero, line 568: ah ok so the java writes to the results, and the golang controller consumes and updates our CRD instances' statuses 

sure, I think that make sense now when I revisit our strategizing discussions :-)

gabemontero (commented):
A series of questions vs. any comments around things that should change.

The one question about config map selector is the one that overlaps with my comment in your bloom filter commit 

Also, I could still use an answer to my question https://github.com/redhat-appstudio/jvm-build-service/pull/272#discussion_r981231429 from the last review

thanks

pkg/controller/controller.go
stuartwdouglas: I can do it here

pkg/reconciler/dependencybuild/scripts/deploy.sh
stuartwdouglas: Yes, this is now a step in the golang code that calls the 'build-request-processor' with specific args to do the deployment.

pkg/controller/controller.go
gabemontero: cool thanks

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/dependencies/RebuildService.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Inject
+    Instance<KubernetesClient> client;
+
+    public void rebuild(String taskRunName, Set<String> gavs) {
+
+        Log.infof("Identified %s Community Dependencies: %s", gavs.size(), new TreeSet<>(gavs));
+
+        var client = this.client.get();
+        var resources = client.resources(TaskRun.class);
```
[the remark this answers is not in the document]
stuartwdouglas: As mentioned in the description this bit is still broken with regards to KCP. This is the bit that the 'bloom filter' approach is trying to solve.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/dependencies/RebuildService.java
```diff
@@ -0,0 +1,38 @@
...
+        resources.withName(taskRunName).editStatus(new UnaryOperator<TaskRun>() {
+            @Override
+            public TaskRun apply(TaskRun taskRun) {
+                List<TaskRunResult> results = new ArrayList<>();
+                if (taskRun.getStatus().getTaskResults() != null) {
+                    results.addAll(taskRun.getStatus().getTaskResults());
+                }
+                results.add(new TaskRunResult("java-community-dependencies", String.join(",", gavs)));
+                taskRun.getStatus().setTaskResults(results);
```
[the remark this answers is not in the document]
stuartwdouglas: It can't happen in the controller, the controller is in KCP land, the actual task is running in the compute cluster. We need some way to propagate this information back to the KCP control plane and at the moment all we have is results (without say creating a shared S3 account and sticking this in a bucket).

The controller does act on these results once they are back in the cluster, but it has no visibility into the compute cluster where the tasks are being run.

pkg/reconciler/artifactbuild/artifactbuild.go
```diff
@@ -595,3 +606,27 @@ func (r *ReconcileArtifactBuild) createLookupScmInfoTask(ctx context.Context, lo
 		},
 	}, nil
 }
+
+func (r *ReconcileArtifactBuild) handleCommunityDependencies(ctx context.Context, split []string, namespace string, log logr.Logger) error {
```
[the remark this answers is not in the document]
stuartwdouglas: We already had code to create it for the contaminants, so that is not actually part of this pr, this just move the updating of the 'contaminants' field on the DependencyBuild from the Java code to something done by the controller. Once this field is updated the controller will automatically act on it.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
stuartwdouglas: Basically now that we want to stop the cache using the Java kube API as much as possible there is no need to involve it in the deployment process at all, so it has just been moved to a discrete step on task. The results of the deployment need to be expressed as results on the PipelineRun, so there were really two choices:

- Post deployment to cache, cache deploys, returns results as JSON, write JSON status update to results
- Deploy directly from pipeline

The originally moved to cache vs sidecar as it would let us avoid mounting the service token in the PR, however that is no longer an option as we need it for the results.

This approach also means that any problems are logged in the PR instead of just getting a 500 response, and any additional memory needed by the deploy step is only used briefly during the deploy process.

The second approach is much simpler ()

pkg/reconciler/dependencybuild/dependencybuild.go
stuartwdouglas: It's the only way I can think of to actually propagate this information back without involving some form of external service.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/dependencies/RebuildService.java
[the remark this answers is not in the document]
gabemontero: ack

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/dependencies/RebuildService.java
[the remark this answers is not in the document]
gabemontero: yea per my comment below at https://github.com/redhat-appstudio/jvm-build-service/pull/272#discussion_r982525113 once I saw the golang code on the other side I course corrected and connected the dots

pkg/reconciler/dependencybuild/buildrecipeyaml.go
gabemontero: OK I think I missed where we moved this stuff off of java code 

Makes sense / I'll see if I can connect those dots when I revisit the tomorrow after processing the responses here

pkg/reconciler/dependencybuild/dependencybuild.go
gabemontero: yep that's right

gabemontero:
tekton bootstrap failed ... let's try again

hack/examples/pipeline-gradle.yaml
gabemontero: so I've talked to a few folks now and am *pretty sure* we'll need to employ ingress and the haproxy stuff like Frederic is doing for the pipeline as code integration in pr 232 in https://github.com/openshift-pipelines/pipeline-service

this new ingress route / haproxy combo should allow a "non kcp aware" client within the initial pipeline run to access the artifact cache deployment our controller created when it sees the userconfig in the same namesapce the initial pipelinerun is in

diving into this is next on my list after getting #270 properly vetted and merged, and then seeing about our integration in infra-deployments

stuartwdouglas:
So this failure does seem to be an actual issue, although i am not sure why as the e2e build works fine for me locally. I will investigate today.

stuartwdouglas:
So I have seen the failures locally, however they appear to be racey, and as soon as I added more logging I stopped getting them. 

I am running the PR with more logging to try and figure out what is going on.

stuartwdouglas:
@gabemontero I think the extra logging I added might be making it run out of memory generating the status report. I have decreased the logging but we may need to change the report code to stream the logs instead of loading them all into memory.

PR #274, opened 2022-09-29
Use bloom filter for caching decisions

stuartwdouglas:
This is based on top of and also includes the commits from https://github.com/redhat-appstudio/jvm-build-service/pull/272


pkg/reconciler/rebuiltartifact/rebuiltartifact.go
```diff
@@ -0,0 +1,102 @@
...
+				hash = hash * -1
+			}
+			var pos = hash / 8
+			var bit = hash % 8
+			filter[pos] = filter[pos] | (1 << bit)
+		}
+	}
+	log.Info("Constructed bloom filter", "filterLength", len(filter))
+	cm := v1.ConfigMap{}
+	err = r.client.Get(ctx, types.NamespacedName{Namespace: request.Namespace, Name: "jvm-build-service-filter"}, &cm)
```
gabemontero: I mentioned this is slack too, but I figured out a better way to make it so that controller runtime does not create a cluster scoped watch / shared informer for configmaps as a result of this call

we'll just add configmaps to the list my KCP PR is curating at https://github.com/redhat-appstudio/jvm-build-service/blob/c4ae49bc3f44458a8b240251c08064139557cfd9/pkg/controller/controller.go#L79-L85

deploy/operator/base/rbac.yaml
```diff
@@ -97,6 +97,18 @@ rules:
...
+  - apiGroups:
+    - ""
+    resources:
+      - configmaps
+    verbs:
+      - get
+      - create
+      - patch
+      - list
+      - watch
```
gabemontero, line 109: why does the controller need watch here since it is just calling get, especially if we add configmap to the list of types we disable for the controller runtime cache like https://github.com/redhat-appstudio/jvm-build-service/blob/c4ae49bc3f44458a8b240251c08064139557cfd9/pkg/controller/controller.go#L79-L85

so long term if we will use the pipeline SA for the cache, since RebuiltArtifacts.java does the watch, we just need to make sure it can do the CM watch. which it might have already.  And if it doesn't, lets create a new role / rolebinding that we aggregate to edit, to the pipeline SA gets it that way

gabemontero (commented):
ok I think I officially get, and hence really like, the bloom filter (though I'm not exactly clear on the use of the adjective "bloom" ;-) ) 

an rbac comment and a second pass with the better way of doing the non caching client access for configmaps in the rebuilt artifact reconclier

also re-read the exchanges to my  questions in #272 and I think "I'm good"

gabemontero:
looks like the shaded dependency in the e2e did not get sorted out though the reason why was not obvious from my very quick scan

PR #277, opened 2022-09-30
Don't hard code the namespace into the task

stuartwdouglas:
This should fix a KCP issue with pipeline runs not being able to talk to the cache.


stuartwdouglas:
@gabemontero I think this is actually all we need, as the namespace now should not actually be resolved until the task is actually running on the compute cluster. 

In theory I think we now potentially have solutions for all the KCP issues, so it will be interesting to see how things go when we get all these fixes in and test them out together.

hack/examples/pipeline-gradle.yaml
```diff
@@ -53,7 +53,7 @@ spec:
         - fetch-repository
       params:
         - name: MAVEN_MIRROR_URL
-          value: http://jvm-build-workspace-artifact-cache.$(context.taskRun.namespace).svc.cluster.local/v1/cache/default/0/
+          value: http://jvm-build-workspace-artifact-cache.$(context.taskRun.namespace).svc.cluster.local
```
gabemontero, line 56: so the issue isn't only whether or not the `$(context.taskRun.namespace)` is correct or not based on what I was told yesterday

the service you are trying to reach may be in the same logical workspace / namespace, but may be on a different compute cluster

hence the classic `svc.cluster.local` DNS won't work

hence we need ingress / haproxy similar to what Frederic demo'ed for PaC and github webhooks.

Now, short term, with only 1 compute cluster, sure this might help, so for our initial onboarding efforts I think we may be OK.  But we can't wait long to the next step on ingress / haproxy.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -77,8 +77,7 @@ func createPipelineSpec(maven bool, namespace string, commitTime int64, userConf
 			{Name: PipelineEnforceVersion, Type: pipelinev1beta1.ParamTypeString},
 			{Name: PipelineRequestProcessorImage, Type: pipelinev1beta1.ParamTypeString},
 			{Name: PipelineIgnoredArtifacts, Type: pipelinev1beta1.ParamTypeString},
-			{Name: PipelineCacheUrl, Type: pipelinev1beta1.ParamTypeString, Default: &pipelinev1beta1.ArrayOrString{Type: pipelinev1beta1.ParamTypeString, StringVal: "http://jvm-build-workspace-artifact-cache." + namespace + ".svc.cluster.local/v1/cache/default/" + strconv.FormatInt(commitTime, 10)}},
-			{Name: "NAMESPACE", Type: pipelinev1beta1.ParamTypeString, Default: &pipelinev1beta1.ArrayOrString{Type: pipelinev1beta1.ParamTypeString, StringVal: namespace}},
+			{Name: PipelineCacheUrl, Type: pipelinev1beta1.ParamTypeString, Default: &pipelinev1beta1.ArrayOrString{Type: pipelinev1beta1.ParamTypeString, StringVal: "http://jvm-build-workspace-artifact-cache.$(context.pipelineRun.namespace).svc.cluster.local/v1/cache/default/" + strconv.FormatInt(commitTime, 10)}},
```
gabemontero, line 80: ok yeah looking at the "don't hard code the namespace" commit, same comment - as was explained to me, though I have not even attempted to set up a KCP env with muliple compute clustes, it is entirely possible that the artiffact cache that a KCP client may create in the same logical workspace/namespace as this pipeline run may land on a different cluster

and as such, the k8s local service HTTP URL convention will fail, as the DNS / route for that does not span compute clusters

gabemontero: but given that, without having ever implemented controller access to an ingress route before, I suspect these changes help us go in that direction.

hack/examples/pipeline-gradle.yaml
stuartwdouglas: Then we also need to re-think the bloom filter approach too, as that assumes that the configmap ends up in the same namespace as the cache. 

If we can't really co-locate anything then this is going to be very difficult. 

Also if the cache is now going to be accessible from an external URL we need to re-think its security. We are going to need to add some kind of authentication.

stuartwdouglas: It will also be an issue for security, as we need the secrets to be in the same namespace as the cache.

gabemontero: yeah when I found this out there was not a lot of detail around how the types where sharded 

I'll attempt to drill down on that next week.

And like I said, regardless, I think this change is positive regardless

PR #283, opened 2022-10-03
PLNSRVCE-687: Add Gradle plugin to get Java version and plugins

dwalluck:


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/gradle/GradleUtils.java
```diff
@@ -25,72 +18,28 @@ public final class GradleUtils {
...
-    public static Optional<String> getGradleVersionFromWrapperProperties(Path propertiesFile) {
-        try (Reader reader = Files.newBufferedReader(propertiesFile)) {
-            var properties = new Properties();
-            properties.load(reader);
-            return getGradleVersionFromWrapperProperties(properties);
-        } catch (IOException e) {
-            Log.errorf("Error reading gradle-wrapper.properties: %s", e.getMessage(), e);
-            return Optional.empty();
+        if (version.compareTo(GradleVersion.version("4.0")) < 0) {
+            throw new IllegalArgumentException("Unsupported Gradle version: " + gradleVersion);
```
stuartwdouglas, line 35: Does this really need to throw an exception? If there is any chance projects using these old versions will build we should still attempt it with our oldest supported version.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/gradle/GradleUtils.java
```diff
@@ -25,72 +18,28 @@ public final class GradleUtils {
     /**
      * Identifier for the plugin {@code com.github.sherter.google-java-format}.
      */
-    public static final String GOOGLE_JAVA_FORMAT_PLUGIN = Pattern.quote("com.github.sherter.google-java-format");
+    public static final String GOOGLE_JAVA_FORMAT_PLUGIN = "com.github.sherter.googlejavaformatgradleplugin.GoogleJavaFormatPlugin";
```
stuartwdouglas, line 21: I know it was not added in this PR, but why does the presence of this plugin trigger JDK11? Does this require JDK 11 as a baseline?

dwalluck: It doesn't work on Java 17, at least out of the box, so it's just meant to try Java 11 first.

It should be possible to support it on Java 17 with JVM args with the right `--add-exports` flags.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/gradle/GradleUtils.java
stuartwdouglas: If it is a simple project it might still build with newer version though right? I am not sure how much the build scripts changed between those versions. 

Either way there is no harm in trying.

stuartwdouglas:
So the old pre-kcp sidecar is failing, and I am not really sure why, as there is nothing obvious in this change that would affect it. Maybe we could increase the memory in the pre-kcp branch and see if that helps.

PR #284, opened 2022-10-04
Further resource leak fixes

stuartwdouglas:


stuartwdouglas:
@gabemontero service registry build has passed, I am going to merge this and rebase all our PR's on top of it.

gabemontero:
> @gabemontero service registry build has passed, I am going to merge this and rebase all our PR's on top of it.

awesome / sounds good @stuartwdouglas 

If you have not already, I'll bump the test timeout back out to 2 hours.  It looks like the final test completed in about 70 minutes.  Feels like maybe 90 minutes is too fine a margin if more rebuilds kick in over time.

`    types.go:25: time: 2022-10-04 06:34:54.5519995 +0000 UTC m=+4213.842270335: dependencybuild FINAL created count: 116 complete count: 89, failed count: 25, contaminated count: 1`

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/util/RequestCleanup.java
```diff
@@ -0,0 +1,45 @@
...
+
+import io.quarkus.arc.Arc;
+import io.quarkus.arc.Unremovable;
+import io.quarkus.logging.Log;
+
+/**
+ * A utility to ensure resources are correctly cleaned up at the end of the request.
+ * <p>
+ * This is useful as some HTTP client resources have lifecycles that are not super well defined so
+ * a try/finally block cannot be used. This ensures that these resources are correctly cleaned up.
```
gabemontero, line 19: interesting .... fwiw I at least have not seen the analogous situation pop up with the golang k8s clients (which are just http clients under the covers).  Wonder if that is a difference in the clients, or my lack of exposure to intermittent problems out there.

stuartwdouglas: This was actually an issue with the HTTP Maven client, not the golang one. I think the leak is actually me just missing a close, I will update the 'async maven-metadata.xml' PR to deal with it properly and remove this.

Basically you can't just use these resources in a try/finally block, as the results of the HTTP connection may have to 'bubble up (in some cases being streamed straight back to the client if we decide not to cache)'.

gabemontero: ok thanks for the confirmation, as I suspected, that this was specific to the client you were using / how you were using it

PR #288, opened 2022-10-04
Use multiple threads for maven-metadata.xml

stuartwdouglas:
Otherwise generating this file can be quite slow the first time it is accessed.

The HTTP connections were also not being closed, which was probably the root cause of the resource leak.


stuartwdouglas:
/hold

Failure is related

PR #290, opened 2022-10-04
More periodic upds

gabemontero:
I noticed in recent periodic run that 1 DB was still in running state when the test declared success @stuartwdouglas 

Here's an (untested locally as of yet) attempt to start tracking DBs in building state and make 0 DB's in building part of the "success" criteria.

### DIFF
--- a/openshift-with-appstudio-test/e2e/periodic_test.go
+++ b/openshift-with-appstudio-test/e2e/periodic_test.go
@@ -169,7 +169,7 @@ func TestServiceRegistry(t *testing.T) {
 		state := sync.Map{}
 		var changed uint32
 		exitForLoop := false
-		timeoutChannel := time.After(90*time.Minute)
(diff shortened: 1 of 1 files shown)


stuartwdouglas:
I have seen this in a recent run, and I know why this is happening.

- Because of the `maven-metadata.xml` rewriting Netty has become the slowest build. This is because it runs plugins on every module to compare the api to the previous release. This uses a different maven-metadata.xml for each module, and Netty has a huge amount of releases, so this is quite slow. 
- Netty is contaminated, so when the Netty build finishes it kicks of JCTools. However once the Netty build is done there is now 0 pipelines running for a period of time, so the counter will hit 0 and the test will be done.
- The JCTools build then gets kicked off again. 

The solution to this is to fix the slow Netty build for now.

stuartwdouglas:
Even though this won't fix the JCTools still building issue I think it is worth having the count anyway.

PR #291, opened 2022-10-04
Don't run japicmp as part of the build

stuartwdouglas:


stuartwdouglas:
This failure is actually a result of an earlier PR, so this can be merged.

PR #292, opened 2022-10-05
Add preprocessor tasks

stuartwdouglas:
There are some things that require physical manipulation of the build files that are not really possible with PME/GME (e.g. invalid repositories will cause GME to fail before it can remove them).

This also allows us to do transformations that are not supported by PME/GME.


stuartwdouglas:
/hold

This needs some more work before it is ready

stuartwdouglas:

I have updated this to just do the Maven transformations. Gradle is still a work in progress, however this should help us out with some of the remaining maven failures.

stuartwdouglas:
This seems to make things a lot better, down to 20 build failures now, so I am merging.

PR #295, opened 2022-10-05
Remove old modules and images

stuartwdouglas:
The sidecar and analyser in CI dockerfiles have been emptied as they are still referenced, everything else has been removed.


gabemontero:
@stuartwdouglas based on what you did here, I think if we remove the analyzer refs in https://github.com/redhat-appstudio/jvm-build-service/blob/main/.ci/common-ci-settings-and-functions.sh#L58-L60 we can then make the openshift/release change to remove sidecar / analyzer there, and then we can remove those ci dockerfiles completely.

I'm still getting caught up on PRs, but assuming I don't see one for that, I'l get that going today.

PR #296, opened 2022-10-05
Make sure artifact states are updated on rebuild

stuartwdouglas:
If only a single ABR from a DependencyBuild is annotated we should annotated them all so their state will be updated. Otherwise they will not be added as owners and their state will not be updated if the build passes.

### DIFF
--- a/pkg/reconciler/artifactbuild/artifactbuild.go
+++ b/pkg/reconciler/artifactbuild/artifactbuild.go
@@ -126,9 +126,14 @@ func (r *ReconcileArtifactBuild) Reconcile(ctx context.Context, request reconcil
 		//log.Info("cluster set on obj ", r.clusterSetOnObj(&abr))
 		//first check for a rebuild annotation
 		if abr.Annotations[Rebuild] == "true" {
-			return r.handleRebuild(ctx, &abr)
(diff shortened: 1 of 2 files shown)


gabemontero (commented):
not directly related to this change, but looking at this part of the code reminds me of a question that has occurred to me in the past.

the update of the status to new, followed by immediately by the removal of the annotation, it makes we wonder about potential race conditions as the events for those 2 separate updates arrive at https://github.com/redhat-appstudio/jvm-build-service/blob/7a8dc12d982a19077911f4852c57bea88cab4dae/pkg/reconciler/artifactbuild/artifactbuild.go#L128-L141

So I could see these timing scenarios:
1) the update of status to new happens, then event fires, and before the removal of the rebuild annotation occurs, the event the status new event results in a call to handleRebuild (which is where we started, and leads to another potential set to new, update status, etc.

should we perhaps see if the state is already new before doing that status update?
and similarly check if the annotation is there before deleting it?

*maybe* updating the status to the same state will not result in an event occuring again, but that feels like it assumes implementation details of controller-runtime, etc.

2) both the updates status to new and the delete occur before the event is processed, when the event occurs, since the annotation is going, it should flow straight into new

stuartwdouglas:
Done

gabemontero (approved):
/lgtm

PR #299, opened 2022-10-06
Don't wait forever for tekton

stuartwdouglas:


gabemontero:
see if we can get clean PaC runs before I merge anyway

/retest

PR #306, opened 2022-10-10
Add gradle repositories init script

stuartwdouglas:
This change adds an init script to the gradle build that will replace all repositories with the HACBS cache repository.

Presently some artifacts were downloaded from central, which mean't that contamination tracking was not working correctly.


pkg/reconciler/dependencybuild/scripts/gradle-build.sh
```diff
@@ -34,13 +34,20 @@ export LC_ALL="en_US.UTF-8"
 ADDITIONAL_ARGS=$(echo "$@" | sed 's/build publish \?//')
 echo ADDITIONAL_ARGS="${ADDITIONAL_ARGS}"
 
+INIT_SCRIPTS=""
+for i in .hacbs-init/*
+do
+  INIT_SCRIPTS="$INIT_SCRIPTS -I $i"
```
dwalluck, line 40: Do we want to number the scripts so that there's a set order to them in case there is more than one?

pkg/reconciler/dependencybuild/scripts/gradle-build.sh
```diff
@@ -34,13 +34,20 @@ export LC_ALL="en_US.UTF-8"
...
+done
+echo "INIT SCRIPTS: $INIT_SCRIPTS"
+
 if [ -n "$(params.ENFORCE_VERSION)" ]; then
     gradle-manipulator --no-colour --info --stacktrace -l "${GRADLE_HOME}" $(params.GRADLE_MANIPULATOR_ARGS) -DversionOverride=$(params.ENFORCE_VERSION) "${ADDITIONAL_ARGS}" generateAlignmentMetadata || exit 1
 else
-    gradle-manipulator --no-colour --info --stacktrace -l "${GRADLE_HOME}" $(params.GRADLE_MANIPULATOR_ARGS) "${ADDITIONAL_ARGS}" generateAlignmentMetadata || exit 1
+    gradle-manipulator $INIT_SCRIPTS --no-colour --info --stacktrace -l "${GRADLE_HOME}" $(params.GRADLE_MANIPULATOR_ARGS) "${ADDITIONAL_ARGS}" generateAlignmentMetadata || exit 1
```
dwalluck, line 47: This argument should be added to line 45 as well.

stuartwdouglas: Done

pkg/reconciler/dependencybuild/scripts/gradle-build.sh
stuartwdouglas: The shell expansion will do them alphabetically anyway, so if we start adding more we can order them in the code that actually copies them to disk.

PR #307, opened 2022-10-11
Don't mark AB as succeeded if it is not deployed

stuartwdouglas:
Check that the output of the build matches what is expected from the ArtifactBuild objects, and fail the AB if the build did not produce this artifact.

I noticed that jackson-databind was missing its info, but being lumped into the jackson-core build and marked as a success.

### DIFF
--- a/pkg/reconciler/dependencybuild/dependencybuild.go
+++ b/pkg/reconciler/dependencybuild/dependencybuild.go
@@ -582,6 +582,7 @@ func (r *ReconcileDependencyBuild) handleBuildPipelineRunReceived(ctx context.Co
 				} else if i.Name == artifactbuild.DeployedResources {
 					//we need to create 'DeployedArtifact' resources for the objects that were deployed
 					deployed := strings.Split(i.Value, ",")
+					db.Status.DeployedArtifacts = deployed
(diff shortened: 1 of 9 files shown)


pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -582,6 +582,7 @@ func (r *ReconcileDependencyBuild) handleBuildPipelineRunReceived(ctx context.Co
 				} else if i.Name == artifactbuild.DeployedResources {
 					//we need to create 'DeployedArtifact' resources for the objects that were deployed
 					deployed := strings.Split(i.Value, ",")
+					db.Status.DeployedArtifacts = deployed
```
gabemontero, line 585: a unit test to verify this field is getting set?

gabemontero (commented):
just a minor unit test comment from me

pkg/reconciler/dependencybuild/dependencybuild.go
stuartwdouglas: Done

gabemontero (approved):
/lgtm

PR #315, opened 2022-10-13
Improve maven plugin regex

dwalluck:


stuartwdouglas:
Awesome, looks like Kafka worked.

PR #323, opened 2022-10-17
PLNSRVCE-762: activate ResourceQuota watch on KCP; rm remaining deployment related claims on KCP

gabemontero:


gabemontero:
I need to verify via controller logs/events that the apicurio test shows the proper throttling behavior based on the openshfit pod count we use in that e2e

/hold

gabemontero:
ok throttle events fired, apicurio test results look as expected

now, while that as running, I found a timing but with low pod quota counts while testing on KCP, pushing a fix now and making sure it is OK in CI with bigger pod quota counts

gabemontero:
throttling events etc. still good

this is ready to merge from my end @stuartwdouglas if you want to take a look

PR #325, opened 2022-10-17
Add -DdokkaPlugin=false to GME args

dwalluck:


stuartwdouglas:
I just noticed the new GME release. I still think this should be optional for now though, as once we start dealing with network isolation dokka will cause problems.

PR #327, opened 2022-10-18
PLNSRVCE-557 Allow build pipeline limits to be configured

stuartwdouglas:


stuartwdouglas:
This is one of the last things still open in [HACBS-6](https://issues.redhat.com//browse/HACBS-6)

gabemontero (approved):
yep this change also would lend it self to some coordination or follow up PR to expose these setting as changeable parameters in my helm template PR https://github.com/redhat-appstudio/jvm-build-service/pull/329 @stuartwdouglas

PR #329, opened 2022-10-18
PLNSRVCE-762: add prelim helm charts for workspace resource controller integration

gabemontero:
So this is the first pass at scripting for what I have been doing (along with bootstrapping from infra deployments with local copies of https://github.com/redhat-appstudio/infra-deployments/pull/827) to test on KCP.

In this first pass, I chose Helm (vs. Kustomize) based on a sidebar conversation I had with @sbose78 and where he might be steering the wokspace resource controller work that @rajivnathan has.

That said, we can always pivot from Helm to whatever easily enough.

In addition to the workspace controller work, there is also a dependency on the discussions we've had in the cabal around builder service accounts, where short term, I follow the stop gap approach the build-service controller employs of creating the `pipeline` SA in KCP, but ultimately, the direction being established in https://github.com/redhat-appstudio/book/pull/5 or follow on to that work is also needed long term for setting up jvm build service in a user hacbs workspace.

Thus, as you'll see here, we have to apply rbac on both the kcp and compute cluster sides to facilitate 
- the artifact cache needing to access configmaps for @stuartwdouglas 's bloom filter for cache image registry content
- the dependencybuild pipeline runs needs to update tekton status (namely the results) so that the controller can propogate the state machine accordingly

Other details
- I downloaded helm via curl -L https://mirror.openshift.com/pub/openshift-v4/clients/helm/latest/helm-linux-amd64 -o /usr/local/bin/helm
- my infra-deployments `preview.env` sets
- - `export ROOT_WORKSPACE="~"`
- - `export CLUSTER_KUBECONFIG=<path to compute cluster kubeconfig`
- - `export KCP_KUBECONFIG=/tmp/ckcp-admin.kubeconfig` ..... I use the install ckcp script in infra-deployments for bring up kcp
- - `export PIPELINE_SERVICE_LOCAL_DEPLOY=true`
- - also the gitops `MY_GITHUB*` env vars are still needed to fully bootstrap appstudio
- my new scripts reference the 2 KUBECONFIG env vars from `preview.env`

First, run `./deploy/init-hacbs-user-workspace-sa-rbac.sh ` against your compute/workload cluster.

Then, I left these bits as manual (feels like something the workspace resource controller handles independently of jvm build service)
- from KCP, enter the `hacbs` workspace in the infra-deployments bootstrapped env, i.e. `oc ws hacbs` 
- from KCP, create a test namesapce i.e. `oc create ns jvm-bld-test`
- from KCP, enter that new test namespace i.e. `oc project jvm-bld-test` or `kubectl config set-context --current --namespace=jvm-bld-test` and then run `./deploy/init-hacbs-user-workspace.sh `

### DIFF
--- a/deploy/init-hacbs-user-workspace-sa-rbac.sh
+++ b/deploy/init-hacbs-user-workspace-sa-rbac.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+DIR=`dirname $0`
+#COMPUTE_KUBECONFIG should match whatever you have set in your infra-deployments preview.env file
+#HABCS_WORKSPACE_NAMESPACE is the name of whatever namespace in the KCP hacbs ws that you ran init-hacbs-user-workspace.sh from
+KUBECONFIG=$COMPUTE_KUBECONFIG helm install --set kcpNamespace=$HABCS_WORKSPACE_NAMESPACE $DIR/kcp-hacbs-workspace-compute-cluster-rbac --debug --generate-name
--- a/deploy/kcp-hacbs-workspace-compute-cluster-rbac/Chart.yaml
+++ b/deploy/kcp-hacbs-workspace-compute-cluster-rbac/Chart.yaml
@@ -1,0 +1,6 @@
+apiVersion: v2
+appVersion: 1.16.0
+description: A Helm chart for initializing a hacbs user workspace's compute cluster rbac for jvm build service
+name: hacbs-user-workspace-jvm-build-service-compute-cluster-rbac
+type: application
+version: 0.1.0
--- a/deploy/kcp-hacbs-workspace-init/templates/userconfig.yaml
+++ b/deploy/kcp-hacbs-workspace-init/templates/userconfig.yaml
@@ -1,0 +1,36 @@
+#TODO  long term, do we want to merge this with <repoHome>/deploy/base/config.yaml ??
+---
+apiVersion: jvmbuildservice.io/v1alpha1
+kind: UserConfig
+metadata:
+  name: jvm-build-config
+spec:
+  enableRebuilds: true
+  mavenBaseLocations:
+    maven-repository-300-jboss: "https://repository.jboss.org/nexus/content/groups/public/"
+    maven-repository-301-jitpack: "https://jitpack.io"
+    maven-repository-302-confluent: "https://packages.confluent.io/maven"
+    maven-repository-303-gradle: "https://repo.gradle.org/artifactory/libs-releases"
+    maven-repository-304-eclipselink: "https://download.eclipse.org/rt/eclipselink/maven.repo"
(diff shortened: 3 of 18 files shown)


deploy/kcp-hacbs-workspace-init/templates/userconfig.yaml
```diff
@@ -0,0 +1,36 @@
+#TODO  long term, do we want to merge this with <repoHome>/deploy/base/config.yaml ??
```
gabemontero, line 1: @stuartwdouglas @dwalluck I wanted to highlight this

we probably should reach a consensus on where the master/all possible options we want to expose lives

For example, in @dwalluck 's latest PR https://github.com/redhat-appstudio/jvm-build-service/pull/328/files#diff-323f42935c5af98a0b4152788343529680bef21786b46d6d23f8c23c6d785755R198 where he is updating the e2e's set up of the user config to add more kotlin locations, it may be time to work off of yaml vs. golang for stuff like that?

WDYT?

gabemontero: over time @stuartwdouglas exposing any of these fields as template parameters is what I was trying to describe during our sync up last night

deploy/init-hacbs-user-workspace-sa-rbac.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
gabemontero, line 1: based on the latest placement service cabal discussion today with @sbose78 and team, this compute cluster side rbac/sa initialization will most likely be some new api/contract, most likely exposed managed by the pipeline service, but that is not yet "official" as the design discussion that not reach the conclusion of Jira's getting opened, consensus being reached, etc.

deploy/kcp-hacbs-workspace-compute-cluster-rbac/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
```
gabemontero, line 1: based on the latest placement service cabal discussion today with @sbose78 and team, this compute cluster side rbac/sa initialization will most likely be some new api/contract, most likely exposed managed by the pipeline service, but that is not yet "official" as the design discussion that not reach the conclusion of Jira's getting opened, consensus being reached, etc.

deploy/kcp-hacbs-workspace-init/templates/userconfig.yaml
stuartwdouglas: A lot of these repos are really required for rebuilding dependencies, rather than something end users will want. I realised recently as this list has grown that a lot of these they should actually be a property of the build recipe repo rather than the users workspace. 

I am not actually sure how to do that yet though, in a lot of ways it would have been easier when we had the sidecar.

PR #330, opened 2022-10-18
PLNSRVCE-762: update pac infra-deployment update to use regex for commit update

gabemontero:
@Michkov this use of regex with sed updated all the sha instances correctly in my local copy of https://github.com/redhat-appstudio/infra-deployments/pull/827 and also correctly left the remaining content untouched.

The only concern I could come up with is whether the `sed` available to the PaC pipelinerun has the `-E` option.

So I took a look at https://github.com/redhat-appstudio/build-definitions/blob/main/tasks/buildah.yaml#L21, pulled that buildah image, booted up a container with it, an confirmed `sed` has the `-E` option.

PTAL

@stuartwdouglas FYI

### DIFF
--- a/.tekton/controller-push.yaml
+++ b/.tekton/controller-push.yaml
@@ -20,10 +20,7 @@ spec:
       value: Dockerfile
     - name: infra-deployment-update-script
       value: >
-        sed -i 
-        -e 's|\(https://github.com/redhat-appstudio/jvm-build-service/.*?ref=\)\(.*\)|\1{{ revision }}|'
-        -e 's/\(newTag: \).*/\1{{ revision }}/'
-        components/hacbs/jvm-build-service/base/kustomization.yaml
+        sed -i -E 's/[0-9a-f]{40}/{{ revision }}/g' components/hacbs/jvm-build-service/base/kustomization.yaml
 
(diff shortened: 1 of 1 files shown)


.tekton/controller-push.yaml
```diff
@@ -20,10 +20,7 @@ spec:
       value: Dockerfile
     - name: infra-deployment-update-script
       value: >
-        sed -i 
-        -e 's|\(https://github.com/redhat-appstudio/jvm-build-service/.*?ref=\)\(.*\)|\1{{ revision }}|'
-        -e 's/\(newTag: \).*/\1{{ revision }}/'
-        components/hacbs/jvm-build-service/base/kustomization.yaml
+        sed -i -E 's/[0-9a-f]{40}/{{ revision }}/g' components/hacbs/jvm-build-service/base/kustomization.yaml
```
gabemontero, line 23: well, I should clarify, I substituted a random string for `{{ revision }}` in my testing @Michkov

stuartwdouglas:
The only potential negative I can see with this approach is that it will need to be revisited if we every have multiple repositories at play in the infra-deployments repo (e.g. if we were also referencing builder-images there this would also pick up the builder image hashes).

Michkov (approved):

PR #331, opened 2022-10-19
Build with lowest JDK first

stuartwdouglas:
Some builds don't specify the bytecode level, so we can end up with a higher level if we rebuild with a newer JDK.

This can cause very hard to diagnose build issues.


gabemontero (approved):
/lgtm

PR #334, opened 2022-10-20
Fix cache transformation bug

stuartwdouglas:
If a sha was requested before the artifact the resulting transformed jar would end up corrupted.


gabemontero:
could this change effect `make e2etest` and that is why that failed, or was that failure a flake or related to something else @stuartwdouglas ?

stuartwdouglas:
It would not affect e2etest, the cache is not used there.

PR #336, opened 2022-10-20
Fix builder priority

stuartwdouglas:
I accidently put these in the wrong order


dwalluck:
> I accidently put these in the wrong order

I noticed and I think that this caused a regression in one of the apicurio builds

stuartwdouglas:
It did, this was how I noticed it as well

PR #338, opened 2022-10-21
Improve tag mapping

stuartwdouglas:


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/location/LookupScmLocationCommand.java
```diff
@@ -126,17 +127,19 @@ public void run() {
...
+                        if (versionExactContains.size() == 1) {
                             //only one contained the full version
-                            selectedTag = exactContains.iterator().next();
+                            selectedTag = versionExactContains.iterator().next();
                         } else {
-                            for (var i : exactContains) {
+                            for (var i : versionExactContains) {
                                 //look for a tag that ends with the version (i.e. no -rc1 or similar)
                                 if (i.endsWith(version)) {
                                     if (selectedTag == null) {
```
dwalluck, line 145: I am a little confused about the `for` loop. Can't we just check `endsWith` instead of `contains` back on line 129?

stuartwdouglas: You could, if you added some additional variables to track state, but I think this logic is easier to understand. This for loop will only ever be a handful of items, so it really does not matter either way.

PR #343, opened 2022-10-24
Go 1.18

stuartwdouglas:
@gabemontero everything else seems to require 1.18 so it seems like we might as well use it as well?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi8/go-toolset:1.17.7-13.1655148239 as builder
+FROM golang:1.18 as builder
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi8/go-toolset:1.17.7-13.1655148239 as builder
+FROM golang:1.18 as builder
```
stuartwdouglas, line 2: @gabemontero do you know if we have an official go 1.18 image anywhere? Most projects seem to use this one but dockerhub images can run into rate limiting issues.

stuartwdouglas:
Ok, lets just hold off for now then.

PR #352, opened 2022-10-28
Uninstall Helm charts if they already exist

stuartwdouglas:


gabemontero (approved):
cool idea @stuartwdouglas thanks 

that said, can you update the README and update the steps that call for entering the hacbs ws, create the test project, etc. and just mention the script will do it based on what HACBA_WORKSPACE_NAMESPACE is set to

PR #353, opened 2022-10-28
Ignore PipelineRuns synced from KCP

stuartwdouglas:
If the JVM build service is also running on the compute cluster (e.g. for non-KCP based testing) ignore any pipelines synched from KCP.


gabemontero (approved):
/lgtm

PR #354, opened 2022-10-28
Use correct secret format

stuartwdouglas:
This is not a long term thing, but lets us re-use this secret from the HACBS test pipeline.


deploy/kcp-hacbs-workspace-init/templates/secret.yaml
```diff
@@ -4,4 +4,11 @@ metadata:
   name: jvm-build-secrets
 stringData:
   registry.token: {{ .Values.quayToken }}
-  .dockerconfigjson: {{ .Values.quayToken }}
+  .dockerconfigjson: |
```
gabemontero, line 7: given the PR description that this is not a long term thing, maybe put a TODO comment about eventually removing this, and if possible detail when/what triggers the removal

thanks

PR #355, opened 2022-10-28
Add tmp chains fix

stuartwdouglas:


cmd/deletechainsfinalizer/main.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		panic(err)
+	}
+	namespaceList, err := clientset.CoreV1().Namespaces().List(context.TODO(), v1.ListOptions{})
+	if err != nil {
+		return
+	}
+	for _, namespace := range namespaceList.Items {
+		println("Namespace: " + namespace.Name)
+		if strings.Contains(namespace.Name, "jvm") || strings.Contains(namespace.Name, "kcp") {
```
gabemontero, line 48: so do you run this against the cluster kubeconfig, kcp kubeconfig, either, or both?

gabemontero (approved):
mystery solved :-)

cmd/deletechainsfinalizer/main.go
stuartwdouglas: Just against compute.

The first issue I ran into was fixed by updating the image, this needs to be fixed upstream.

PR #356, opened 2022-11-02
untested ingress placeholder for ingressed based access to the artfact cache

gabemontero:
this is a preliminary, very, very, very tiny step in the long journey of developing https://issues.redhat.com/browse/PLNSRVCE-764

so small that I'm not officially linking it to the story

fwiw I tried creating this from a hacbs enabled user workspace off of my kcp cps home workspace to see if kcp-glbc would post an accessible host in the ingress status,  but none of the objects from my hacbs enabled user workspace were getting synced to the kcp cps workload cluster; if I can enlist the SMEs in question to triage, sort out, and try again, I'll post status in https://issues.redhat.com/browse/PLNSRVCE-764

@stuartwdouglas @sbose78 @Mo3m3n FYI

### DIFF
--- a/deploy/kcp-hacbs-workspace-init/templates/service.yaml
+++ b/deploy/kcp-hacbs-workspace-init/templates/service.yaml
@@ -7,6 +7,7 @@ spec:
   - name: http
     port: 80
     targetPort: 8080
+    protocol: TCP
   selector:
     app: jvm-build-workspace-artifact-cache
   type: ClusterIP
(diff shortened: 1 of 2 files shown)


deploy/kcp-hacbs-workspace-init/templates/service.yaml
```diff
@@ -7,6 +7,7 @@ spec:
   - name: http
     port: 80
     targetPort: 8080
+    protocol: TCP
```
gabemontero, line 10: when I tried deploying this in a hacbs user workspace in cps stable it complained about the protocol not being set on the service

stuartwdouglas (approved):

stuartwdouglas:
Before we get this working we are going to need some kind of auth on the cache. It should be easy enough to use HTTP basic auth I think, however we would need to make sure we are using TLS as basic auth transmits the password in plaintext.

PR #357, opened 2022-11-02
Use git reset instead of git checkout

dwalluck:


dwalluck:
Fixes issue with checkout conflict for google/error-prone.

stuartwdouglas (approved):

stuartwdouglas:
Good idea

PR #382, opened 2022-11-28
switch from tektonwrappers to pending pipelineruns for throttling

gabemontero:
as part of investigating rate limiting / throttling in upstream tekton with an eye towards broadening the use of such concepts beyond jvm-build-service for stonesoup, I discovered that the "pending" bit for pipelineruns was in fact implemented (TEP-0015).

this PR is an initial pass at using the pending bit vs. our `TektonWrapper` ... if we move forward here, I'll make updates to remove that CRD, as well as rename some packages, etc.

the basic e2e passed for me locally (where I could see PRs start in pending state before our controller moved them out of pending state), but the initial apicurio test failed for me in the initial analyse-dependencies stage (which is unrelated to my changes); so I'm getting up a PR here to compare results and if need be engage @stuartwdouglas for debug

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -145,7 +145,7 @@ func NewManager(cfg *rest.Config, options ctrl.Options, kcp bool) (ctrl.Manager,
 		return nil, err
 	}
 
-	if err := tektonwrapper.SetupNewReconcilerWithManager(mgr, kcp); err != nil {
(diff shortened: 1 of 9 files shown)


stuartwdouglas (approved):

stuartwdouglas:
Apicurio issue was unrelated, it's passing now.

PR #396, opened 2022-12-05
Initial support for private repos

stuartwdouglas:
This adds support for both private metadata repos, and private forks of dependency repositories.

At present it is limited to a single provider (.e.g just GitHub).

### DIFF
--- a/pkg/reconciler/artifactbuild/artifactbuild.go
+++ b/pkg/reconciler/artifactbuild/artifactbuild.go
@@ -7,6 +7,8 @@ import (
 	"encoding/hex"
 	"fmt"
 	"github.com/redhat-appstudio/jvm-build-service/pkg/reconciler/systemconfig"
+	"k8s.io/apimachinery/pkg/labels"
+	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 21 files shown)


pkg/reconciler/artifactbuild/artifactbuild.go
```diff
@@ -580,6 +591,29 @@ func (r *ReconcileArtifactBuild) handleRebuild(ctx context.Context, abr *v1alpha
 	abr.Status.SCMInfo = v1alpha1.SCMInfo{}
 	abr.Status.Message = ""
 	err := r.client.Status().Update(ctx, abr)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	//now delete old pipelines
```
gabemontero, line 598: just for my education @stuartwdouglas - is deleting old pipelineruns actually part of private source code repo support, or is this just an optimization you added while you were in here?

If the latter, I think I see why that is OK, since we are setting our state back to new in line 590 as part of presumably a rebuild, right?

gabemontero (commented):
just one question @stuartwdouglas

PR #400, opened 2022-12-08
Remove disk usage health check

stuartwdouglas:
This is not really useful, and causes isssues with the builds.


gabemontero (approved):
/lgtm

PR #402, opened 2022-12-12
Add ability to clear the cache

stuartwdouglas:
There are circumstances where this might be nessesary.

### DIFF
--- a/docs/users.adoc
+++ b/docs/users.adoc
@@ -1,0 +1,29 @@
+= Users Guide
+
+WARNING: This is preliminary documentation
+
+
+== User Commands
+
+End users can interact with the JVM Build Service using annotations on various objects. These annotations are detailed below.
+
+=== ArtifactBuild Annotations
+
+`jvmbuildservice.io/rebuild`:: This annotation can be used to trigger a rebuild of an `ArtifactBuild`. It has two possible values, `true` and `failed`. If this is true a rebuild will always be triggered. For `failed` a rebuild is only triggered if the build was not successful.
+
+For example to rebuild all failed dependencies you can use:
+
+`kubectl annotate artifactbuilds.jvmbuildservice.io --all jvmbuildservice.io/rebuild=failed`
+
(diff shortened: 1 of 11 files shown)


docs/users.adoc
```diff
@@ -0,0 +1,29 @@
...
+`kubectl annotate artifactbuilds.jvmbuildservice.io --all jvmbuildservice.io/rebuild=failed`
+
+To rebuild a single specific dependency you can use:
+
+`kubectl annotate artifactbuilds.jvmbuildservice.io agroal.api.1.15-401ad867 jvmbuildservice.io/rebuild=true`
+
+
+=== JBSConfig Annotations
+
+`jvmbuildservice.io/clear-cache`::
```
gabemontero, line 25: don't have a strong opinion either way, but just curious, what lead you to using an annotation for this vs. a new field?  

fwiw, I think a first class field would force you down the path of spec/status for the jsbconfig

gabemontero (commented):
a question and a suggestion @stuartwdouglas

docs/users.adoc
stuartwdouglas: Annotation was what I used for ArtifactBuild rebuilds, and you said at the time that this was a relatively standard practice. I just assumed this would be the same.

gabemontero: yep sounds good / just making sure 

the possible distinction point, and if you want to move from annotations to spec / status API fields centers around the positive effect of noting the "status" or "results" of the request to clear the cache

i.e. you can keep use of the annotation and apply a label on completion per my suggestion below

or you add spec / status fields to the type, where the requester sets the bit in the spec, and then the controller / cache sets the bit in the status when the cache clear has been completed

certainly the annotation followed by the label is also suitable for something like this.  

but some form of "status notification" that the clear cache has been processed seems prudent / needed to me

hopefully I'm making my suggestion, and why I am making it, clearer

gabemontero (approved):
/lgtm

PR #403, opened 2022-12-12
Do final build in offline mode

stuartwdouglas:


stuartwdouglas:
This approach won't work, as the deploy step fails.

PR #406, opened 2022-12-15
Don't dump skipped pods

stuartwdouglas:


openshift-with-appstudio-test/e2e/util.go
```diff
@@ -507,6 +508,11 @@ func innerDumpPod(req *rest.Request, baseDirectory, localDirectory, podName, con
 	defer readCloser.Close()
 	var b []byte
 	b, err = io.ReadAll(readCloser)
+	if skipSkipped && len(b) < 1000 {
```
gabemontero, line 511: in theory, going through the pod's status, figuring out which containers have valid log contents based on their status, and then using https://github.com/redhat-appstudio/jvm-build-service/blob/059c41bd3bc42d7b67262f8b6149f6ccbedef24d/vendor/k8s.io/api/core/v1/types.go#L5394 and multilpe calls to only get the container logs you want would be more precise than the less than 1000 check

but certainly this is simpler

you decide which way you want to go

gabemontero (commented):
one suggestion though I'm fine with this simpler form

openshift-with-appstudio-test/e2e/util.go
stuartwdouglas: I looked at this, but from a kube pov the container exited successfully, it just successfully ran whatever that code is that prints the 'skipped' message.

gabemontero: ok cool let's stick with what you got then

gabemontero (approved):
/lgtm

PR #414, opened 2023-01-11
PLNSRVCE-991: disable/delete cache based on jbsconfig

Michkov:
Disable deployment (0 replicas) of cache when it's enableRebuilds is set to false.

Remove cache resources expect PVC when jbsconfig is removed


gabemontero:
the e2e could not get a cluster

/retest

now, as to the change, granted, I've been away a while, but I thought the cache was still used even if rebuilds is off; hence, I would not think this would work

of course the source of truth here is @stuartwdouglas but I know he is out until the 20th or something

but let's see what happens with an e2e that actually gets a cluster and runs the base line e2e

Michkov:
@gabemontero tests passed, but removed that replica changes based on enableRebuilds, now it's only deletion of cache with deletion of jbsconfig.

psturc:
👀  I'm going to verify this on my cluster...

@Michkov would it be worth (as a follow-up) to extend the current in-repo-e2e to test the scenario when the jbsconfig is removed?

pkg/reconciler/jbsconfig/jbsconfig.go
```diff
@@ -60,6 +60,42 @@ func (r *ReconcilerJBSConfig) Reconcile(ctx context.Context, request reconcile.R
 	jbsConfig := v1alpha1.JBSConfig{}
 	err := r.client.Get(ctx, request.NamespacedName, &jbsConfig)
 	if err != nil {
+		// Deleted JBSConfig - delete cache resources
+		if errors.IsNotFound(err) && request.Name == v1alpha1.JBSConfigName {
+			service := &corev1.Service{}
+			err = r.client.Get(ctx, types.NamespacedName{Name: v1alpha1.CacheDeploymentName, Namespace: request.Namespace}, service)
```
gabemontero, line 66: I would just delete straight away, and then check if you got a not found error, where you just move on if you did get a not found error, and but then log otherwise

I'd also create an warning event if we could not delete any of the cache related artifacts 

do this for all the items below

gabemontero (changes requested):
had to go back and look @Michkov , but remember now that the e2e runs with enableBuild == true ... see https://github.com/redhat-appstudio/jvm-build-service/blob/cb7025176f7ff7600417cbe725bb07f842c4b4c8/openshift-with-appstudio-test/e2e/util.go#L182

so the passing run at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/414/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1613272903766249472/artifacts/jvm-build-service-in-repo-e2e/jvm-build-service-e2e-in-repo-e2e/build-log.txt effective did not employ the new code that included the replica change

but all a moot point, as you removed the replica changes  

if we were ever to revisit that, we'd definitely need to add a verifying e2e where we set rebuilds to false in the same PR before merging

As to the @psturc suggestion of adding an e2e for the current scenario, yes, a very good idea IMO.  I'm fine with whatever staging you guys agree to (this PR, follow up PR), given the "safer" nature of this change.

Of course, murphy's law, when @stuartwdouglas comes back he'll tell us we missed something obvious ;-), but this should be easy enough to revert.

Lastly, one coding suggestion below.

thanks

pkg/reconciler/jbsconfig/jbsconfig.go
Michkov: added the warning events on objects which having issue with deletion.

Michkov:
Created PLNSRVCE-998 for e2e-test coverage.

pkg/reconciler/jbsconfig/jbsconfig.go
gabemontero: yes sorry that was unclear .... I mean on the object being deleted

I think the other part of my comment also was not clear (just delete vs. get/delete).  Let my try a code snippet to clear up:

```
service := &corev1.Service{}
service.Name = v1alpha1.CacheDeploymentName
service.Namespace = request.Namespace
err = r.client.Delete(ctx, service)
if err != nil && !errors.IsNotFound(err) {
   // log error and record event
}
// otherwise just continue
```

IIRC there are no resource version checks or anything like that on the delete.  I just works off of the name/namespace.

Michkov: thanks, updated.

gabemontero (approved):
/lgtm

PR #417, opened 2023-01-13
Remove ServiceMonitor since monitoring is set externally

Michkov:
Based on [PLNSRVCE-945](https://issues.redhat.com//browse/PLNSRVCE-945) there should be separate namespace aggregating ServiceMonitors.

Connected in https://github.com/redhat-appstudio/infra-deployments/pull/1137


gabemontero (commented):
hmmm just occurred to me (where as it did not occur to me during the review of https://github.com/redhat-appstudio/infra-deployments/pull/1137 ) that @stuartwdouglas as much as possible want jvm-build-service to be independent of infra-deployments for use outside of the full stonesoup

this change of course makes the functioning of metrics dependent on getting deployed from infra-deployments.

maybe that is OK for him, but I'd like to get his feedback

off the top of my head, I see 2 options
1) add the creation of the servivcemonitor and requisite rbac as part of the kustomization developer flow instead of having it off of deploy/operator/base as part of this PR
2) just open a tracking Jira to circle back to this when @stuartwdouglas comes back, but merge this in the interim, where maybe that develop kustomization flow is done as part of that jira, or he decides he does not care about metrics when running outside of stonesoup.

I'm fine either way @Michkov .... just respond with which way you want to go

stuartwdouglas:
I think this approach is better, so it still works on my cluster: https://github.com/redhat-appstudio/jvm-build-service/pull/428

PR #420, opened 2023-01-16
fix: give priority to image name from env var

psturc:
Recently I stumbled upon an issue with rebuilding dependencies. the issue was caused by a wrong image name specified for a reqprocessor, that was generated by a controller.

I expected the image name from env vars (`JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE`, `JVM_BUILD_SERVICE_CACHE_IMAGE`) should have a priority, i.e. first check if these env vars exist and if not, then generate the image names based on the controller image name etc. But currently it's not the case.

This update should solve it. 

If it makes sense, I will add some unit test.

### DIFF
--- a/pkg/reconciler/util/util.go
+++ b/pkg/reconciler/util/util.go
@@ -42,6 +42,14 @@ func GetImageName(ctx context.Context, client client.Client, log logr.Logger, su
 			return "", err
 		}
 	}
+
+	// Get the image name using a controller's env var (if the env var value is specified)
+	ciImageName := os.Getenv(envvar)
+	if len(ciImageName) != 0 {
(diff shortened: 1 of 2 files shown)


psturc:
@gabemontero @Michkov would you have a time to give this one a look? If it makes sense I will add some unit test later.

gabemontero (commented):
Yeah go for it, add the unit tests, etc. @psturc 

I tried to revisit all the different paths here, can't think of any issues.  Also can't recall why I coded it the way I did before. The order of precedence you noted makes sense.

gabemontero:
no clusters available

/retest

gabemontero (approved):
/lgtm

PR #433, opened 2023-01-25
Remove legacy KCP code

stuartwdouglas:
We were still using the KCP controller-runtime fork. This reverts back to standard controller-runtime, and removes all legacy KCP code.


gabemontero (approved):
/lgtm

PR #471, opened 2023-02-10
Use docker build from the catalog

stuartwdouglas:


stuartwdouglas:
@Michkov I was hoping to utilise the snyk scan to help resolve https://issues.redhat.com/browse/STONEBLD-432 but it does not actually run due to a missing secret. Do we have any way to actually run this at the moment?

stuartwdouglas:
Actually I have just enabled dependabot, which is probably better than snyk anyway.

Michkov:
@stuartwdouglas this bundle does not have infra-deployment-update task, we do not have new bundles yet.

stuartwdouglas:
closing, as this does not help with 432 anyway.

PR #497, opened 2023-02-16
Use 'default' repo list rather than 'all'

stuartwdouglas:


stuartwdouglas:
@dwalluck this will cause failures, but it should reduce the load on the cache. We need to add repos to the ones that fail before merging this.

rnc:
Why was this closed please?

stuartwdouglas:
I accidentally committed it here: https://github.com/redhat-appstudio/jvm-build-service/pull/706/commits/90eb3d6859b9412a455c1618d7ca66015ec86396#diff-b98df68065a725fb4509f65dc0c24c3237a77d76a1bfe8f3c0df236f770f1fe0R91

Basically I started out with the 'additional repos specified' case as also using all, then changed it to default but did not notice I had also extracted the lower value as well.

I don't think this will cause any problems, maven builds should basically automatically pick up the correct repositories.

PR #506, opened 2023-02-17
Update Ant Ivy default resolver

dwalluck:
It was using the cache URL directly. Now it uses a chain which has the m2 local repo and the cache.


stuartwdouglas (approved):

stuartwdouglas (changes requested):
The failure looks related

dwalluck:
> The failure looks related

Fixed. The code was OK, but test wasn't updated for the new code.

In the previous code, the default Maven repo had the same name as the chain (which includes the Maven repo and the maven local repo), so it wasn't using the chain.

stuartwdouglas (approved):

PR #507, opened 2023-02-19
ScmLocator: allow plugging in a fallback ScmLocator

aloubyansky:
- Turned `ScmLocator` into an interface
- Moved the previous impl of the `ScmLocator` to `GitScmLocator` with a possibility to configure a fallback `ScmLocator`
- extracted the POM-based SCM locator fallback into `PomScmLocator` that also implements `ScmLocator` and set it as the default fallback in `LookupScmLocationCommand`

FYI @stuartwdouglas


aloubyansky:
@stuartwdouglas in case you merge this one, would you mind making another release? Thanks!

stuartwdouglas:
Released

PR #515, opened 2023-03-01
STONEBLD-847 Use TLS to access the Cache

stuartwdouglas:
Early PR to test this out and get early feedback.

### DIFF
--- a/hack/examples/pipeline-gradle.yaml
+++ b/hack/examples/pipeline-gradle.yaml
@@ -53,7 +53,7 @@ spec:
         - fetch-repository
       params:
         - name: MAVEN_MIRROR_URL
-          value: http://jvm-build-workspace-artifact-cache.$(context.taskRun.namespace).svc.cluster.local
+          value: https://jvm-build-workspace-artifact-cache-tls.$(context.taskRun.namespace).svc.cluster.local
         - name: CONTEXT_DIR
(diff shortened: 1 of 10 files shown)


stuartwdouglas:
@gabemontero do you have any thoughts on this? Basically I am using the 'service.beta.openshift.io/serving-cert-secret-name' annotation to generate keys/certificates, that I can then use to serve HTTPS from the Cache.

Unfortunately consuming it is a mess, I need to add the CA to the default Java truststore, which I am doing by hacking in some code to to it at the start of a task step. It would be ideal if I can bake this into the image instead but I am not sure if that is possible, as it sounds like it will be different for different deployments.

stuartwdouglas:
So the main issue I have is that I can't control the client code, so custom trust managers that understand X509 certs are probably out.

Because I have control of the JDK installation its possible I could install a custom SecurityProvider that would be able to handle this, but that is a big 'maybe' and I don't really have time to investigate this approach at the moment. I know we have existing code that can handle turning a PEM file into a Java Keystore in memory, but I am not really sure how I would go about making this active by default. 

I think for now I might just stay with the current approach, and investigate how to fix it properly later.

stuartwdouglas:
This is blocked on a builder images update to fix an issue with the JDK7 image.

PR #516, opened 2023-03-01
STONEBLD-892 remove pending pipelinerun related code; leverage k8s resource quota in e2e; sandbox quotas in stonesoup

gabemontero:
hey @stuartwdouglas 

my official follow up from [PLNSRVCE-928](https://issues.redhat.com//browse/PLNSRVCE-928) and its findings that we discussed back in January, where the pending pipelinerun stuff is not providing any real value if k8s resource quotas are in place

tl;dr at https://docs.google.com/presentation/d/1-59oAK3fdlWlrvFXTnheRLFtSLh1fOYiE-7BEh4-C1I/edit#slide=id.p

So I've removed the functionality in the PR to free up cycles/mem in the controller

the service registry / periodic should sufficiently throttle with the policy in place


stuartwdouglas:
One of the Apicurio failures looks like a PipelineRun failed to launch any pods. I don't think this is unique to this PR (we have definitely seen it before), but I do wonder if moving to a pod based limit will mean that sometimes pipelines will fail if there are too many concurrent runs. 

I have opened https://github.com/redhat-appstudio/jvm-build-service/pull/517 to get the PipelineRun yaml as well as the log files to try and get more information about what has happened. Once this is merged I will rebase this one and see if it happens again.

gabemontero:
> One of the Apicurio failures looks like a PipelineRun failed to launch any pods. I don't think this is unique to this PR (we have definitely seen it before), but I do wonder if moving to a pod based limit will mean that sometimes pipelines will fail if there are too many concurrent runs.
> 
> I have opened #517 to get the PipelineRun yaml as well as the log files to try and get more information about what has happened. Once this is merged I will rebase this one and see if it happens again.

yeah during my https://issues.redhat.com//browse/PLNSRVCE-928 testing @stuartwdouglas I never saw k8s fail a pod launch because of quota (i.e. a timeout)

now, I could see launched containers get cpu/mem constrained and lead to resource constraint flakes

but yeah, let's wait on #517 and re-analyze with that change in to confirm one way or the other

gabemontero:
wow @stuartwdouglas @dwalluck since I last paid any attention the dependency builds are getting really close now for service registry :-) 

`    types.go:24: time: 2023-03-03 05:37:22.778199327 +0000 UTC m=+2736.327186512: dependencybuild FINAL created count: 130 complete count: 126, failed count: 4, contaminated count: 0`

taking a look at the failures now

gabemontero:
Yeah, moving forward, if there is a pipeline run failed to launch a pod, I would be curious about cross referencing any pods whose name prefix correlates with the pipelinerun name in the events.json.  

A resource constraint around pod launch should be captured in the events.

gabemontero:
Lastly,  if we ever see cpu/mem related resource constraints, even with the 50 pod count limit, instead of reverting back to employing pending pipeline runs, we'll want to replicate more of the sandbox `ResourceQuota` and `LimitRanges` in the periodic e2e

Where we pick somewhere in between what I had from https://github.com/redhat-appstudio/jvm-build-service/pull/422 which were the original sandbox defaults, to https://github.com/redhat-appstudio/jvm-build-service/pull/421, which was optimized for fastest completion.

dwalluck:
> 3\. db github.com/charleskorn/kaml gradle db prefix 786fbf69135af41fb01c89ab924c5551786fbf69135af41fb01c89ab924c5551

This one looks like a recent regression, missing from net.ltgt.gradle:gradle-apt-plugin:0.21 may be a repo issue.

stuartwdouglas (approved):

stuartwdouglas:
Those are all known failures that we are working on. This is looking good so lets get it in.

PR #530, opened 2023-03-08
STONEBLD-910 Openshift Pipelines 1.9

stuartwdouglas:
DO NOT MERGE YET

This is for testing against 1.9 in CI, so we know we are ready.


deploy/openshift-pipelines-subscription.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: openshift-pipelines-operator
   namespace: openshift-operators
 spec:
-  channel: pipelines-1.8
+  channel: pipelines-1.9
```
Michkov, line 7: update also go.mod -> https://github.com/redhat-appstudio/jvm-build-service/blob/main/go.mod#L13, `v0.41.1` matches to Openshift Pipelines 1.9

stuartwdouglas: I did not want to do that for testing, as I was testing if the existing code could work against 1.9 so the infra-deployments PR could merge. Updating go.mod in this PR would mean that the older version running on infra-deployments would be untested.

stuartwdouglas: @Michkov do we have an official base image for go 1.19? The upgrade brings in dependencies that require it, and there does not seem to be a UBI based image.

Michkov: seems that newest image contains 1.18 - https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690

stuartwdouglas: It turns out that even though the build output says 1.19 is required 1.18 seems to work.

PR #534, opened 2023-03-09
Do not exit due to microdnf return code

dwalluck:
Even `exit 0` was bad as it was stopping the build.


stuartwdouglas (approved):

stuartwdouglas:
This script is appended into a single big script, rather than being run directly, which is why exit would stop it.

PR #539, opened 2023-03-13
Revert "STONEBLD-693 Ability to test Java Artifacts"

stuartwdouglas:
Reverts redhat-appstudio/jvm-build-service#511


stuartwdouglas:
@dwalluck turns out this was not the issue.

PR #540, opened 2023-03-14
STONEBLD-959 Experiments with offline builds

stuartwdouglas:


pkg/reconciler/dependencybuild/scripts/maven-build.sh
```diff
@@ -27,7 +27,11 @@ EOF
 cp -r $(workspaces.source.path)/workspace $(workspaces.source.path)/source
 #we can't use array parameters directly here
 #we pass them in as goals
-mvn -B -e -s "$(workspaces.build-settings.path)/settings.xml" $@ "-DaltDeploymentRepository=local::file:$(workspaces.source.path)/artifacts" "org.apache.maven.plugins:maven-deploy-plugin:3.0.0-M2:deploy" | tee $(workspaces.source.path)/logs/maven.log
+mvn -B -e -s "$(workspaces.build-settings.path)/settings.xml" $@ "-DaltDeploymentRepository=local::file:$(workspaces.source.path)/artifacts" "org.apache.maven.plugins:maven-deploy-plugin:3.0.0-M2:deploy" | tee $(workspaces.source.path)/logs/maven-original.log
+
+rm -r $(workspaces.source.path)/artifacts
+
+unshare -n -Ufp -r -- mvn -B -e -s "$(workspaces.build-settings.path)/settings.xml" $@ "-DaltDeploymentRepository=local::file:$(workspaces.source.path)/artifacts" "org.apache.maven.plugins:maven-deploy-plugin:3.0.0-M2:deploy" | tee $(workspaces.source.path)/logs/maven.log
```
stuartwdouglas, line 34: @ralphbean @sbose78 @Michkov What do you guys think about this approach to Hermetic Java builds?

Basically we run the build normally, then delete the result and then re-run via 'unshare -n' which should prohibit network access. These results from the offline build are then used for deployment.

I am also wondering if we should be archiving all the dependencies used in the build, although creating an image with them for every build is not very space efficient.

java-components/build-request-processor/src/main/resources/gradle/repositories.gradle
```diff
@@ -3,17 +3,20 @@ apply plugin: RepositoryPlugin
...
         def fixRepositories = {
             def bintray = false
             all { ArtifactRepository repo ->
                 if (!(repo instanceof MavenArtifactRepository) ||
-                        (repo.url.toString() != ENTERPRISE_REPOSITORY_URL && repo.url.toString() != BINTRAY_URL )) {
+                        (repo.url.toString() != ENTERPRISE_REPOSITORY_URL)) {
                     if (repo.toString().contains("PluginArtifactRepository") || (Boolean.getBoolean("ALLOW_IVY_REPO") && repo.toString().contains("DefaultIvyArtifactRepository"))) {
                         return
                     }
+                    if (repo instanceof IvyArtifactRepository) {
```
ralphbean, line 16: What's up with this change? Is it related?

pkg/reconciler/dependencybuild/scripts/gradle-build.sh
```diff
@@ -82,6 +82,11 @@ cp -r $(workspaces.source.path)/workspace $(workspaces.source.path)/source
 
 gradle $INIT_SCRIPTS -DAProxDeployUrl=file:$(workspaces.source.path)/artifacts --info --stacktrace -Prelease.version=$(params.ENFORCE_VERSION) "$@"  | tee $(workspaces.source.path)/logs/gradle.log
```
ralphbean, line 83: Adjust this to output to logs/gradle-online.log just like the ant build script above?

pkg/reconciler/dependencybuild/scripts/maven-build.sh
ralphbean: Do you have any data yet to suggest that this increases the failure rate of dependency rebuilds significantly?

stuartwdouglas: It does a bit, but they are mostly ant builds that are likely doing something they shouldn't anyway: 

Build with this enabled:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/540/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-apicurio-e2e/1637756596568199168/artifacts/jvm-build-service-in-repo-apicurio-e2e/jvm-build-service-in-repo-apicurio-e2e/artifacts/jvm-build-service-report/index.html

Without:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/480/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-apicurio-e2e/1636172251504054272/artifacts/jvm-build-service-in-repo-apicurio-e2e/jvm-build-service-in-repo-apicurio-e2e/artifacts/jvm-build-service-report/index.html

Most of the failures seem to be related to a bug with the ant script that I will fix today and re-run. 

Micrometer seems to be the only genuine failure, it seems to still be attempting to hit the cache. Note that I can't use maven offline mode as the deploy plugin will fail in offline mode (even for local file deployments).

java-components/build-request-processor/src/main/resources/gradle/repositories.gradle
stuartwdouglas: Oops, this was accidentally commited.

pkg/reconciler/dependencybuild/scripts/maven-build.sh
```diff
@@ -27,7 +27,12 @@ EOF
 cp -r $(workspaces.source.path)/workspace $(workspaces.source.path)/source
 #we can't use array parameters directly here
 #we pass them in as goals
-mvn -B -e -s "$(workspaces.build-settings.path)/settings.xml" $@ "-DaltDeploymentRepository=local::file:$(workspaces.source.path)/artifacts" "org.apache.maven.plugins:maven-deploy-plugin:3.0.0-M2:deploy" | tee $(workspaces.source.path)/logs/maven.log
+mvn -B -e -s "$(workspaces.build-settings.path)/settings.xml" $@ "-DaltDeploymentRepository=local::file:$(workspaces.source.path)/artifacts" "org.apache.maven.plugins:maven-deploy-plugin:3.0.0-M2:deploy" | tee $(workspaces.source.path)/logs/maven-online.log
+
+rm -r $(workspaces.source.path)/artifacts
+microdnf install iproute
```
stuartwdouglas, line 33: We will not need this in the final PR, we would add this to the images.

stuartwdouglas:
@ralphbean something that I am worried about in general is archiving of the build dependencies. If we are pulling from a lesser known maven repo and that repo disappeared (e.g. what happened with bintray), we don't have a strategy of making sure they are archived forever. 

It's not directly related to this PR, but it does kinda tie in because it means that even though the build is offline we don't necessarily have everything used by the build archived. 

We could create a 'build container' that has all the dependencies, and have this container run the actual build step, but it would not be super efficient in terms of storage, you would have lots of artifacts duplicated across images.

pkg/reconciler/dependencybuild/scripts/maven-build.sh
ralphbean: Cool. 1 out of ~200 is not too bad. And, what does the path look like for a user who encounters a dependency like micrometer that genuinely fails? Is the only path to engage with upstream and agree on patches to build without going back out to cache? or, do you suspect there is some path via `jvm-build-data/build-deps/`?

In any case, I think I'm +1 to merge this and we can reap feedback from users. The increased confidence in provenance is worth the 1 failure in 200 tradeoff, in my opinion.

ralphbean:
> @ralphbean something that I am worried about in general is archiving of the build dependencies. If we are pulling from a lesser known maven repo and that repo disappeared (e.g. what happened with bintray), we don't have a strategy of making sure they are archived forever.

That makes sense. We have a feature tracker for general source archival at https://issues.redhat.com/browse/STONE-115. We should add jvm-build-service use cases to its requirements. It doesn't currently have a blocker or critical priority (cc @brianwcook) so it is unlikely to get attention from @sbose78 before May.

stuartwdouglas:
I am closing this for now, it is not the correct way to do offline builds.

PR #548, opened 2023-03-16
GitScmLocator option to open local recipe repos w/o cloning them

aloubyansky:
FYI @stuartwdouglas 

I was not sure whether it'd be a good idea to change the default, so I didn't.


java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/scm/GitScmLocator.java
```diff
@@ -136,7 +176,7 @@ public TagInfo resolveTagInfo(GAV toBuild) {
         //look for SCM info
         var recipes = recipeGroupManager
                 .requestArtifactInformation(
-                        new ArtifactInfoRequest(Set.of(toBuild), Set.of(BuildRecipe.SCM, BuildRecipe.BUILD)))
```
aloubyansky, line 139: It looked like it was OK to limit it to the SCM here.

stuartwdouglas (approved):

PR #561, opened 2023-03-23
Only add .withAllowInsecureProtocol for SBT after 1.3.0

rnc:


pkg/reconciler/dependencybuild/scripts/sbt-build.sh
```diff
@@ -31,10 +31,15 @@ cat > "$HOME/.sbt/repositories" <<EOF
   my-maven-proxy-releases: $(params.CACHE_URL), allowInsecureProtocol
 EOF
 
-mkdir "$HOME/.sbt/1.0/"
-cat >"$HOME/.sbt/1.0/global.sbt" <<EOF
+# withAllowInsecureProtocol is only after 1.3.0
+# https://github.com/sbt/librarymanagement/pull/318
+if [ $(sbt sbtVersion | tail -1 | sed 's/.* //' | tr -d '.') -ge 130 ]; then
+    mkdir "$HOME/.sbt/1.0/"
```
stuartwdouglas, line 37: It turn out that this fails in some cases, because the 'sbt sbtVersion' command creates it. I created https://github.com/redhat-appstudio/jvm-build-service/pull/565 to fix the issue.

PR #572, opened 2023-03-27
Add more logging. Add ability to use a specific branch for the recipe…

rnc:
… database

### DIFF
--- a/java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/scm/GitScmLocator.java
+++ b/java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/scm/GitScmLocator.java
@@ -74,7 +74,7 @@ public class GitScmLocator implements ScmLocator {
         }
 
         /**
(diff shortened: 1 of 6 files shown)


java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/scm/GitScmLocator.java
```diff
@@ -141,7 +141,14 @@ private RecipeGroupManager getRecipeGroupManager() {
                 if (remotePattern == null || remotePattern.matcher(i).matches()) {
                     log.infof("Cloning recipe repo %s", i);
                     try {
-                        repoManager = RecipeRepositoryManager.create(i, "main", Optional.empty(),
+                        // Allow cloning of another branch via <url>#<branch> format.
+                        String branch = "main";
```
stuartwdouglas, line 145: This is used in other places as well. Lookup all uses of the `RecipeRepositoryManager.create` function.

It would probably be better if this was pushed down into `RecipeRepositoryManager`

stuartwdouglas (changes requested):

PR #576, opened 2023-03-29
STONEBLD-809 Add JvmBuildStatus object

stuartwdouglas:
This object tracks the results of an upstream build. It will eventually allow for PRs to be re-run when all dependencies are built, and provide a way to track the status of any given build.

It will also be used to provide source code downloads of the rebuilt artifacts.


deploy/crds/base/jvmbuildservice.io_jvmbuildstatus.yaml
```diff
@@ -0,0 +1,108 @@
...
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.6.2
+  creationTimestamp: null
+  name: jvmbuildstatus.jvmbuildservice.io
+spec:
+  group: jvmbuildservice.io
+  names:
+    kind: JvmBuildStatus
```
stuartwdouglas, line 13: Not sure about this name.

stuartwdouglas:
@sbose78 FYI this adds a new object that tracks which ArtifactBuild's are needed for each components PipelineRun.

This should eventually allow us to re-trigger builds/pull requests when their dependencies are built, and provide an easy way for the UI to see where needed dependencies are up to. 

One outstanding issue is that there is no obvious way to manage the lifecycle of these objects. The PipelineRun they are associated with will be pruned very quickly, but we will want these objects to stay for longer.

deploy/base/maven-v0.2.yaml
```diff
@@ -84,6 +84,8 @@ spec:
       description: The total number of dependencies that came from an untrusted source, such as Maven central
     - name: JAVA_COMMUNITY_DEPENDENCIES
       description: The list of community dependencies
+    - name: JAVA_DEPENDENCIES
```
stuartwdouglas, line 87: I am not sure about this approach, as it is an extra very long results field. An alternative would be to allow the analyser to create it directly, although that would require making sure the pipeline account had the correct permissions.

This was actually how things worked in the pre-kcp code.

PR #577, opened 2023-03-29
Remove colors and correct cross-build command

rnc:
According to https://www.scala-sbt.org/1.x/docs/Cross-Build.html and https://www.scala-sbt.org/1.x/docs/Cross-Build-Plugins.html there shouldn't be a space between `+` and `publish`.
Also removed colour to avoid in the logs:
```
[0m[[0m[31merror[0m] [0m[0mNot a valid command: +[0m
```
In SBT 1.2.8 it is `-no-color` but that was changed after 1.3.0 to `--no-colors` - and currently SBT 1.8.0 is invoked as the toolVersion which has `--no-colors`

### DIFF
--- a/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommand.java
+++ b/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommand.java
@@ -239,14 +239,11 @@ public class LookupBuildInfoCommand implements Runnable {
             if (Files.exists(path.resolve("build.sbt"))) {
                 //TODO: initial SBT support, needs more work
                 Log.infof("Detected SBT build in %s", path);
-
-                var specifiedJavaVersion = "11"; //hard coded for now
-
                 info.tools.put(JDK, new VersionRange("7", "17", "8"));
                 info.tools.put(SBT, new VersionRange("1.8.0", "1.8.0", "1.8.0"));
                 info.toolVersion = "1.8.0";
                 info.invocations.add(new ArrayList<>(
-                        List.of(SBT, "+", "publish"))); //the plus tells it to deploy for every scala version
+                        List.of(SBT, "--no-colors", "+publish"))); //the plus tells it to deploy for every scala version
             }
             if (AntUtils.isAntBuild(path)) {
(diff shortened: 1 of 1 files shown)


stuartwdouglas:
In that doc on the command line (which this is) there is a space, in the build file there is not. Are you sure this is correct? 

We really need some SBT based tests, @dwalluck any chance you could create a SBT test project like you did for gradle?

rnc:
> In that doc on the command line (which this is) there is a space, in the build file there is not. Are you sure this is correct?
> 
> We really need some SBT based tests, @dwalluck any chance you could create a SBT test project like you did for gradle?

I agree its badly documented in SBT. I see when building scala-collection-compat I get

```
[0m[[0m[31merror[0m] [0m[0mNot a valid command: +[0m
[0m[[0m[31merror[0m] [0m[0mExpected ID character[0m
[0m[[0m[31merror[0m] [0m[0mExpected whitespace character[0m
[0m[[0m[31merror[0m] [0m[0mExpected '++'[0m
[0m[[0m[31merror[0m] [0m[0mExpected '+-'[0m
[0m[[0m[31merror[0m] [0m[0m+[0m
[0m[[0m[31merror[0m] [0m[0m ^[0m
```
implying its not handling the `+ publish` - this is with SBT 1.2.8 - while i think both of your current test SBT builds use SBT after 1.3.0. So far I haven't managed to find references inside sbt to whether they changed handling of + after 1.3.0.

stuartwdouglas:
Lets make sure it doesn't break the apicurio builds then, otherwise we will need to adjust the handling based on version.

rnc:
> Lets make sure it doesn't break the apicurio builds then, otherwise we will need to adjust the handling based on version.

Sure, sounds good. I think https://www.scala-sbt.org/1.x/docs/Combined+Pages.html#Pass+arguments+to+a+command+or+task+in+batch+mode elaborates (slightly!) on passing arguments through to SBT while https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html mentions them in context of the interactive console. So potentially we could either quote or try removing the space.

stuartwdouglas:
What we really need is the ability to do faster testing around all this. I am looking at trying to do minikube+gh actions testing which should be able to give much faster and more focused results, but it is not simple.

PR #581, opened 2023-03-30
STONEBLD-1098 replace missing metrics

stuartwdouglas:


stuartwdouglas:
@gabemontero is there any 'standard' way of checking in the integration test that the metrics are actually working? The unit test was not enough in this case as the registration code was accidentally deleted, and nothing actually checked the metrics at runtime.

gabemontero:
> @gabemontero is there any 'standard' way of checking in the integration test that the metrics are actually working? The unit test was not enough in this case as the registration code was accidentally deleted, and nothing actually checked the metrics at runtime.

Not super easily, but not "impossible"

The best "standard" way I can give you is curling the metrics service endpoint, a la https://docs.openshift.com/container-platform/4.12/monitoring/managing-metrics.html

But that involves creating a test pod in the test cluster that does the curl using the service's intra-cluster  host name and the metrics service's port ... i.e. similar how the artifact cache is accessed from the the db/ab pipielineruns

For openshift metrics, we had things like the instructions at https://docs.openshift.com/container-platform/4.12/monitoring/managing-metrics.html#viewing-a-list-of-available-metrics_managing-metrics but the jvm build service metrics are not wired to be collected by openshift, so I don't think that is an option

there are also developer metrics for presumably adding non OCP metrics  ... I really am unfamiliar with those

I think the test pod curling the metrics port would be your best option, if you want to go down this path.

Or you could argue this is a path that won't change much, and you can afford the one or two blips that occur in order to make sure it works.

stuartwdouglas:
I think I might just leave it then. If there was a simple way to do it then I would look into it but it sounds like it is not worth the effort.

PR #592, opened 2023-04-05
Fix Spring repo and remove broken kotlinnative14linux

rnc:


java-components/cache/src/main/resources/application.properties
```diff
@@ -41,12 +41,10 @@ store.dokkadev.url=https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev
 store.dokkadev.type=maven2
 store.ajoberstar.url=https://ajoberstar.org/bintray-backup
 store.ajoberstar.type=maven2
-store.spring-plugins.url=https://repo.springsource.org/plugins-release
```
stuartwdouglas, line 44: So these are kinda legacy, but are still useful when running locally so I have not removed them.

For rebuilds repositories are now stored in the git repo. For user code it should be defined in the JBS config which is passed in as env variables. These values should have no effect when running on the cluster, and will only take effect when running locally and testing an initial build.

rnc: But that repo doesn't actually exist? And as per the pastes from the chat I am currently getting errors...

stuartwdouglas: Sorry, I was just trying to clarify what these are for. They probably should have %dev. added in from of them all to make it clearer. 

Did you see the error in the cluster or running it locally?

rnc: I saw the problem in the cluster. But I'm afraid I still don't follow in that how can these be used whether in dev mode or not if the repository physically doesn't exist?

stuartwdouglas: Well it will not get anything from them in dev mode, but it should not cause an error (it would just be ignored, and slow things down slightly). What I am saying is just that this is mostly just legacy cruft, so if it is affecting things on the cluster somewhere that is a problem.

spring-plugins for rebuilds should come from: https://github.com/redhat-appstudio/jvm-build-data/blob/875cbf84acdabdaf5fbcc3f2951e7b92bce9a341/repository-info/spring-plugins.yaml#L1

stuartwdouglas: https://github.com/apheleia-project/apheleia/blob/e04b9ff784e4b9cb20f0a754d364d092344dbde1/deploy/user-namespace/jbs-config.yaml#L19

This would explain it, its in the aphelia config. I will merge this but for the most part we should just delete all these. originally there was no differentiation between user level repos and rebuild repos, so we added them all to the cache config for easy debugging in dev mode.

rnc: Ok. I had put in https://github.com/apheleia-project/apheleia/pull/17 but wasn't sure how these configuration files were used given the information in jvm-build-data.

PR #604, opened 2023-04-21
STONEBLD-1211 Add finalizers to all pipelineruns

stuartwdouglas:


pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -756,7 +759,7 @@ func (r *ReconcileDependencyBuild) handleBuildPipelineRunReceived(ctx context.Co
 		}
 
 	}
-	return reconcile.Result{}, nil
+	return artifactbuild.RemovePipelineFinalizer(ctx, pr, r.client)
```
gabemontero, line 762: This has always bugged me about github PR reviews, that it only allows comments around changed lines :-) 

But if you go up 12 lines @stuartwdouglas to line 750 here, where you `return r.handleStateCompleted(ctx, &db, log)` 

Looking at that method, I don't see any removing of the finalizer there.

So it looks like that might be a path where you leak the finalizer, if you don't get any more PipelineRun Events after whichever one lead you down that path.

Of course, it has been a while since I've looked at this code, so I might be missing something

gabemontero: I believe you want to follow the pattern you used in artifactbuild.go:

```
	err = r.client.Status().Update(ctx, &abr)
	if err != nil {
	}
	result, err2 := RemovePipelineFinalizer(ctx, pr, r.client)
	if err2 != nil {
		return result, err2
	}
```

where you updated the ab, returned if you got an error, but otherwise went on the remove the finalizer

gabemontero (commented):
I believe I found a finalizer leak @stuartwdouglas  ... take at look at my attempt to explain why, as well as cite what you should do, following the pattern you already employed elsewhere

pkg/reconciler/dependencybuild/dependencybuild.go
stuartwdouglas: Good catch, I missed that return statement.

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -223,7 +224,7 @@ func (r *ReconcileDependencyBuild) handleStateAnalyzeBuild(ctx context.Context,
 		r.eventRecorder.Eventf(pr, v1.EventTypeWarning, "MissingOwner", msg, pr.Namespace, pr.Name)
 		log.Info(msg, pr.Namespace, pr.Name)
 
-		return reconcile.Result{}, nil
+		return artifactbuild.RemovePipelineFinalizer(ctx, pr, r.client)
```
gabemontero, line 227: did I find another one ... down below at line 239-240:

```
	if db.Status.State != v1alpha1.DependencyBuildStateAnalyzeBuild {
		return reconcile.Result{}, nil
	}
```

pkg/reconciler/artifactbuild/artifactbuild.go
```diff
@@ -257,10 +258,38 @@ func (r *ReconcileArtifactBuild) handlePipelineRunReceived(ctx context.Context,
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+	result, err2 := RemovePipelineFinalizer(ctx, pr, r.client)
```
gabemontero, line 261: also, in the dependencybuild.go versions of this, if the ownerfefs were missing, you removed the finalizer:
```
	ownerRefs := pr.GetOwnerReferences()
	if len(ownerRefs) == 0 {
		msg := "pipelinerun missing ownerrefs %s:%s"
		r.eventRecorder.Eventf(pr, v1.EventTypeWarning, msg, pr.Namespace, pr.Name)
		log.Info(msg, pr.Namespace, pr.Name)
		return artifactbuild.RemovePipelineFinalizer(ctx, pr, r.client)
	}
	ownerName := ""
	for _, ownerRef := range ownerRefs {
		if strings.EqualFold(ownerRef.Kind, "dependencybuild") || strings.EqualFold(ownerRef.Kind, "dependencybuilds") {
			ownerName = ownerRef.Name
			break
		}
	}
	if len(ownerName) == 0 {
		msg := "pipelinerun missing dependencybuild ownerrefs %s:%s"
		r.eventRecorder.Eventf(pr, v1.EventTypeWarning, "MissingOwner", msg, pr.Namespace, pr.Name)
		log.Info(msg, pr.Namespace, pr.Name)

		return artifactbuild.RemovePipelineFinalizer(ctx, pr, r.client)
	}
```
but you are not at the top of this method .... maybe that is because in the ab case, you are certain the finalizer would not have been added?

even if that is the case, adding the call there as well feels like a benign, better safe than sorry sort of thing

gabemontero (commented):
sorry for the review choppiness ... maybe found a couple of more leaks ... see what you think

gabemontero (approved):
/lgtm

PR #605, opened 2023-04-25
STONEBLD-1088 Add rebuilt artifact

rnc:


java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/rebuilt/RebuiltDownloadCommand.java
```diff
@@ -0,0 +1,191 @@
...
+
+            String image = spec.getImage();
+            int lastColon = image.lastIndexOf(':');
+            String repository = image.substring(0, lastColon);
+            String tag = image.substring(lastColon + 1);
+
+            System.out.println("Downloading image from " + repository + " with tag " + tag);
+
+            // Pull the image from the remote registry
+            dockerClient.pullImageCmd(repository).withAuthConfig(dockerClient.authConfig()).withTag(tag).start()
```
stuartwdouglas, line 126: I would use Jib for this rather than the docker client, as AFAIK this requires a running docker demon on the users computer. An example would be the code here: https://github.com/redhat-appstudio/jvm-build-service/blob/b4031912abe34ec4a89425d43351628489bc6d4d/java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/client/ociregistry/OCIRegistryRepositoryClient.java#L134

stuartwdouglas (changes requested):

stuartwdouglas (approved):

stuartwdouglas (changes requested):
Actually the failure is related. You will need to change https://github.com/redhat-appstudio/jvm-build-service/blob/af62f0b6564f62b0d8fe329606189a068374e7c5/java-components/cli/src/main/docker/Dockerfile.all-in-one#L6 to use `-DskipTests` instead.

Once you introduce test-jar artifacts into a build -Dmaven.test.skip becomes problematic, as it does not build the test jars.

rnc:
> 

Thanks, I tried to view the failure logs but I don't have access

stuartwdouglas (approved):

PR #608, opened 2023-04-28
STONEBLD-751: Add excludes for verify command

dwalluck:


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/verifier/VerifyBuiltArtifactsCommand.java
```diff
@@ -157,34 +231,43 @@ private void initMaven(Path globalSettingsFile, Path settingsFile) throws IOExce
...
+            }
+
+            return numFailures;
         } catch (OutOfMemoryError e) {
             //HUGE hack, but some things are just too large to diff in memory
             //but we would need a complete re-rewrite to handle this
             //these are usually tools that have heaps of classes shaded in
             //we just ignore this case for now
             Log.errorf(e, "Failed to analyse %s as it is too big", file);
-            return false;
```
stuartwdouglas, line 187: This should return 0. If we run out of memory we still need to be able to deploy it.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/verifier/VerifyBuiltArtifactsCommand.java
```diff
@@ -157,34 +231,43 @@ private void initMaven(Path globalSettingsFile, Path settingsFile) throws IOExce
...
+    }
+
+    private int handleJar(Path file, String coords, List<String> excludes) {
         try {
             var optionalRemoteFile = resolveArtifact(coords, remoteRepositories, session, system);
 
             if (optionalRemoteFile.isEmpty()) {
                 Log.warnf("Ignoring missing artifact %s", coords);
-                return false;
+                return 1;
```
stuartwdouglas, line 251: This should not fail

stuartwdouglas (changes requested):

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/verifier/VerifyBuiltArtifactsCommand.java
dwalluck: Obviously, `false` should have been `0` and not `1`. I will fix it. The general idea was to return `1` for generic errors, and `1 + count` for an error list of size `count`.

stuartwdouglas:
@dwalluck another class of failures that I think are ok is if there is new 'synthetic bridge' methods generated. Older compilers didn't generate them, but newer ones do. 

That said if we did fail the JDK7 build would probably pass verification, so maybe that is not an issue. A lot of these issues seem to be resulting from compilation with a newer JDK, so maybe if we enable enforcement it won't be an issue.

stuartwdouglas:
I rebased https://github.com/redhat-appstudio/jvm-build-service/pull/448 on top of your PR to see what the results look like with verification enabled.

stuartwdouglas:
So looking at https://github.com/redhat-appstudio/jvm-build-service/pull/448 even building with JDK7 does not fix the synthetic methods issue, so we should definitely ignore new public methods with 'synthetic bridge' access flags, as these are expected.

stuartwdouglas:
I think we should merge this, and then track further work in different issues. @dwalluck does this sounds ok to you?

PR #615, opened 2023-05-15
Fix codegen

stuartwdouglas:


stuartwdouglas:
@gabemontero so my 'make generate' started adding this change and I have no idea why. The really odd thing is that on CI both the original version and this PR come up as being clean (i.e. when CI runs make generate it does not modify the code at all). 

Just wondering if you have any explanation as to what might be going on here?

gabemontero:
> @gabemontero so my 'make generate' started adding this change and I have no idea why. The really odd thing is that on CI both the original version and this PR come up as being clean (i.e. when CI runs make generate it does not modify the code at all).
> 
> Just wondering if you have any explanation as to what might be going on here?

My guess @stuartwdouglas is that the version of "something* from https://github.com/redhat-appstudio/jvm-build-service/blob/main/hack/update-codegen.sh on your system is different than the version used in CI.

I just tried it and I see updates too.

Now, I am running with golang 1.19 locally, but I see that the github action that calls `make generate` uses 1.18 over at https://github.com/redhat-appstudio/jvm-build-service/blob/main/.github/workflows/go-ci.yaml#L25  

That is the only spot I see where `make generate` happens.

That might have something to do with it.  Which golang are you using?

stuartwdouglas:
The really weird thing was that neither of them produced an error in CI, I was seeing changes locally, CI was passing. I pushed those changes, and CI still passed. 

I am using go1.18.6

PR #618, opened 2023-05-15
STONEBLD-1277 Add diagnostic dockerfile functionality to CLI

rnc:


deploy/patch-yaml.sh
```diff
@@ -18,7 +18,7 @@ if [ "$(uname)" = "Darwin" ]; then
 fi
 
 DIR=`dirname $0`
-find $DIR -name development -exec rm -r {} \;
+rm -rf ./operator/overlays/development ./overlays/development
```
stuartwdouglas, line 21: The scripts need to use $DIR as the base dir, otherwise they will not work correctly when executed from another directory.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -331,19 +332,29 @@ func createPipelineSpec(tool string, commitTime int64, jbsConfig *v1alpha12.JBSC
...
+		"\nUSER 0" +
+		"\nWORKDIR /root" +
 		"\nENV CACHE_URL=" + doSubstitution("$(params."+PipelineCacheUrl+")", paramValues, commitTime, buildRepos) +
-		"\nCOPY --from=build-request-processor /deployments/ /root/build-request-processor" +
-		"\nCOPY --from=build-request-processor /lib/jvm/jre-17 /root/system-java" +
-		"\nCOPY --from=cache /deployments/ /root/cache" +
-		"\nRUN mkdir -p /root/workspace && mkdir -p /root/settings && microdnf install vim" +
+		"\nENV BUILD_POLICY_DEFAULT_STORE_LIST=central,redhat,jboss,gradleplugins,confluent,gradle,eclipselink,jitpack,jsweet,jenkins,spring-plugins,dokkadev,ajoberstar,googleandroid,kotlinnative14linux,jcs,kotlin-bootstrap,kotlin-kotlin-dependencies" +
+		"\nRUN mkdir -p /root/project /root/software/settings && microdnf install vim curl procps-ng bash-completion" +
+		"\nRUN wget -qO- https://github.com/swsnr/mdcat/releases/download/mdcat-2.0.3/mdcat-2.0.3-x86_64-unknown-linux-musl.tar.gz | tar -xz -C /opt/" +
```
stuartwdouglas, line 340: I don't really know if this is worth it, vs just displaying the markdown. At the very least it should verify the sha of what it has downloaded.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -331,19 +332,29 @@ func createPipelineSpec(tool string, commitTime int64, jbsConfig *v1alpha12.JBSC
...
+		"\nRUN echo " + base64.StdEncoding.EncodeToString([]byte("#!/bin/sh\n/root/software/system-java/bin/java -jar /root/software/build-request-processor/quarkus-run.jar "+doSubstitution(strings.Join(preprocessorArgs, " "), paramValues, commitTime, buildRepos)+"\n")) + " | base64 -d >/root/preprocessor.sh" +
 		"\nRUN echo " + base64.StdEncoding.EncodeToString([]byte(doSubstitution(build, paramValues, commitTime, buildRepos))) + " | base64 -d >/root/build.sh" +
-		"\nRUN echo " + base64.StdEncoding.EncodeToString([]byte("#!/bin/sh\n/root/settings.sh\n/root/preprocessor.sh\ncd /root/workspace/workspace\n/root/build.sh "+strings.Join(extractArrayParam(PipelineGoals, paramValues), " "))) + " | base64 -d >/root/run-full-build.sh" +
-		"\nRUN chmod +x /root/*.sh"
+		"\nRUN echo " + base64.StdEncoding.EncodeToString([]byte("#!/bin/sh\n/root/settings.sh\n/root/preprocessor.sh\ncd /root/project/workspace\n/root/build.sh "+strings.Join(extractArrayParam(PipelineGoals, paramValues), " ")+"\n")) + " | base64 -d >/root/run-full-build.sh" +
+		"\nRUN chmod +x /root/*.sh" +
+		"\n# Both of these should use jbs remote URL as the user may not have them locally." +
```
stuartwdouglas, line 354: I don't think the docs should be here, they should be in the CLI command. 

Most of the docs are about how to build and run the images, so if you get here you have already done that.

stuartwdouglas (changes requested):

pkg/reconciler/dependencybuild/buildrecipeyaml.go
rnc: I agree there are pros/cons - I was trying to make it a nicer 'user experience' and being able to highlight to the user what they should do. I do agree about the SHA though.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
rnc: Do you mean store the remote doc under a different directory?

rnc: Now split the doc so build instructions are in the CLI and execution instructions (e.g. cache) are in the image.

stuartwdouglas: I would keep it even simpler. I would:
- use plain text instructions so it does not need the mdcat dependency. 
- Embed the plain text into the golang code (see the go:embed module, this file has examples already for the build scripts).
- Use the base64 trick I am using above so it ends up the the dockerfile

I would actually just put the instructions directly in the entry script so you only need a single line to handle it.

stuartwdouglas (changes requested):

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildDiagnosticCommand.java
```diff
@@ -0,0 +1,122 @@
...
+                            .get(i)
+                            .getJavaVersion();
+                }
+                System.out.println("Writing " + fileName);
+                Files.writeString(Paths.get(targetDirectory.toString(), fileName),
+                        dockerFiles.get(i));
+            }
+            System.out.println(CommandLine.Help.Ansi.AUTO
+                    .string("""
+                            \n@|green To build the image, a standard build command e.g. |@@|yellow podman build . |@@|green (or |@@|yellow podman build -f <Dockerfile> .|@@|green ) is sufficient.
```
stuartwdouglas, line 103: Minor nit, but can we list the commands for the specific image generated?

e.g. Something like


```
To diagnose the JDK8 Build:

podman build . -f Dockerfile.somebuild.failed.jdk8 -t somebuild.jdk8
podman run -it somebuild.jdk8

To diagnose the JDK11 Build:

podman build . -f Dockerfile.somebuild.failed.jdk11 -t somebuild.jdk11
podman run -it somebuild.jdk11
```

rnc: Done. Typical output now looks like:

```
java -jar target/quarkus-app/quarkus-run.jar build diagnostic -b=0aeae1d9bc1758ff9e368b527280024b
Target directory: /home/rnc/Work/Apheleia/jvm-build-service/java-components/cli
Selected build: 0aeae1d9bc1758ff9e368b527280024b of https://github.com/jvm-build-service-code/wsdl4j.git:1.6.3

To diagnose the JDK8 build:

podman build -f Dockerfile.wsdl4j.succeed.jdk8 -t wsdl4j.succeed.jdk8 .
podman run -it wsdl4j.succeed.jdk8
```

pkg/reconciler/dependencybuild/buildrecipeyaml.go
rnc: Moved all instructions into `entry-script.sh`, removed use of mdcat and removed all `wget` usage.

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildDiagnosticCommand.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        int lastSlash = theBuild.getSpec().getScm().getScmURL().lastIndexOf('/') + 1;
+        String name = theBuild.getSpec().getScm().getScmURL().substring(lastSlash).replaceFirst("\\.git.*", "");
+        System.out.println("Target directory: " + targetDirectory);
+        System.out.println(
+                "Selected build: " + theBuild.getMetadata().getName() + " of " + theBuild.getSpec().getScm().getScmURL() + ':' +
+                        theBuild.getSpec().getVersion() + '\n');
+
+        List<String> dockerFiles = theBuild.getStatus().getDiagnosticDockerFiles();
+        if ((theBuild.getStatus().getFailedBuildRecipes() == null ? 0 : theBuild.getStatus().getFailedBuildRecipes().size())
```
stuartwdouglas, line 79: This code does not work if the build has failed (which is generally the case when you are using this).

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildDiagnosticCommand.java
```diff
@@ -0,0 +1,122 @@
...
+                }
+                fileName = "Dockerfile." + tagName;
+                System.out.println(
+                        CommandLine.Help.Ansi.AUTO
+                                .string(
+                                        """
+                                                @|green To diagnose the JDK%s build:
+
+                                                |@@|yellow podman build -f %s -t %s .
+                                                podman run -it %s|@
```
stuartwdouglas, line 109: This does not actually work quite right, it gets tagged as localhost/foo, but podman run foo does not work. This would work with docker.

```
Successfully tagged localhost/kotlinx-datetime.failed.jdk11:latest
d11f6aa29f1555324eda8f4d18f56c3d489730b77fd678f60904d39fb1d295aa
[root@6c29787e7b8f project]# podman run -it kotlinx-datetime.failed.jdk11
WARN[0000] "/" is not a shared mount, this could cause issues or missing mounts with rootless containers 
WARN[0000] Using rootless single mapping into the namespace. This might break some images. Check /etc/subuid and /etc/subgid for adding sub*ids if not using a network user 
? Please select an image: 
  ▸ registry.access.redhat.com/kotlinx-datetime.failed.jdk11:latest
    registry.redhat.io/kotlinx-datetime.failed.jdk11:latest
    docker.io/library/kotlinx-datetime.failed.jdk11:latest
```

stuartwdouglas (changes requested):

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildDiagnosticCommand.java
rnc: Strange, I tested with podman (on linux) and it worked ok - I can take another look.

stuartwdouglas:
I tested this out today and it worked great.

PR #635, opened 2023-05-23
Add a way to confiure a base dir for cloning recipe repositories

aloubyansky:
This is to allow a client to clean up the disk after it's done.


stuartwdouglas (approved):
LGTM, but I can't merge till after summit due to code freeze.

PR #636, opened 2023-05-24
STONEBLD-1085 Merge all build steps into a single step

stuartwdouglas:
This provides a significant memory saving, as the pipeline pods now require less that 50% of the memory they required before.

A standard build was consuming close to 3G before this change, now it is slightly over 1G.


stuartwdouglas:
This has significantly better memory usage but fails on JDK7 builds as JDK7 can't run the preprocessor/deploy tasks.

stuartwdouglas:
So this breaks the JDK7 builds, because JDK7 builds are still UBI7.

I am not really sure how to address that, JDK7 is a bit of a nightmare.

PR #641, opened 2023-05-26
Gradle + Kotlin Build Improvements

stuartwdouglas:


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommand.java
```diff
@@ -212,8 +212,10 @@ private void doBuildAnalysis(String scmUrl, String scmTag, String context, Build
                 inv.add(GRADLE);
                 inv.addAll(GradleUtils.getGradleArgs(path));
                 if (skipTests) {
-                    inv.add("-x");
-                    inv.add("test");
+                    //this looks wrong (normally this would be -x test)
+                    //however this is read by an init script that conditionally disables the test plugin
+                    //-x test will fail if any module in the project does not have tests configured
+                    inv.add("-DskipTests=true");
```
rnc, line 218: Which script are you referring to? Excluding a custom task (e.g. `./gradlew -x myFunctionalTest`) will fail if it doesn't exist but I haven't ever seen `-x test` fail even if there are no test sources - can you show me an example of where it does (and with which Gradle version) please?

stuartwdouglas: Oops, I missed adding the script. 

The issue is this one: https://github.com/Kotlin/kotlinx-datetime/blob/bbbbd9c1de0689dadb04d62de6221d3e7497b4f0/build.gradle.kts#L3

I think because it sets up the kotlin plugin in settings.kts, then disables it in the build file there is no test task and -x test fails.

stuartwdouglas:
Damn, I don't think the multiplatform fixes are working as expected, publishing ends up being skipped.

PR #645, opened 2023-05-30
STONEBLD-845 use a mutex rather than a sync map

stuartwdouglas:
If there are multiple notifactions for the same DB in quick success the map access could race, resulting in an inconsistent state.

### DIFF
--- a/openshift-with-appstudio-test/e2e/periodic_test.go
+++ b/openshift-with-appstudio-test/e2e/periodic_test.go
@@ -115,7 +115,8 @@ func TestServiceRegistry(t *testing.T) {
 		var dbContaminatedCount uint32
 		var dbBuildingCount uint32
 		var createdDB int32
-		state := sync.Map{}
+		stateMap := map[string]string{}
+		stateMutex := sync.Mutex{}
 		exitForLoop := false
 		timeoutChannel := time.After(2 * time.Hour)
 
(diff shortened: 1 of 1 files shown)


openshift-with-appstudio-test/e2e/periodic_test.go
```diff
@@ -144,25 +167,34 @@ func TestServiceRegistry(t *testing.T) {
 				if !ok {
 					continue
 				}
-				s, k := state.Load(db.Name)
+				stateMutex.Lock()
```
stuartwdouglas, line 170: @gabemontero we occasionally see the tests fail, even though all builds have completed successfully, with a the building count being 1 and the completed count not being updated. 

My theory is that if you get to notifications for the same DependencyBuild in quick success this state map might be racey.

Although as I am writing this I have realized that this won't help if there is a chance notification can be processed out of order, so this PR probably won't help.

Basically if I a `DependencyBuildStateBuilding` and a `DependencyBuildStateComplete` notification triggered basically at the same time, then if the `DependencyBuildStateComplete` notification is processed before the `DependencyBuildStateBuilding` notification then the completion and building count will not match reality. 

I need to think about this some more.

stuartwdouglas: That would work, but would mean that when the flake happens it would still wait for timeout before reporting success. Maybe the best solution is to just have a periodic check on the actual state of everything outstanding to make sure there is no flakes, which is a bit yuck but probably the best way to handle it.

stuartwdouglas: I have gone for adding a 'freshness check' in the sync block, so we only actually process the state transition if the object is current. 

This should stop state transitions from being processed out of order.

openshift-with-appstudio-test/e2e/periodic_test.go
```diff
@@ -144,25 +167,53 @@ func TestServiceRegistry(t *testing.T) {
...
+				//net result will be correct as the state changes themselves will be processed in the correct order
+				s, k := stateMap[db.Name]
+				current, err := jvmClient.JvmbuildserviceV1alpha1().DependencyBuilds(ta.ns).Get(ctx, db.Name, metav1.GetOptions{})
+				if err != nil {
+					ta.Logf("Failed to get latest DependencyBuild version" + err.Error())
+					stateMutex.Unlock()
+					break
+				}
+				if current.ResourceVersion != db.ResourceVersion {
+					ta.Logf(fmt.Sprintf("Skipping watch event as DependencyBuild %s was not the most recent version", db.Name))
```
stuartwdouglas, line 186: Looking at how many times this message is generated I think this is clearly not just theoretical.

PR #656, opened 2023-06-06
Propagate verification results to the pipeline

stuartwdouglas:
This makes the easily viewable on the test report


stuartwdouglas:
@dwalluck FYI

PR #663, opened 2023-06-06
STONEBLD-753  Add additionalBuilds support

rnc:
…ure-java


java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/build/CustomDeserialiser.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public CustomDeserialiser(Class<BuildRecipeInfo> vc) {
+        super(vc);
+    }
+
+    @Override
+    public Map<String, BuildRecipeInfo> deserialize(JsonParser p, DeserializationContext ctxt)
+            throws IOException {
+        BuildRecipeInfo parentBuild = (BuildRecipeInfo) p.getParsingContext().getParent().getCurrentValue();
```
stuartwdouglas, line 27: I don't really understand this merge logic? 

I am guessing it is trying to use the parent values and merge them in, but AFAIK there is no guarantee the parent is fully constructed at this point? (i.e. do you get different outcomes based on where the additionalBuilds field appears in the YAML)?

Personally I don't think we need this, if some info needs to be copied it is not a big deal, but it has the advantage of making the build recipe be very explicit.

java-components/build-recipes-database/src/test/resources/test-recipes/build-info/github.com/lz4/lz4-java/build.yaml
```diff
@@ -0,0 +1,18 @@
+---
+enforceVersion: true
+additionalArgs:
+  - "-Dfoo=foo"
+preBuildScript: |
+    sed -i -e 's|2\.7\.8|2.5.3|g;' -e 's|-SNAPSHOT|.0|g;' ivy.xml ivy-pure-java.xml
+postBuildScript: |
+    mvn -B -e -s "$(workspaces.build-settings.path)/settings.xml" org.apache.maven.plugins:maven-deploy-plugin:3.1.0:deploy-file "-Durl=file:$(workspaces.source.path)/artifacts" -DpomFile=dist/lz4-java-$(params.ENFORCE_VERSION).pom -Dfile=dist/lz4-java-$(params.ENFORCE_VERSION).jar -Dfiles=dist/lz4-java-$(params.ENFORCE_VERSION)-sources.jar,dist/lz4-java-$(params.ENFORCE_VERSION)-javadoc.jar -Dclassifiers=sources,javadoc -Dtypes=jar,jar
+additionalDownloads:
```
stuartwdouglas, line 9: This illustrates the problem with merging. The pure java version does not need this, but there is no way to un-require it if they are merged.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommand.java
```diff
@@ -51,6 +52,7 @@ public class LookupBuildInfoCommand implements Runnable {
     @CommandLine.Option(names = "--scm-tag", required = true)
     String tag;
 
+    // TODO: Is this a subdirectory to operate upon?
```
stuartwdouglas, line 55: yes

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -165,7 +165,8 @@ func createPipelineSpec(tool string, commitTime int64, jbsConfig *v1alpha12.JBSC
 		gitArgs = "echo \"$GIT_TOKEN\"  > $HOME/.git-credentials\nchmod 400 $HOME/.git-credentials\n"
 		gitArgs = gitArgs + "echo '[credential]\n        helper=store\n' > $HOME/.gitconfig\n"
 	}
-	gitArgs = gitArgs + "git clone $(params." + PipelineParamScmUrl + ") $(workspaces." + WorkspaceSource + ".path)/workspace && cd $(workspaces." + WorkspaceSource + ".path)/workspace && git reset --hard $(params." + PipelineParamScmHash + ")"
+	// ### Total hack to remove fragment component. Tekton substitutes these params.xxx .... is there a better way of handling this?
```
stuartwdouglas, line 168: yes, we should remove it in the go code that passes the parameters into tekton. See dependencybuild.go where PipelineParamScmUrl is set for the pipeline.

stuartwdouglas (changes requested):

java-components/build-recipes-database/src/test/resources/test-recipes/build-info/github.com/lz4/lz4-java/build.yaml
rnc: The pure-java version just generates different files to the dist directory - without this deploy command nothing gets deployed I thought?

java-components/build-recipes-database/src/main/java/com/redhat/hacbs/recipies/build/CustomDeserialiser.java
rnc: Ok, hadn't seen that about parent construction which means it was order dependent - I'll remove that and update the doc in the upcoming pr for the jvm-build-data.

java-components/build-recipes-database/src/test/resources/test-recipes/build-info/github.com/lz4/lz4-java/build.yaml
stuartwdouglas: I mean the additionalDownloads section, the java version should not need the native dependencies.

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -1038,3 +1042,18 @@ func RemovePipelineFinalizer(ctx context.Context, pr *pipelinev1beta1.PipelineRu
 	}
 	return reconcile.Result{}, nil
 }
+
+// This is to remove any '#xxx' fragment from a URI so that git clone commands don't need seperate adjustment
+func modifyURLFragment(log logr.Logger, paramValues *[]pipelinev1beta1.Param) {
```
stuartwdouglas, line 1047: I would modify this before it ends up in the parameter array. Basically instead of using `db.Spec.ScmInfo.SCMURL` around line 563 when setting the parameter, store this in a variable and strip it from the variable.

stuartwdouglas (changes requested):

pkg/reconciler/dependencybuild/dependencybuild.go
rnc: Sure. Out of curiosity what is `PipelineParamChainsGitUrl` used for?

PR #666, opened 2023-06-07
STONEBLD-1354 Create image repo automatically

stuartwdouglas:


pkg/reconciler/jbsconfig/jbsconfig.go
```diff
@@ -634,3 +695,110 @@ func (r *ReconcilerJBSConfig) processSpiImageSecret(ctx context.Context, config
...
+
+	repo, robotAccount, err := r.generateImageRepository(log, config)
+	if err != nil {
+		return err
+	}
+	if repo == nil || robotAccount == nil {
+		return errors2.New("unknown error in the repository generation process")
+	}
+	controllerutil.AddFinalizer(config, ImageRepositoryFinalizer)
+	config.Spec.Owner = r.quayOrgName
```
stuartwdouglas, line 712: TODO before merging: create Status fields for these so this is not updating the spec.

PR #675, opened 2023-06-12
STONEBLD-751: Add verification excludes support to build recipes

dwalluck:


pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
```diff
@@ -84,6 +84,7 @@ type BuildRecipe struct {
 	DisableSubmodules   bool                 `json:"disableSubmodules,omitempty"`
 	AdditionalMemory    int                  `json:"additionalMemory,omitempty"`
 	Repositories        []string             `json:"repositories,omitempty"`
+	Excludes            []string             `json:"excludes,omitempty"`
```
stuartwdouglas, line 87: You don't need this as the values are encoded into the script

java-components/build-recipes-database/src/test/resources/test-recipes/build-info/github.com/quarkusio/quarkus/_version/1.9999/build.yaml
```diff
@@ -1,3 +1,5 @@
 ---
 additionalArgs:
   - "-Dquarkus1=true"
+excludes:
```
stuartwdouglas, line 4: Sorry I did not pick up on this earlier, but I think we need a more descriptive name for this (and all the downstream stuff like CRDs etc). 

'excludes' does not really mean much, maybe 'allowedDifferences'?

stuartwdouglas (changes requested):

stuartwdouglas (approved):

PR #690, opened 2023-06-20
STONEBLD-1359 Add Status labels to ArtifactBuild

rnc:


pkg/reconciler/util/util.go
```diff
@@ -2,6 +2,8 @@
 
 import (
 	"context"
+	"crypto/md5"
```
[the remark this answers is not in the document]
stuartwdouglas: You need to add the //#nosec comments, see the current version of artifactbuild.go

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/artifactwatch/ScmLookup.java
```diff
@@ -83,9 +83,17 @@ public void onAdd(ArtifactBuild newObj) {
                             newObj.getStatus().setMessage("");
                             newObj.getStatus().setState(ArtifactBuild.DISCOVERING);
                         } catch (Exception e) {
-                            Log.errorf(e, "Failed to update rebuilt object");
+                            Log.errorf(e, "Failed to update rebuilt object " + i);
                             newObj.getStatus().setMessage(e.getMessage());
                             newObj.getStatus().setState(ArtifactBuild.MISSING);
+                            // Matches status labels in pkg/reconciler/util/util.go
```
stuartwdouglas, line 89: Don't update the labels here. The operator will take care of it. If you do an update here the operator will attempt to reverse the change, as the status has not been updated yet, and will likely cause 'already modified' exceptions.

This whole approach of doing this change in the cache is not 100% idiomatic, but saves a huge amount of memory and CPU. Basically ideally the operator would kick of a pipeline, the pipeline would return this info in a result, and then the operator would act on it to update the objects, but that kicks off hundreds of pipelines in some cases.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/artifactwatch/ScmLookup.java
```diff
@@ -64,11 +64,11 @@ public void onAdd(ArtifactBuild newObj) {
...
                         try {
                             if (newObj.getMetadata().getAnnotations() != null
-                                    && newObj.getMetadata().getAnnotations().containsKey(REBUILD)) {
+                                    && newObj.getMetadata().getAnnotations().containsKey(ModelConstants.REBUILD)) {
                                 //if this is a forced rebuild we always update the SCM info
                                 //there is a good chance there may be a new recipe
                                 recipeManager.forceUpdate();
-                                newObj.getMetadata().getAnnotations().remove(REBUILD);
+                                newObj.getMetadata().getAnnotations().remove(ModelConstants.REBUILD);
                                 client.resource(newObj).patch();
```
stuartwdouglas, line 72: Good point about this not being consistent. We should probably return here after this change, and it will get picked up on the next invocation.

pkg/reconciler/artifactbuild/artifactbuild.go
```diff
@@ -567,6 +573,40 @@ func (r *ReconcileArtifactBuild) handleCommunityDependencies(ctx context.Context
 	return nil
 }
 
+func (r *ReconcileArtifactBuild) updateLabel(ctx context.Context, log logr.Logger, abr *v1alpha1.ArtifactBuild) (bool, error) {
+	result := false
+
+	if abr.Labels == nil {
+		abr.Labels = make(map[string]string)
```
stuartwdouglas, line 580: It does not really matter, but everywhere else in the code I have been using `map[string]string{}` to create an empty map.

stuartwdouglas (changes requested):

pkg/reconciler/artifactbuild/artifactbuild.go
rnc: When I looked in the golang docs on how to create maps, most seemed to highlight the make built-in which is why I used that I'm not sure if there is any difference internally between the two? From what I can see there doesn't seem to be though the make allows 'hinting' of initial allocation requirements.
I've done a search and while the code is using both patterns, Gabe used both on the same commit :-)

PR #698, opened 2023-06-27
STONEBLD-1437 Use ownerreferences for jbsconfig

rnc:


pkg/reconciler/jbsconfig/jbsconfig.go
```diff
@@ -34,11 +34,14 @@ import (
...
 
-const TlsServiceName = v1alpha1.CacheDeploymentName + "-tls"
-const TestRegistry = "jvmbuildservice.io/test-registry"
-const ImageRepositoryFinalizer = "jvmbuildservice.io/quay-repository-finalizer"
-const DeleteImageRepositoryAnnotationName = "image.redhat.com/delete-image-repo"
-const UploadSecretName = "jvm-build-service-temp-upload-secret" //#nosec
+const (
+	TlsServiceName                      = v1alpha1.CacheDeploymentName + "-tls"
+	TestRegistry                        = "jvmbuildservice.io/test-registry"
+	JBSConfigFinalizer                  = "jvmbuildservice.io/jbsconfig-finalizer"
```
stuartwdouglas, line 40: Looks like this is left over from the previous attempt? With the ownership model you should not need the finalizer?

rnc: I kept that bit to ensure that `handlePossibleRepositoryCleanup` would be called

stuartwdouglas: That is handled by the ImageRepositoryFinaliser below though?

stuartwdouglas: Oh, you changed the name, I guess the one below is the one that could be removed.

stuartwdouglas: Actually I just had a look at the existing ImageRepositoryFinaliser code looks like it is still all that is needed.

pkg/reconciler/artifactbuild/artifactbuild.go
```diff
@@ -219,10 +219,9 @@ func removePipelineFinalizer(ctx context.Context, pr *pipelinev1beta1.PipelineRu
 }
 
 func setArtifactState(log *logr.Logger, abr *v1alpha1.ArtifactBuild, state string) {
-	abr.Status.State = state
-	switch state {
-	case v1alpha1.ArtifactBuildStateFailed, v1alpha1.ArtifactBuildStateComplete:
-		log.Info(fmt.Sprintf("ArtifactBuild %s has state set to %s", abr.Name, state))
+	if abr.Status.State != state {
+		log.Info(fmt.Sprintf("ArtifactBuild %s changing state from %s to %s", abr.Name, abr.Status.State, state))
+		abr.Status.State = state
```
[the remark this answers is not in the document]
rnc: I'm not sure about that - perhaps we can discuss offline. I am seeing lines like 
```
setArtifactState(&log, abr, v1alpha1.ArtifactBuildStateFailed)
return reconcile.Result{}, r.client.Status().Update(ctx, abr)
```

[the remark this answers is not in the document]
stuartwdouglas: Yes, but that particular code is the state machine for other states, it would only be a no-op if it was in `handleStateFailed`.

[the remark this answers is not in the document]
rnc: I've rolled this change back.

stuartwdouglas (approved):

PR #702, opened 2023-07-10
Clean up cache config

stuartwdouglas:
Remove a heap of config that only took effect when running locally, it was confusing and not really useful.


java-components/cache/src/main/resources/application.properties
```diff
@@ -8,53 +8,23 @@ quarkus.container-image.group=redhat-appstudio
...
+
+# default repo config
 build-policies=default
-%test.build-policies=default,central-only,prefer-rebuilt
-build-policy.default.store-list=rebuilt,central,jboss,redhat
-build-policy.central-only.store-list=central
-build-policy.prefer-rebuilt.store-list=rebuilt,redhat,central
-%dev.build-policy.default.store-list=rebuilt,central,jboss,gradleplugins,confluent,gradle,eclipselink,redhat,jitpack,jsweet,jenkins,spring-plugins,dokkadev,ajoberstar,googleandroid,kotlinnative14linux,jcs,kotlin-bootstrap,kotlin-kotlin-dependencies
 store.central.url=https://repo.maven.apache.org/maven2
 store.central.type=maven2
```
rnc, line 18: Why is this one being left in?

java-components/cache/src/main/resources/application.properties
```diff
@@ -8,53 +8,23 @@ quarkus.container-image.group=redhat-appstudio
...
 quarkus.jib.working-directory=/deployments
 quarkus.micrometer.enabled=true
-%dev.quarkus.kubernetes-client.trust-certs=true
-%dev.quarkus.log.category."com.redhat.hacbs.artifactcache".level=DEBUG
 quarkus.aws.devservices.localstack.image-name=localstack/localstack:0.11.5
 quarkus.http.limits.max-body-size=1000M
 quarkus.jib.user=0
+
+# default repo config
 build-policies=default
```
rnc, line 16: The configuration.adoc says `By default this cache is configured to have a `default` and `central-only` configuration as follows: build-policies=default,central-only`  ; this seems inconsistent?

PR #708, opened 2023-07-13
Test with Pipelines 1.11

stuartwdouglas:


deploy/minikube-ci.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 DIR=`dirname $0`
-kubectl apply -f https://github.com/tektoncd/pipeline/releases/download/v0.41.1/release.yaml
+kubectl apply -f https://github.com/tektoncd/pipeline/releases/download/v0.44.3/release.yaml
```
gabemontero, line 4: 1.11 is actually at 0.47.2 if you want to match up as close as possible

deploy/minikube-development.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 DIR=`dirname $0`
-kubectl apply -f https://github.com/tektoncd/pipeline/releases/download/v0.41.1/release.yaml
+kubectl apply -f https://github.com/tektoncd/pipeline/releases/download/v0.44.3/release.yaml
```
gabemontero, line 4: 1.11 is actually at 0.47.2 if you want to match up as close as possible

go.mod
```diff
@@ -10,21 +10,21 @@ require (
 	github.com/openshift/client-go v0.0.0-20210521082421-73d9475a9142
 	github.com/prometheus/client_golang v1.15.0
 	github.com/prometheus/client_model v0.3.0
-	github.com/tektoncd/pipeline v0.41.1
+	github.com/tektoncd/pipeline v0.44.3
```
gabemontero, line 13: 1.11 is actually at 0.47.2 if you want to match up as close as possible

gabemontero (changes requested):

stuartwdouglas:
We will need https://github.com/openshift/release/pull/41260 in first before merging this

stuartwdouglas:
I am going to hold this for now an update go in a separate PR

stuartwdouglas:
@psturc I need a bit of advice on how to get the jvm-build-service-in-repo-e2e test passing.

It seems to be failing because it is using go 1.18 to run the tests, while Tekton 0.47.3 requires 1.19 due to the use of atomic.Pointer in a dependency.

I am guessing that I need to change the image that is actually running the tests to an image that contains 1.19, however I am not really sure about the correct way to do that, or if there are other requirements for this image.

I am guessing that I need to add something like this: https://github.com/openshift/release/blob/da8fca55f163319cba2d5897dba1c137bb41003e/ci-operator/step-registry/redhat-appstudio/e2e/redhat-appstudio-e2e-ref.yaml#L19

Although this is a complete guess and because I don't have permission to run rehearsals on the release repo anyway I want to talk to you before I attempt anything.

psturc:
@stuartwdouglas that is strange, because the job is using [e2e-test-runner image](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/708/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1680791234139918336#1:build-log.txt%3A9), which is based on [go-v1.19 image](https://github.com/openshift/release/blob/d49e288d20fab37ba1d3349f3ea9b54d0e7de6a5/ci-operator/config/redhat-appstudio/e2e-tests/redhat-appstudio-e2e-tests-main.yaml#L5)

I will look into it more

stuartwdouglas:
There is an extra line in the passing one:

```
 INFO[2023-07-17T04:07:27Z] Tagging ci/redhat-appstudio:e2e-test-runner into pipeline:ci-redhat-appstudio-e2e-test-runner. 
INFO[2023-07-17T04:07:27Z] Tagging ci/redhat-appstudio-e2e-test-runner:latest into pipeline:root. 
```

The in-repo one only has the second line. Not sure if that is relevant though.

psturc:
@stuartwdouglas [got a fix, currently testing it](https://github.com/openshift/release/pull/41333). the reference to the image that was being used for testing was outdated

PR #713, opened 2023-07-17
STONEBLD-1510 Move to v1 APIs

stuartwdouglas:
Leaving as draft until the Tekton upgrade goes in

### DIFF
--- a/pkg/reconciler/artifactbuild/controller.go
+++ b/pkg/reconciler/artifactbuild/controller.go
@@ -10,7 +10,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/redhat-appstudio/jvm-build-service/pkg/apis/jvmbuildservice/v1alpha1"
-	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	v1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 )
 
 func SetupNewReconcilerWithManager(mgr ctrl.Manager) error {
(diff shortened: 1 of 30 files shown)


pkg/reconciler/artifactbuild/controller.go
```diff
@@ -10,7 +10,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/redhat-appstudio/jvm-build-service/pkg/apis/jvmbuildservice/v1alpha1"
-	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	v1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
```
rnc, line 13: Question - why is this named with beta when its v1 (i.e. couldn't it be e.g. `pipelinev1`)  ?

stuartwdouglas: To stop a massive amount of conflicts with your PR (and every other PR).

The envtest failure is real and I am not sure how to fix it. I don't think this test really adds much value, so to get this in we might just have to delete it.

rnc: Ah I see, yes that would be painful :-) Is the plan to change them in future during a 'quiet' period?

stuartwdouglas: yea, that was the plan.

stuartwdouglas:
Closing for now, I think we can hold of on this as it introduces some extra complexity.

PR #737, opened 2023-07-18
Fix taskrun determination

rnc:


java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildLogsCommand.java
```diff
@@ -112,12 +112,12 @@ public void run() {
 
             System.out.println("---------   Logs for PipelineRun " + pipelineRun.getMetadata().getName() + " ("
                     + (success ? "SUCCEEDED" : "FAILED") + ") ---------");
-            var references = pipelineRun.getStatus().getChildReferences();
+            var references = pipelineRun.getStatus().getTaskRuns().keySet();
```
stuartwdouglas, line 115: What was the issue you were seeing? getTaskRuns is deprecated in newer version of Tekton.

rnc: Huh, ok I didn't know that. I kept getting `No TaskRuns found` attempting to output a log. Maybe I'll wait until all current PRs have been merged and then I'll rebase again to verify.

stuartwdouglas: Hmm, any chance they have been deleted? 

I only know about the deprecation because of my work to upgrade to keep in sync with RHTAP and move to the v1 Kube model.

rnc: I did a fresh rebuild in order to have something to output (in case that was the problem). 

Question - if tekton is being upgraded in STONEBLD-1510 - should the java imports change from e.g. io.fabric8.tekton.pipeline.v1beta1.TaskRun ?

stuartwdouglas: We can do the Tekton upgrade without upgrading to the v1 api, technically the v1 API is in the current version but for some reason it is half complete. 

I have a WIP PR  upgrading but it needs to wait till after the tekton upgrade to actually test it out.

rnc: Given https://github.com/redhat-appstudio/jvm-build-service/pull/713 was closed can this be revisited?

stuartwdouglas: How do you reproduce the issue? I just tested this and it worked fine. I would rather stick with getChildReferences if possible because it will be an issue when we do upgrade.

rnc: I do e.g. `java -Dquarkus.log.level=INFO -jar target/quarkus-app/quarkus-run.jar build logs -b=080dbba8b3ffba35739ebe5bce69a2be` I must admit, I'm a bit confused that it works for you! I can't seem to get any logs out using the original codebase ( I just get `No TaskRuns found`).

rnc: Testing using deployed env (via `make dev ; ./deploy/openshift-development.sh`) on the stonesoupengineering cluster.

rnc:
After some discussion offline we figured out I still had pipelines 1.9.3 instead of 1.11 installed. This meant the tekton version was too old and it didn't populate the child references hence the problem I was seeing. Therefore resolving the installation problem means this PR isn't required.

PR #786, opened 2023-07-27
STONEBLD-1569 Use preprocessor to enforce version

stuartwdouglas:


stuartwdouglas:
This approach has some issues, closing for now.

PR #788, opened 2023-07-27
STONEBLD-1198 Check for preexisting builds

rnc:


java-components/cli/pom.xml
```diff
@@ -52,6 +52,10 @@
             <groupId>com.google.cloud.tools</groupId>
             <artifactId>jib-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>commons-logging-jboss-logging</artifactId>
+        </dependency>
```
rnc, line 58: Without this, CLI operations like download were failing (even from `main`)

stuartwdouglas: Odd, I wonder if it is related to the Quarkus upgrade.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommand.java
```diff
@@ -302,6 +328,64 @@ private void doBuildAnalysis(String scmUrl, String scmTag, String context, Build
                 info.setAllowedDifferences(buildRecipeInfo.getAllowedDifferences());
                 Log.infof("Got build recipe info %s", buildRecipeInfo);
             }
+            if (sharedRegistries != null) {
```
stuartwdouglas, line 331: I would also check the main rebuilt repo here, so it is not treated differently to the shared ones. 

This would also be useful for testing, you could run the existing tests, then in a new namespace reference the same repo and make sure it picks up the artifacts from the first test.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/containerregistry/ContainerRegistryDeployer.java
```diff
@@ -81,12 +82,15 @@ public ContainerRegistryDeployer(
...
+
         String fullName = host + (port == 443 ? "" : ":" + port) + "/" + owner + "/" + repository;
         if (!token.isBlank()) {
             if (token.trim().startsWith("{")) {
-                //we assume this is a .dockerconfig file
-                try (var parser = MAPPER.createParser(token)) {
-                    DockerConfig config = parser.readValueAs(DockerConfig.class);
+                try {
+                    //we assume this is a .dockerconfig file
+                    DockerConfigTemplate config = MAPPER.readValue(token, DockerConfigTemplate.class);
```
rnc, line 92: This was just to simplify the code - using Jib there is an class for the json so seemed simpler to use that.

java-components/resource-model/src/main/java/com/redhat/hacbs/resources/model/v1alpha1/DependencyBuildSpec.java
```diff
@@ -27,13 +23,4 @@ public DependencyBuildSpec setVersion(String version) {
...
-
-    public List<BuildRecipe> getBuildRecipes() {
-        return buildRecipes;
-    }
-
-    public DependencyBuildSpec setBuildRecipes(List<BuildRecipe> buildRecipes) {
-        this.buildRecipes = buildRecipes;
-        return this;
-    }
 }
```
rnc, line 26: These I thought shouldn't be in spec but in status

stuartwdouglas:
FYI https://github.com/redhat-appstudio/jvm-build-service/pull/796 will conflict in a minor way with this. It turns out it is possible to just generate the Java classes to keep them in sync.

rnc:
@stuartwdouglas I don't have permissions to set this back to draft mode

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/containerregistry/ContainerRegistryDeployer.java
```diff
@@ -82,48 +78,12 @@ public ContainerRegistryDeployer(
...
-                password = decoded.substring(pos + 1);
-            }
-        } else {
-            Log.errorf("No token configured");
-            username = null;
-            password = null;
-        }
-        Log.infof("Using username %s to publish to %s/%s/%s", username, host, owner, repository);
-        Log.infof("Prepend tag is %s", prependTag);
+        this.credential = ContainerUtil.processToken(fullName, token);
```
rnc, line 82: Moved to using Credentials across build-request-processor for consistency.

java-components/resource-model/src/main/java/com/redhat/hacbs/resources/model/v1alpha1/Util.java
```diff
@@ -0,0 +1,23 @@
+package com.redhat.hacbs.resources.model.v1alpha1;
+
+import com.redhat.hacbs.resources.model.v1alpha1.jbsconfigstatus.ImageRegistry;
+
+public class Util {
+    public static ImageRegistry parseRegistry(String registry) {
```
rnc, line 6: Moved ImageRegistry parser to model so it may be used by both cache and build-request-processor

pkg/reconciler/jbsconfig/jbsconfig.go
```diff
@@ -631,18 +624,9 @@ func (r *ReconcilerJBSConfig) handleNoImageSecretFound(ctx context.Context, conf
 			binding.Namespace = config.Namespace
 			imageRegistry := config.ImageRegistry()
 			url := "https://"
-			if imageRegistry.Host == "" {
-				url += "quay.io"
-			} else {
-				url += imageRegistry.Host
-			}
+			url += imageRegistry.Host
```
rnc, line 627: Moved the defaulting of `quay.io` to `ImageRegistry()` call to make it more easily reused.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -55,11 +53,9 @@ var hermeticBuildEntryScript string
 
 func createPipelineSpec(tool string, commitTime int64, jbsConfig *v1alpha12.JBSConfig, systemConfig *v1alpha12.SystemConfig, recipe *v1alpha12.BuildRecipe, db *v1alpha12.DependencyBuild, paramValues []pipelinev1beta1.Param, buildRequestProcessorImage string) (*pipelinev1beta1.PipelineSpec, string, error) {
 
-	marshaled, err := json.Marshal(recipe)
-	if err != nil {
-		return nil, "", err
-	}
-	imageId := util.HashString(string(marshaled) + buildRequestProcessorImage + tool + db.Name)
+	// Rather than tagging with hash of json build recipe, buildrequestprocessor image and db.Name as the former two
```
stuartwdouglas, line 56: I don't think we want to change this imageId here. I think we just want a new additional tag with the name.

I don't even know if the current imageId was the correct choice, or if it should have just been random. 

Basically my thinking is that if you ever need to rebuild something because of a problem you might still want to reference the old artifacts, but because we are not renaming currently this will just overwrite all the tags. If we have a random image id then there will always be a reference to the old image so the legacy artifacts will still be accessible (kinda, assuming you have stored this id somewhere). 

That said as this is not really implemented properly anywhere yet making the imageId the name and then creating the concept of an archiveId or something at a later date is probably the way to go.

stuartwdouglas (approved):

stuartwdouglas:
The minikube failure just looks like a timeout, for some reason the builds seen to have run so slow that it did not complete in time.

stuartwdouglas:
Actually both minikube and in-repo-e2e failed in the same way. I will investigate further

stuartwdouglas:
The reason for the 'in-repo-e2e' failures is that the image id needs to have 'prependTag' added to it.

This is something used in testing to make all the tags unique. Currently it is failing because it is just downloading existing artifacts rather than doing the builds as expected.

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -429,8 +429,33 @@ func (r *ReconcileDependencyBuild) handleStateAnalyzeBuild(ctx context.Context,
...
+
+		jbsConfig := &v1alpha1.JBSConfig{}
+		err := r.client.Get(ctx, types.NamespacedName{Namespace: db.Namespace, Name: v1alpha1.JBSConfigName}, jbsConfig)
+		if err != nil && !errors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+		// As the tests utilise prependTag for uniqueness (See also
+		// build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/Gav.java) so check for that
+		// here to avoid reusing images when we want differentiation.
+		prependTag := len(jbsConfig.Spec.PrependTag) > 0
```
stuartwdouglas, line 441: So the problem with this is that then we won't be able to test that this is working. 

If this is set images should be tagged with prependTag + imageId, and the lookup command should be passed it as a parameter to look for prependTag + imageId.

stuartwdouglas (changes requested):

PR #793, opened 2023-07-31
STONEBLD-1515 test multi version images

stuartwdouglas:
This should not be merged as-is, but it a test of using single builder images for multiple JDK versions.

### DIFF
--- a/deploy/operator/config/system-config.yaml
+++ b/deploy/operator/config/system-config.yaml
@@ -5,15 +5,11 @@ metadata:
   name: cluster
 spec:
   builders:
-    jdk8:
-      image: quay.io/redhat-appstudio/hacbs-jdk8-builder:f0c0dbbd5926e7a38af245b4e48a38ddd2d17359
-      priority: 3000
-    jdk11:
-      image: quay.io/redhat-appstudio/hacbs-jdk11-builder:f0c0dbbd5926e7a38af245b4e48a38ddd2d17359
-      priority: 2000
-    jdk17:
+    ubi8:
       image: quay.io/redhat-appstudio/hacbs-jdk17-builder:f0c0dbbd5926e7a38af245b4e48a38ddd2d17359
+      tag: jdk:8;11;17,maven:3.8,gradle:8.0.2;7.4.2;6.9.2;5.6.4;4.10.3,sbt:1.8.0,ant:1.10.13
       priority: 1000
-    jdk7:
+    ubi7:
       image: quay.io/redhat-appstudio/hacbs-jdk7-builder:f0c0dbbd5926e7a38af245b4e48a38ddd2d17359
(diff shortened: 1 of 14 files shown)


deploy/operator/config/system-config.yaml
```diff
@@ -6,8 +6,8 @@ metadata:
 spec:
   builders:
     ubi8:
-      image: quay.io/redhat-appstudio/hacbs-jdk17-builder:f0c0dbbd5926e7a38af245b4e48a38ddd2d17359
-      tag: jdk:8;11;17,maven:3.8,gradle:8.0.2;7.4.2;6.9.2;5.6.4;4.10.3,sbt:1.8.0,ant:1.10.13
+      image: quay.io/redhat-appstudio/pull-request-builds:hacbs-jdk17-builder-1fa75e6c1ff4716b6dd100ba3346e0f25ffe771a
+      tag: jdk:8;11;17,maven:3.8.8;3.9.5,gradle:8.4;8.3;8.0.2;7.4.2;6.9.2,sbt:1.8.0,ant:1.10.13
```
rnc, line 10: If tag is coming from jvm-build-service-builder-images via https://github.com/redhat-appstudio/jvm-build-service/blob/main/deploy/operator/config/kustomization.yaml#L11 does tag need to be specified here as well? Or is this done for testing only?

deploy/overlays/dev-template/config.yaml
```diff
@@ -4,11 +4,6 @@ kind: JBSConfig
...
-  owner: QUAY_USERNAME
-  relocationPatterns:
-    - relocationPattern:
-        buildPolicy: "default"
-        patterns:
-          - pattern:
-              from: (io.github.stuartwdouglas.hacbs-test.simple):(simple-jdk17):(99-does-not-exist)
-              to: io.github.stuartwdouglas.hacbs-test.simple:simple-jdk17:0.1.2
+  registry:
+    owner: QUAY_USERNAME
```
rnc, line 8: This is the same change as in https://github.com/redhat-appstudio/jvm-build-service/pull/880/files#diff-d4012663d2198be9ac5255ec8dcd2078a1a65dec31b3ffb4ce11f36fea664b71 so will conflict. 
Is relocationPatterns no longer required?

deploy/operator/config/system-config.yaml
stuartwdouglas: this is testing only

deploy/overlays/dev-template/config.yaml
stuartwdouglas: I am going to remove this before it goes in. I was having some issues with my cluster and I was wondering if this was the cause.

stuartwdouglas: relocationPatterns was never really required, it was there for testing, and I don't know if we are ever going to really use it.

PR #796, opened 2023-08-01
STONEBLD-1579 Generate Java Kube classes

stuartwdouglas:


java-components/resource-model/src/main/resources/crds/jvmbuildservice.io_artifactbuilds.yaml
```diff
@@ -0,0 +1,85 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
```
rnc, line 5: What is the difference between this CRD and the one in deploy/crds/base/jvmbuildservice.io_artifactbuilds.yaml ?

PR #807, opened 2023-08-08
STONEBLD-1592 Use ResultsUpdater for LookupBuildInfo

rnc:


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommand.java
```diff
@@ -133,19 +127,10 @@ public void run() {
...
         } catch (Exception e) {
             Log.errorf(e, "Failed to process build info for " + scmUrl);
-            if (message != null) {
-                try {
-                    Files.writeString(message, "Failed to analyse build for " + scmUrl + ". Failure reason: " + e.getMessage());
-                } catch (IOException ex) {
-                    Log.errorf(e, "Failed to write result");
-                }
-            }
+            resultsUpdater.get().updateResults(taskRun, Map.of(
```
stuartwdouglas, line 132: Shouldn't this be the message rather than BUILD_INFO? The build info should be JSON

rnc: Ah, I'd trimmed it down so in the error case, it just passed the string back through build_info and processed it 
 https://github.com/redhat-appstudio/jvm-build-service/pull/807/files#diff-1dcd0d035636ba6f1cb8632a7ea1820f33bca4ecdc6a118dfb0f67b0734498f3R309

PR #813, opened 2023-08-14
Update GH Action versions

rnc:


.github/workflows/minikube.yaml
```diff
@@ -83,15 +83,15 @@ jobs:
...
+      - uses: actions/checkout@v3
       - name: Setup Registry
         run: |
           export DEV_IP=172.16.1.1
           sudo ifconfig lo:0 $DEV_IP
           docker run -d -p 5000:5000 --restart=always  registry:2
           sudo echo '{ "insecure-registries": ["172.16.1.1:5000"] }' | sudo tee  /etc/docker/daemon.json
       - name: Start minikube
-        uses: medyagh/setup-minikube@eefe3e32b8e0cbd99330ebb68c5fc2d0c2a5e68c
+        uses: medyagh/setup-minikube@latest
```
stuartwdouglas, line 94: prodsec recommends using explicit hashes for non-trusted actions. Thinking about it the codecov action should probably also be using a hash.

stuartwdouglas (changes requested):

.github/workflows/minikube.yaml
rnc: Removed this change for now.

PR #843, opened 2023-09-13
Add kube-linter check

mmorhun:
This PR adds `kube-linter` check that runs only if change to the operator deployment configuration is made.

### DIFF
--- a/deploy/operator/base/deployment.yaml
+++ b/deploy/operator/base/deployment.yaml
@@ -22,7 +22,7 @@ spec:
         runAsNonRoot: true
       containers:
         - name: hacbs-jvm-operator
-          image: hacbs-jvm-operator
+          image: hacbs-jvm-operator:next
           imagePullPolicy: IfNotPresent
           ports:
             - containerPort: 8080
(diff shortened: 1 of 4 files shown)


deploy/operator/base/deployment.yaml
```diff
@@ -22,7 +22,7 @@ spec:
         runAsNonRoot: true
       containers:
         - name: hacbs-jvm-operator
-          image: hacbs-jvm-operator
+          image: hacbs-jvm-operator:latest
```
rnc, line 25: I'm not sure that would help as https://docs.kubelinter.io/#/generated/checks?id=latest-tag would still complain? Would it not be better to exclude this check in the configuration file?

mmorhun: Changed tag name as @stuartwdouglas suggested.

deploy/operator/base/deployment.yaml
```diff
@@ -22,7 +22,7 @@ spec:
         runAsNonRoot: true
       containers:
         - name: hacbs-jvm-operator
-          image: hacbs-jvm-operator
+          image: hacbs-jvm-operator:next
```
stuartwdouglas, line 25: I think the CI failure is legit, and is likely caused by this update https://github.com/redhat-appstudio/jvm-build-service/blob/91a173d2a2421b0d9e169b9678044a9558001aae/deploy/openshift-ci.sh#L22

PR #847, opened 2023-09-14
STONEBLD-1577 Add tekton-result

rnc:


deploy/tekton-results.sh
```diff
@@ -0,0 +1,35 @@
...
+# Create new TLS Secret from cert.
+kubectl -n ${NAMESPACE} delete --ignore-not-found secret tekton-results-tls
+kubectl create secret tls -n ${NAMESPACE} tekton-results-tls \
+--cert=/tmp/cert.pem \
+--key=/tmp/key.pem
+
+kubectl apply -f pvc.yaml
+# From https://github.com/tektoncd/operator/blob/main/config/crs/kubernetes/result/operator_v1alpha1_result_cr.yaml
+kubectl apply -f tekton_operator_v1alpha1_result_cr.yaml
+kubectl apply -f tekton_route.yaml
```
rnc, line 35: Should I combine these yaml files into one file?

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildLogsCommand.java
```diff
@@ -39,10 +55,13 @@ public class BuildLogsCommand implements Runnable {
     @CommandLine.Option(names = "-n", description = "The build number", defaultValue = "-1")
     int buildNo;
 
+    @CommandLine.Option(names = "-l", description = "Use legacy retrieval")
+    boolean legacyRetrieval = false;
```
rnc, line 59: I kept the legacy retrieval for now so I could compare the logs being retrieved from the two methods. I guess once we're ok I can remove it? (or maybe in a future iteration)

java-components/cli/src/main/openapi/tekton-results.yaml
```diff
@@ -0,0 +1,764 @@
...
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                # STONEBLD-1568 - Log returns a sequence of json objects as a string not a Log. These need
+                #                 to be separately parsed and then reassembled.
+                # See https://github.com/tektoncd/results/blob/main/docs/large-log-files.md
+                # $ref: "#/components/schemas/Log"
+                type: string
```
rnc, line 300: Only change to the schema. From speaking to the tekton-results devs it can return (unexpectedly depending upon the size of the log) either a Log or, effectively, a series of JSON objects formatted as per the large-log-files reference

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildLogsCommand.java
```diff
@@ -90,88 +109,151 @@ public void run() {
...
+        } else {
 
-                    var finishTime = OffsetDateTime.parse(i.getState().getTerminated().getFinishedAt(), formatter);
-                    Duration duration = Duration.between(startTime, finishTime);
-                    startTime = finishTime;
-                    System.out.println("\n\n#####################################################");
-                    System.out.println(
-                            "### Container " + i.getName() + " finished at " + i.getState().getTerminated().getFinishedAt()
-                                    + " in " + duration.getSeconds() + " seconds");
+            RouteSpec routeSpec = client.routes().inNamespace("openshift-pipelines").withName("tekton-results").get().getSpec();
```
rnc, line 198: Rather than having to have another CLI option to pass in the route, I thought we could just look up the one defined by `tekton_route.yaml`.

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildLogsCommand.java
stuartwdouglas: I would leave both, but use legacy as a fallback (i.e. lookup the results entry, if it is not present attempt to read it from kube).

This was it will still work even if results is not installed on the cluster. At this stage I want results to be an optional dependency rather than a hard requirement.

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildLogsCommand.java
stuartwdouglas: This is a good idea in theory, however for RHTAP the users don't have any permissions on the openshift-pipelines namespace.

Can we do some sensible defaults based on the cluster address?

For RHTAP the results path looks like:

`https://console.redhat.com/api/k8s/plugins/tekton-results/workspaces/sdouglas1/apis/results.tekton.dev`

(this is due to it going through a proxy)

Maybe attempt to lookup the address, if it fails (likely due to permissions), default to the RHTAP prod one?

stuartwdouglas (changes requested):

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildLogsCommand.java
rnc: I've changed to to attempt the routes call and if that fails fall back to assembling a prod url. I've made the 'prod' url customisable through the CLI in case someone is using staging.

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildLogsCommand.java
```diff
@@ -39,10 +61,19 @@ public class BuildLogsCommand implements Runnable {
     @CommandLine.Option(names = "-n", description = "The build number", defaultValue = "-1")
     int buildNo;
 
+    @CommandLine.Option(names = "-l", description = "Use legacy retrieval")
+    boolean legacyRetrieval = false;
+
+    @CommandLine.Option(names = "-u", description = "URL for Tekton-Results")
+    String tektonUrl = "https://console.redhat.com/";
```
stuartwdouglas, line 68: This should not have the trailing slash, you end up with 'https://console.redhat.com/:443' otherwise.

stuartwdouglas (changes requested):

stuartwdouglas:
So I tried testing this out on prod and there were some minor issues with the URI. Once I had fixed those though it was failing with an authentication issue.

I figured the simplest way to work around that was to just use the Kube Clients HTTP Client instead, so I came up with this simple example that can read from prod: https://github.com/redhat-appstudio/jvm-build-service/commit/376186f544b0a561fb191510d1f28418749348b3

What do you think about removing the open API stuff and just using the kube client directly. It's only a single very straightforward call so I don't think the generated code buys us much.

java-components/cli/src/main/java/io/github/redhatappstudio/jvmbuild/cli/builds/BuildLogsCommand.java
```diff
@@ -117,6 +119,9 @@ public void run() {
                 }
                 buildNumbers.add(i);
             }
+            for (int i = 0; i < theBuild.getStatus().getBuildAttempts().size(); ++i) {
+                buildNumbers.add(i);
+            }
```
rnc, line 124: Can you explain this change (compared to the pre-existing loop above). I didn't think I changed this code?

stuartwdouglas: If all the pipeline runs have been deleted then you end up with an empty list.

I probably should have deleted the code above, build attempts should be the definitive record of how many builds have been done.

stuartwdouglas:
This works, I think there are some general improvements we can make to the log viewer (e.g. prompt for the build if none are specified), however this is a separate issue.

PR #858, opened 2023-09-28
Free disk space for minikube

rnc:


.github/workflows/minikube.yaml
```diff
@@ -33,6 +33,15 @@ jobs:
     runs-on: ubuntu-latest
     name: Minikube Test Sample Project
     steps:
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@76866dbe54312617f00798d1762df7f43def6e5c # v1.2.0
```
rnc, line 37: There appears to be a number of actions that do this. Other options are:
* https://github.com/marketplace/actions/maximize-build-disk-space [ does more than just removing software ]
* https://github.com/insightsengineering/disk-space-reclaimer [ fork of `jlumbroso/free-disk-space` that might be more recent ] 

The current configured usage is fairly minimal with only android/dotnet/haskell deletion which gains an extra ~16GB.

stuartwdouglas (approved):

stuartwdouglas:
Can you do this for the other minikube test as well?

PR #887, opened 2023-10-23
Allow automatic generation of AWS CodeArtifact tokens

rnc:


stuartwdouglas:
> @stuartwdouglas Do you want the code from Aphelia that deleted the package before attempting publishing (to avoid https://docs.aws.amazon.com/codeartifact/latest/ug/packages-overview.html#package-publishing-overwrite-assets) ported as well? Or leave that until needed?

I think we will need it at some point, but lets not worry about it right now.

PR #893, opened 2023-10-26
Remove Gabe from owners

stuartwdouglas:
@gabemontero I figure there is not much point in you still being in here? 

Also the approvers section was doubled up for some reason.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,14 +2,7 @@
 
 approvers:
 - stuartwdouglas
-- gabemontero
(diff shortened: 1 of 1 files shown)


gabemontero:
> @gabemontero I figure there is not much point in you still being in here?
> 
> Also the approvers section was doubled up for some reason.

makes sense @stuartwdouglas 

we can always add me back if something changes

thanks

gabemontero (approved):

PR #895, opened 2023-10-27
Add git repo creation

rnc:
@stuartwdouglas There are a few TODO I wanted to discuss with you (and it needs some further tidy up) but the basic functionality of creating a repo appears to work (https://github.com/rnc/commons-net was created from a build I did)

### DIFF
--- a/deploy/base-development.sh
+++ b/deploy/base-development.sh
@@ -17,7 +17,7 @@ fi
 DIR=`dirname $0`
 kubectl apply -f $DIR/namespace.yaml
 kubectl config set-context --current --namespace=test-jvm-namespace
-kubectl delete --ignore-not-found secret jvm-build-image-secrets jvm-build-git-secrets jvm-build-maven-repo-secrets jvm-build-maven-repo-aws-secrets jvm-build-s3-secrets
+kubectl delete --ignore-not-found secret jvm-build-image-secrets jvm-build-git-secrets jvm-build-maven-repo-secrets jvm-build-maven-repo-aws-secrets jvm-build-s3-secrets jvm-build-git-repo-secrets
 
 if [ -n "$QUAY_ORG" ] && [ -n "$QUAY_TOKEN" ]; then
     kubectl delete --ignore-not-found secret  -n image-controller quaytoken
@@ -28,6 +28,9 @@ kubectl create secret generic jvm-build-git-secrets --from-literal .git-credenti
 https://$GITHUB_E2E_ORGANIZATION:$GITHUB_TOKEN@github.com
 https://test:test@gitlab.com
 "
+if [ -n "$GIT_DEPLOY_TOKEN" ]; then
+    kubectl create secret generic jvm-build-git-repo-secrets --from-literal gitdeploytoken="$GIT_DEPLOY_TOKEN"
+fi
 if [ -n "$MAVEN_PASSWORD" ]; then
     kubectl create secret generic jvm-build-maven-repo-secrets --from-literal mavenpassword="$MAVEN_PASSWORD"
 fi
--- a/deploy/crds/base/jvmbuildservice.io_jbsconfigs.yaml
+++ b/deploy/crds/base/jvmbuildservice.io_jbsconfigs.yaml
@@ -87,7 +87,7 @@ spec:
                 type: object
               enableRebuilds:
                 type: boolean
-              gitDeployment:
+              gitSourceArchive:
                 properties:
                   identity:
                     type: string
--- a/deploy/crds/base/jvmbuildservice.io_jbsconfigs.yaml
+++ b/deploy/crds/base/jvmbuildservice.io_jbsconfigs.yaml
@@ -87,6 +87,13 @@ spec:
                 type: object
               enableRebuilds:
                 type: boolean
+              gitDeployment:
+                properties:
+                  identity:
+                    type: string
+                  url:
+                    type: string
+                type: object
               hermeticBuilds:
                 type: string
               host:
--- a/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/DeployCommand.java
+++ b/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/DeployCommand.java
@@ -39,6 +39,7 @@ import com.redhat.hacbs.classfile.tracker.ClassFileTracker;
 import com.redhat.hacbs.classfile.tracker.TrackingData;
 import com.redhat.hacbs.container.analyser.dependencies.SBomGenerator;
 import com.redhat.hacbs.container.analyser.deploy.containerregistry.ContainerRegistryDeployer;
+import com.redhat.hacbs.container.analyser.deploy.git.Git;
 import com.redhat.hacbs.container.analyser.deploy.mavenrepository.MavenRepositoryDeployer;
 import com.redhat.hacbs.container.results.ResultsUpdater;
 import com.redhat.hacbs.recipies.util.FileUtil;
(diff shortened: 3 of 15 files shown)


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/DeployCommand.java
```diff
@@ -380,6 +393,13 @@ public void accept(String s, String hash) {
                     deploymentPath);
             deployer.deploy();
         }
+        if ((isNotEmpty(gitURL) || isNotEmpty(gitIdentity)) && gitToken.isPresent()) {
```
stuartwdouglas, line 396: I wonder if we want to do this first? I think it might be better to have a git branch with no corresponding deployment than a deployment with no source.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/DeployCommand.java
```diff
@@ -380,6 +393,13 @@ public void accept(String s, String hash) {
...
+            var git = Git.builder(gitURL, gitIdentity, gitToken.get());
+            String name = scmUri.endsWith(".git") ? scmUri.substring(0, scmUri.length() - 4) : scmUri;
+            name = name.substring(name.lastIndexOf("/") + 1);
```
stuartwdouglas, line 399: I think we also need to work the rest of the URI into the name, by transforming it into a 'safe' name for a git repo.

e.g. something like github.com/foo/bar would end up as github.com/<user>/foo-bar. 

Possibly we also need to think about what if github and gitlab have the same org/repo combo but different, but that seems unlikely, unless a project has changed from one to the other.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/DeployCommand.java
rnc: Do you mean move it above image/maven repo deployment?

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/DeployCommand.java
rnc: Hmm. Good point, though even if they both have the same combination we should be deploying to a new user/org so it shouldn't matter. But I think the originating user/org would be required as otherwise we might have foo/bar and foo/foo so without the user/org they would clash.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/git/Git.java
```diff
@@ -0,0 +1,88 @@
...
+import io.quarkus.logging.Log;
+
+public abstract class Git {
+
+    protected CredentialsProvider credentialsProvider;
+
+    public abstract void create(String name)
+            throws IOException;
+
+    // TODO: Maybe remove, don't think we need this actually.
```
stuartwdouglas, line 22: I don't think so either.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/DeployCommand.java
rnc: Moved to https://github.com/redhat-appstudio/jvm-build-service/pull/895/files#diff-9b9ad1a14aa7ca87134f689dba59451f601be0ac2e865537bc68c543515bc827R153

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/git/Git.java
```diff
@@ -0,0 +1,102 @@
...
+            StoredConfig jConfig = jRepo.getConfig();
+            Log.infof("Updating current origin of %s to %s", jConfig.getString("remote", "origin", "url"),
+                    httpTransportUrl);
+            jConfig.setString("remote", "origin", "url", httpTransportUrl);
+            jConfig.save();
+            Log.infof("Pushing to %s with content from %s (branch %s, commit %s, tag %s)", httpTransportUrl, path,
+                    jRepo.getBranch(), commit, tagName);
+
+            // TODO: Discuss use of annotated tag?
+            Ref tagRef = jGit.tag().setAnnotated(true).setName(tagName).setForceUpdate(true).call();
```
rnc, line 46: Currently I'm using annotated tags - unless we want to use lightweight?

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/git/Git.java
```diff
@@ -0,0 +1,102 @@
...
+            throws IOException;
+
+    protected void pushRepository(Path path, String httpTransportUrl, String commit) {
+        try (var jGit = org.eclipse.jgit.api.Git.init().setDirectory(path.toFile()).call()) {
+            // TODO: Discuss tag naming - should a rebuild produce the same results ? In which case why use a different tag?
+            //   As JBS might have changed the code? We could append the image-id i.e.
+            //   apache/commons-net.git@rel/commons-net-3.9.0    DependencyBuildStateComplete   75ecd81c7a2b384151c990975eb1dd10
+            //   Tag would be
+            //      rel/commons-net-3.9.0-75ecd81c7a2b384151c990975eb1dd10
+            var tagName = jGit.describe().setTags(true).setTarget(commit).call();
```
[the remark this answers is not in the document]
stuartwdouglas: I think potentially we want two different additional tags, and two different ID concepts to go with it.

I think we want two different IDs:
- A stable build id (e.g. the DependencyBuild name)
- A completely unique UID for the build

We could then create two tags with both IDs appended or prepended. The idea would be that if we ever rebuild something the stable ID would be overwritten, but the unique one would always exist. We could do the same for image deployments, so that even when there are rebuilds we never loose anything.

deploy/base-development.sh
```diff
@@ -17,7 +17,7 @@ fi
 DIR=`dirname $0`
 kubectl apply -f $DIR/namespace.yaml
 kubectl config set-context --current --namespace=test-jvm-namespace
-kubectl delete --ignore-not-found secret jvm-build-image-secrets jvm-build-git-secrets jvm-build-maven-repo-secrets jvm-build-maven-repo-aws-secrets jvm-build-s3-secrets
+kubectl delete --ignore-not-found secret jvm-build-image-secrets jvm-build-git-secrets jvm-build-maven-repo-secrets jvm-build-maven-repo-aws-secrets jvm-build-s3-secrets jvm-build-git-repo-secrets
```
stuartwdouglas, line 20: I think we may have too many secrets :-) At some point we may want to consolidate these, today is not that day however.

deploy/crds/base/jvmbuildservice.io_jbsconfigs.yaml
```diff
@@ -87,6 +87,13 @@ spec:
                 type: object
               enableRebuilds:
                 type: boolean
+              gitDeployment:
```
stuartwdouglas, line 90: Sorry I didn't pick this up before, but this is not really 'deployment'. Maybe gitSourceArchive would be a better name? I know this is minor but the CRDs need to be backwards compatible so changing things later is a pain.

stuartwdouglas (changes requested):

deploy/base-development.sh
rnc: Agreed :-)

PR #915, opened 2023-11-09
Remove quaytoken

stuartwdouglas:
This is now handled by the image controller

### DIFF
--- a/deploy/base-development.sh
+++ b/deploy/base-development.sh
@@ -18,11 +18,6 @@ DIR=`dirname $0`
 kubectl apply -f $DIR/namespace.yaml
 kubectl config set-context --current --namespace=test-jvm-namespace
 
-if [ -n "$QUAY_ORG" ] && [ -n "$QUAY_TOKEN" ]; then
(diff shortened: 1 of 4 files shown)


rnc:
@stuartwdouglas Doesn't https://github.com/redhat-appstudio/jvm-build-service/blob/main/deploy/operator/overlays/ci-template/kustomization.yaml#L10 need changing as well? 

Also I see in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_jvm-build-service/915/pull-ci-redhat-appstudio-jvm-build-service-main-jvm-build-service-in-repo-e2e/1724593677684183040#1:build-log.txt%3A115 that it has the log line `deploy/install-openshift-pipelines.sh: line 15: kubectl: command not found` as if kubectl is not installed?

stuartwdouglas:
I think you are right about the ci-template, no idea about missing kubectl, could be a ci glitch

stuartwdouglas:
The RHTAP e2e is broken ATM due to external changes, I will hold off merging this until it is fixed.

PR #917, opened 2023-11-13
Fix issue with rebuilds

stuartwdouglas:


stuartwdouglas:
@rnc FYI, after the change to lookup existing artifacts it was no longer possible to rebuild things without deleting them from the registry. This should fix that.

PR #923, opened 2023-11-15
STONEBLD-1518: Allow disabling plugins from YAML

dwalluck:


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/maven/MavenPrepareCommand.java
```diff
@@ -87,12 +89,26 @@ private void handleBuild(Path file, boolean topLevel) throws IOException {
         }
     }
 
-    private boolean handlePlugins(List<Plugin> plugins, boolean pluginManagement, boolean topLevel) {
+    private boolean handlePlugins(List<Plugin> plugins, boolean pluginManagement, boolean topLevel) throws IOException {
         boolean modified = false;
+        BuildRecipeInfo info = BuildRecipe.BUILD.getHandler().parse(Path.of(PLUGIN_INFO + "/maven.yaml"));
```
stuartwdouglas, line 94: This is the preprocessor command, you need to use the disabledPlugins command line argument that was passed in, maven.yaml is not present here.

pkg/apis/jvmbuildservice/v1alpha1/dependencybuild_types.go
```diff
@@ -142,6 +142,7 @@ type BuildRecipe struct {
 	AdditionalMemory    int                  `json:"additionalMemory,omitempty"`
 	Repositories        []string             `json:"repositories,omitempty"`
 	AllowedDifferences  []string             `json:"allowedDifferences,omitempty"`
+	DisabledPlugins     string               `json:"disabledPlugins,omitempty"`
```
stuartwdouglas, line 145: This should really be a string array. I know at the moment it is just used as a comma separated list but it is still better to preserve the structure.

stuartwdouglas (changes requested):

java-components/build-request-processor/pom.xml
```diff
@@ -53,6 +53,10 @@
             <groupId>io.quarkiverse.jgit</groupId>
             <artifactId>quarkus-jgit</artifactId>
         </dependency>
+        <dependency>
```
dwalluck, line 56: Not sure why this change was required. IDEA seemed to want it.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/services/RecipeManager.java
```diff
@@ -131,12 +131,28 @@ public List<BuildToolInfo> getBuildToolInfo(String name) {
...
+
+    public List<String> getDisabledPluginInfo(String name) {
+        List<String> results = new ArrayList<>();
+
+        for (var i : recipeDirs) {
+            var path = i.getDisabledPluginInfo(name);
+
+            if (path.isPresent()) {
+                try {
+                    results.addAll(BuildRecipe.BUILD.getHandler().parse(path.get()).getDisabledPlugins());
```
dwalluck, line 150: @stuartwdouglas Is this correct? Merging if multiple recipe dirs...

rnc:
Could we squash to get rid of `Merge remote-tracking branch 'upstream/main' into STONEBLD-1518`  (or use rebase in future?) ?

PR #935, opened 2023-11-23
feat: Add sonarcloud scanning

sawood14012:
Hi, Since JVM is a complex repo the sonar cloud automatic scan would not work and i would like to split it into 2 scans like this and 2 corresponding projects in sonar cloud 

https://issues.redhat.com/browse/STONEBLD-1940 
https://issues.redhat.com/browse/KONFLUX-737


.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: actions/cache@v3
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Build and analyze
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN_JAVA }}
+        run: mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Dsonar.projectKey=redhat-appstudio-1_jvm-build-service-java
```
rnc, line 52: This will end up running tests duplicating (and duplicating) work done in other workflows. Why can't a single workflow be used?

sonar-project.properties
```diff
@@ -0,0 +1,19 @@
...
+
+# Path to sources
+sonar.sources=.
+sonar.exclusions=**/*test*, **/vendor/**, **/zz_generated.*, **/*.yaml, **/java-components/**
+
+# Source encoding
+sonar.sourceEncoding=UTF-8
+
+# Exclusions for copy-paste detection
+sonar.cpd.exclusions=**/*test*
```
rnc, line 19: Is this only for the GoLang scanning? Can this be a hidden file or specified directly in the workflow?

rnc:
Will this cause the run to fail if the sonar notifies? i.e. can it be configured to notify only?

sonar-project.properties
rnc: How come you renamed it to a hidden file in other repositories e.g. https://github.com/redhat-appstudio/e2e-tests/blob/main/.sonarcloud.properties , https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/.sonarcloud.properties ?

PR #938, opened 2023-11-26
Fix issue from a rebase

stuartwdouglas:
It appears a rebase resulting in some code duplication

### DIFF
--- a/pkg/reconciler/dependencybuild/dependencybuild.go
+++ b/pkg/reconciler/dependencybuild/dependencybuild.go
@@ -355,11 +355,9 @@ func (r *ReconcileDependencyBuild) handleAnalyzeBuildPipelineRunReceived(ctx con
 		if len(unmarshalled.Image) > 0 {
 			log.Info(fmt.Sprintf("Found preexisting shared build with deployed GAVs %#v from image %#v", unmarshalled.Gavs, unmarshalled.Image))
 			db.Status.State = v1alpha1.DependencyBuildStateComplete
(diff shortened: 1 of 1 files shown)


pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -670,13 +668,21 @@ func (r *ReconcileDependencyBuild) handleBuildPipelineRunReceived(ctx context.Co
...
+				} else if i.Name == artifactbuild.PipelineResultVerificationResult {
+					verificationResults = i.Value.StringVal
 				} else if i.Name == artifactbuild.PipelineResultGavs {
 					deployed := strings.Split(i.Value.StringVal, ",")
 					db.Status.DeployedArtifacts = deployed
-				} else if i.Name == artifactbuild.PipelineResultVerificationResult {
-					verificationResults = i.Value.StringVal
+				} else if i.Name == artifactbuild.PipelineResultDeployedResources && len(i.Value.StringVal) > 0 {
+					//we need to create 'DeployedArtifact' resources for the objects that were deployed
+					deployed = strings.Split(i.Value.StringVal, ",")
```
rnc, line 678: Why isn't this using `gavs` from above so that is stored in `run.Results` ?

stuartwdouglas: ah, oops, I did not see your question till after I hit the merge button, this is a bug.

rnc: I'll fix it in #944 :-)

rnc: Ah, nm, I see you've entered https://github.com/redhat-appstudio/jvm-build-service/pull/945/

PR #948, opened 2023-11-29
Minikube script updates

rnc:


deploy/minikube-development.sh
```diff
@@ -2,37 +2,42 @@
...
+kubectl apply -k $DIR/crds/base
+#Load missing CRD
+kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.69.1/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
+sleep 2
 
 $DIR/base-development.sh  $1
 
 # base-development.sh switches to the test-jvm-namespace namespace
-kubectl create sa pipeline
+kubectl create --dry-run=client -o=yaml sa pipeline | kubectl apply -f -
```
stuartwdouglas, line 34: Why this change?

rnc: This is similar to prior changes I made in other scripts - this way, if the script is run twice, it is not an error condition and instead simply configuring if there are any changes, ignoring if there are not. (create versus apply).

stuartwdouglas: ah, I missed that had changed.

PR #949, opened 2023-11-29
 Refactor deployment scripts. Migrate secrets to yaml with envsubst 

rnc:
Currently also includes #948.

This is still missing ~documentation updates, cleanups, etc~ and there are other potential changes (e.g. removing use of patch-yaml.sh) but this is enough for a first cut of restructuring proposal.

Summary:

- Added documentation
- Migrate all secrets creation to yaml files combined with envsubst
- Allow for customisation of the worker namespace
- Restructure namespace and pipeline as part of base/kustomization
  - This means a single kustomize build/kubectl apply can be used instead of multiple scripts/applies.

This could allow in future, as well as overlays/development, to have overlays/production as well ; this could then be invoked by an external 'SHA-referenced' script.


openshift-with-appstudio-test/e2e/util.go
```diff
@@ -189,7 +189,7 @@ func commonSetup(t *testing.T, gitCloneUrl string, namespace string) *testArgs {
 		debugAndFailTest(ta, err.Error())
 	}
 
-	mavenYamlPath := filepath.Join(path, "..", "..", "deploy", "base", "maven-v0.2.yaml")
+	mavenYamlPath := filepath.Join(path, "..", "..", "deploy", "base", "config", "maven-v0.2.yaml")
```
stuartwdouglas, line 192: This should actually be moved to hack, it is only used for the tests/examples.

rnc: Are you talking about `gradle-v0.1.yaml` and`maven-v0.2.yaml` (and there is also a config.yaml in that directory currently used to enableRebuilds and set some mavenBaseLocations) ? All of these are configured as well by the dev-template directory to patch the image used - while I see the e2e tests patch the image on about line 209, from what I can see the examples do not patch the image or load the task themselves?

rnc: I've moved the 3 file referenced and modified the exampes to patch the image according to `QUAY_USERNAME`.

PR #1025, opened 2023-12-19
Update to JDK21 builder image

stuartwdouglas:


stuartwdouglas:
I think this is now timing out as it is trying too many combinations. We need to stop old builds from being built with newer JDKs

PR #1033, opened 2023-12-20
Simplify openshift-ci

rnc:


rnc:
@stuartwdouglas Do any of the tests trigger the redhat-appstudio-jvm-build-service-main.yaml (i.e. the openshift-ci) so I can verify any changes to it still work?

deploy/openshift-ci.sh
```diff
@@ -12,9 +12,26 @@ echo ${JVM_BUILD_SERVICE_CACHE_IMAGE}
...
+        if [ $(date +%s) -gt $endTime ]; then
+            echo "Unable to find $1"
+            exit 1
+        fi
+    done
+}
+
 DIR=`dirname $0`
 echo "Running out of ${DIR}"
-$DIR/install-openshift-pipelines.sh
```
stuartwdouglas, line 17: This is also useful when setting up a new hypershift or clusterbot cluster

stuartwdouglas: Actually I guess it does not really matter.

PR #1036, opened 2023-12-21
Allow more flexible S3_SYNC definition

rnc:


deploy/operator/base/deployment.yaml
```diff
@@ -25,6 +25,10 @@ spec:
           args:
             - "--v=4"
             - "--zap-log-level=info"
+          env:
+            - name: S3_SYNC_ENABLED
+              value: |
+                ${JBS_S3_SYNC_ENABLED}
```
stuartwdouglas, line 31: This won't work with argocd, so it will break the infra-deployments repo.

stuartwdouglas (changes requested):

deploy/operator/base/deployment.yaml
rnc: You mean the infra stuff processes that file directly (somehow?) and will see a unknown string instead of it being substituted? As opposed to argocd being unable to process certain yaml formats?

deploy/operator/overlays/dev-template/kustomization.yaml
```diff
@@ -12,8 +12,17 @@ images:
...
+- patch: |-
+    - op: add
+      path: /spec/template/spec/containers/0/env
+      value:
+        - name: S3_SYNC_ENABLED
+          value: |
+            ${JBS_S3_SYNC_ENABLED}
+  target:
+    kind: Deployment
+    name: hacbs-jvm-operator
```
rnc, line 25: Moved the patch here so all operator patches are together (such as the imagePullPolicy below)

deploy/operator/overlays/dev-template/kustomization.yaml
```diff
@@ -12,8 +12,18 @@ images:
...
-
 patches:
+# Note - see base-development for quoting requirements for
+# JBS_S3_SYNC_ENABLED and the linked kubernete issue.
+- patch: |-
+    - op: add
+      path: /spec/template/spec/containers/0/env
+      value:
+        - name: S3_SYNC_ENABLED
+          value: ${JBS_S3_SYNC_ENABLED}
```
rnc, line 23: Ideally patching this as either `"${JBS_S3_SYNC_ENABLED}"` or
```
value: |
	${JBS_S3_SYNC_ENABLED}
```
would transparently patch through but in the former case the quotes are stripped (which is likely https://github.com/kubernetes-sigs/kustomize/issues/4845 ) and in the latter case the inline literal scalar is completely removed and replaced by `value: ${JBS_S3_SYNC_ENABLED}` which actually changed the meaning due to it being then coerced to a bool type

deploy/operator/base/deployment.yaml
stuartwdouglas: Infra deployments does not use our development or CI scripts, it is using Argo CD, so it just doesn't know about the substitution.

rnc: Ok, thats fine. I've removed the env variable

PR #1054, opened 2024-01-02
Error handling and UI layout

rnc:


java-components/management-console/src/main/java/com/redhat/hacbs/management/github/GithubIntegration.java
```diff
@@ -55,10 +58,19 @@ public class GithubIntegration {
...
 
+    @ConfigProperty(name = "kube.disabled", defaultValue = "false")
+    boolean disabled;
+
     public static final String SUPPLY_CHAIN_CHECK = "Supply Chain Check";
 
     @PostConstruct
     public void setupWatch() {
+        if ((LaunchMode.current() == LaunchMode.TEST
+                && !Objects.equals(System.getProperty(Config.KUBERNETES_NAMESPACE_SYSTEM_PROPERTY), "test")) || disabled) {
```
rnc, line 69: While I copied this block from https://github.com/redhat-appstudio/jvm-build-service/blob/main/java-components/management-console/src/main/java/com/redhat/hacbs/management/watcher/BuildImportWatcher.java#L47 I am unclear as to what/how might set `Config.KUBERNETES_NAMESPACE_SYSTEM_PROPERTY`?

stuartwdouglas: This would be set by Quarkus if we were using kubernetes dev service that creates a mock kube API server.

stuartwdouglas (approved):

PR #1064, opened 2024-01-08
Print discovery logs

rnc:


pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -208,7 +208,7 @@ func (r *ReconcileDependencyBuild) handleStateNew(ctx context.Context, log logr.
 		pr.Spec.Workspaces = []pipelinev1beta1.WorkspaceBinding{{Name: "tls", EmptyDir: &v1.EmptyDirVolumeSource{}}}
 	}
 	pr.Namespace = db.Namespace
-	pr.GenerateName = db.Name + "-build-discovery-"
+	pr.Name = db.Name + "-build-discovery"
 	pr.Labels = map[string]string{artifactbuild.PipelineRunLabel: "", artifactbuild.DependencyBuildIdLabel: db.Name, PipelineTypeLabel: PipelineTypeBuildInfo}
```
rnc, line 212: I believe with `db.Name` the task name should be 'unique' anyway and therefore not require the use of `GenerateName`.

rnc:
@stuartwdouglas When you're back, I'm guessing the change to the discovery name must be interfering with the e2e tests somehow

pkg/reconciler/dependencybuild/dependencybuild.go
stuartwdouglas: So this is an issue with the way we resolve 'contamination', where it just runs discovery a second time. We don't really need to do that as we already have the info, but the way the code is now it just kinda restarts the process. Ideally it would just re-use the initial discovery results.

PR #1066, opened 2024-01-08
Support retrieving logs in console like CLI

rnc:
Requires #1064, #1036, #1065, #1055

Allows the console logs to be retrieved from S3 or Tekton or the fallback legacy method.


java-components/management-console/src/main/java/com/redhat/hacbs/management/resources/BuildLogs.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private static final String DEV_PATH = "/apis/results.tekton.dev";
+
+    private static final String PROD_PATH = "/api/k8s/plugins/tekton-results/workspaces/";
+
+    @Inject
+    S3Client s3Client;
+
+    @ConfigProperty(name = "JBS_S3_SYNC_ENABLED", defaultValue = "true")
```
stuartwdouglas, line 41: This should be jbs.s3.sync.enabled, the config system automatically maps upper case env vars.

stuartwdouglas: Also this will mean that by default the CLI will only pull logs from S3 unless you explicitly set this, it would be better if we could be smarter about this.

stuartwdouglas: i.e. try S3 by default, then fall back to the other methods if it fails. Only print out the exception details if none of them work.

stuartwdouglas (changes requested):

java-components/management-console/src/main/java/com/redhat/hacbs/management/resources/BuildLogs.java
rnc: I think its the inverse - it will always attempt to try to pull logs from S3 as it defaults to true? 

However, I do agree that it doesn't do a fallback which the CLI does (partially).

PR #1089, opened 2024-01-19
Use an index for potentialBuildRecipes

rnc:


pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -655,9 +657,12 @@ func (r *ReconcileDependencyBuild) handleBuildPipelineRunReceived(ctx context.Co
...
 
 				if doRetry {
-					existing := db.Status.PotentialBuildRecipes
-					db.Status.PotentialBuildRecipes = []*v1alpha1.BuildRecipe{attempt.Recipe}
-					db.Status.PotentialBuildRecipes = append(db.Status.PotentialBuildRecipes, existing...)
+					log.Info(fmt.Sprintf("Resetting %d to retry build for %s", db.Status.PotentialBuildRecipesIndex,
+						db.Name))
+					db.Status.PotentialBuildRecipesIndex--
+					if db.Status.PotentialBuildRecipesIndex < 0 {
+						db.Status.PotentialBuildRecipesIndex = 0
```
stuartwdouglas, line 664: This does not look right? The index should always be at least one at this point right?

rnc: Yes it should. This was an attempt to get the tests to pass that I fixed in a different way but forgot to remove this.

PR #1090, opened 2024-01-19
Update to OpenShift Pipelines 1.13

rnc:


rnc:
@stuartwdouglas Its not clear to me why the minikube tests are failing. I'm also slightly concerned as to whether this upgrade will work with tekton-results.

PR #1091, opened 2024-01-19
Fix sample Gradle pipeline

dwalluck:


hack/examples/gradle-v0.1.yaml
```diff
@@ -158,7 +158,7 @@ spec:
                     mavenLocal()
                     maven {
                         name "HACBS Maven Repository"
-                        url "$(params.MAVEN_MIRROR_URL)/v2/cache/user/default/"
```
stuartwdouglas, line 161: /v2/cache/user/default/ is the correct URL, we should be adding gradle to the list of repositories in the JBSConfig.

PR #1124, opened 2024-01-23
UI Enhancements

rnc:
Requires #1066 merged first.

### DIFF
--- a/java-components/management-console/src/main/java/com/redhat/hacbs/management/dto/ArtifactDTO.java
+++ b/java-components/management-console/src/main/java/com/redhat/hacbs/management/dto/ArtifactDTO.java
@@ -1,0 +1,37 @@
+package com.redhat.hacbs.management.dto;
+
+import java.util.Objects;
+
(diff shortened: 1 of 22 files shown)


java-components/management-console/src/main/resources/db/migration/V1.0.0__management-console.sql
```diff
@@ -266,7 +265,6 @@
         contaminated boolean not null,
         creationTimestamp timestamp(6) with time zone not null,
         succeeded boolean not null,
-        uid varchar(255) not null,
```
stuartwdouglas, line 269: We can't really do this with flyway without deleting the database, we need to start creating new versions.

I have actually just added something to Quarkus to hopefully make this easier, that can show you the diff between the current DB and the expected state: https://github.com/quarkusio/quarkus/pull/38358

We probably need a better strategy around this at some point. Maybe we should just drop flyway for now and use hibernate migration instead for this stage.

PR #1133, opened 2024-01-23
Validation fixes for renovate

rnc:


rnc:
@stuartwdouglas FYI, I switched to a different branch to see if I could get the renovate job to validate its own config which it was supposed to with a special 'magic branch'. At time of writing, it hasn't seemed to work.

PR #1179, opened 2024-01-29
Scan Maven build files recursively

dwalluck:


stuartwdouglas:
The test failures look related.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommand.java
```diff
@@ -370,9 +377,106 @@ private boolean searchForBuildScript(BuildRecipeInfo buildRecipeInfo, CacheBuild
...
-            Log.infof("Found Maven pom file at %s", pomFile);
+            paths.add(pomFile);
+        }
+
+        GradleUtils.getGradleBuild(path).ifPresent(paths::add);
+
+        var buildSbt = path.resolve("build.sbt");
+
+        if (Files.exists(buildSbt)) {
+            paths.add(buildSbt);
```
stuartwdouglas, line 390: Because this is adding the files themselves if you have a subdirectory with both an ant and maven script this will now fail when it would have passed before.

stuartwdouglas (changes requested):

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommand.java
dwalluck: In that context, it's still checking only the subdirectory size `potentialSubDir`, not the number of build scripts:

```java
                    for (Path potentialSubDir : stream) {
                        if (Files.isDirectory(potentialSubDir)) {
                            buildScriptPaths = searchForBuildScripts(buildRecipeInfo, potentialSubDir);
                            if (!buildScriptPaths.isEmpty()) {
                                paths.add(potentialSubDir);
                            }
                        }
```

PR #1193, opened 2024-01-30
Use a random name for test remote to avoid clashes

rnc:


java-components/build-request-processor/src/test/java/com/redhat/hacbs/container/analyser/deploy/git/GitTest.java
```diff
@@ -68,11 +70,13 @@ public void testPush()
...
                 // Don't have the tag and cannot guarantee a fork will have it so fetch from primary repository.
+                String newRemote = RandomStringUtils.randomAlphabetic(8);
+                Log.infof("Current repo does not have 0.1 tag; configuring %s to fetch it.", newRemote);
                 initialRepository.remoteAdd()
-                        .setUri(new URIish("https://github.com/redhat-appstudio/jvm-build-service.git")).setName("upstream")
+                        .setUri(new URIish("https://github.com/redhat-appstudio/jvm-build-service.git")).setName(newRemote)
                         .call();
                 initialRepository.fetch().setRefSpecs("refs/tags/0.1:refs/tags/0.1").setTagOpt(TagOpt.NO_TAGS)
-                        .setRemote("upstream").call();
+                        .setRemote(newRemote).call();
```
rnc, line 79: Annoyingly Eclipse silently hides the problem if there is an existing remote with same name - hence using a random string. We only need it to retrieve the tag anyway.

dwalluck: How about adding something like this just in case even though we're using random:

```java
var existingRemotes = config.getSubsections("remote");
var newRemote = randomAlphanumeric(8);
assertThat(existingRemotes).doesNotContain(remote);
```

PR #1195, opened 2024-01-31
Fix pushing gitlab repos

stuartwdouglas:


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/deploy/git/Git.java
```diff
@@ -98,7 +98,7 @@ protected GitStatus pushRepository(Path path, String httpTransportUrl, String co
                     .setCredentialsProvider(credentialsProvider);
             // If we're using GitHub, and it is a new repository we should push
             // the default branch else it doesn't show the code
```
rnc, line 100: Pedantic but the comment should be updated.

rnc (approved):

PR #1227, opened 2024-02-06
Verification failure fixes

stuartwdouglas:


java-components/build-request-processor/src/test/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommandTest.java
```diff
@@ -120,6 +121,7 @@ public void testModelResolver1()
     }
 
     @Test
+    @Ignore
```
rnc, line 124: Why is this test not working and being excluded now?

stuartwdouglas: Initially I disabled recursive scanning, but then figured out a way to do it without incurring the memory cost. I have re-enabled it.

PR #1237, opened 2024-02-08
Discover Java version from maven-compiler-plugin

dwalluck:


dwalluck:
This PR adds the ability to discover the Java version from the source/target in maven-compiler-plugin configuration in the case that the properties are undefined.

However, when the Java version is not defined in the POM, it will fall back to the commit date. The commit date logic is different from the version discovery logic and not changed in this PR.

dwalluck:
For example, for a commit date prior to 2014, it will choose JDK 7 only. But JDK 8 still still supports a subset of JDK 7 targets.

I think JDK 7 is [1.4, 1.7] and JDK 8 is [1.5, 1.8] and JDK 11 is [8, 11]

PR #1249, opened 2024-02-10
Fail verification on change in class version or removal of classes

dwalluck:


stuartwdouglas:
This needs https://github.com/redhat-appstudio/jvm-build-service/pull/1267 to make the minikube tests pass

PR #1277, opened 2024-02-13
Don't verify empty parameters

dwalluck:
This is a more general fix for the `Enum.valueOf()` issue where we have a parameter with `name = null` and `access = 0` as opposed to none whatsoever.


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/verifier/asm/MethodInfo.java
```diff
@@ -36,9 +37,9 @@ public MethodInfo(MethodNode node) {
                 node.name, node.desc, node.signature,
                 List.copyOf(node.exceptions),
                 node.parameters != null
-                        ? node.parameters.stream()
+                        ? node.parameters.stream().filter(p -> p.name == null && p.access == 0)
                                 .collect(Collectors.toMap(n -> n.name, ParameterInfo::new, (x, y) -> x, LinkedHashMap::new))
-                        : null,
+                        : Collections.emptyMap(),
```
dwalluck, line 42: Normally, we'd distinguish between `null` and empty map, but the `collect()` always happens, even if we've filtered out all elements.

PR #1284, opened 2024-02-15
Work for Jenkins demo and minor code changes

rnc:
Associated demo PR is https://github.com/rnc/jbs-demo/pull/3

### DIFF
--- a/java-components/management-console/src/main/java/com/redhat/hacbs/management/model/ContainerImage.java
+++ b/java-components/management-console/src/main/java/com/redhat/hacbs/management/model/ContainerImage.java
@@ -15,6 +15,7 @@ import jakarta.persistence.Transient;
 import jakarta.persistence.UniqueConstraint;
 
 import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import io.quarkus.logging.Log;
 import io.quarkus.panache.common.Parameters;
 
 @Entity()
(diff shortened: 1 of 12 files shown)


java-components/management-console/src/main/java/com/redhat/hacbs/management/model/DependencySet.java
```diff
@@ -25,6 +27,7 @@ public class DependencySet extends PanacheEntity {
     @Column(nullable = false)
     public String type;
 
+    @JsonManagedReference
     @OneToMany(fetch = FetchType.EAGER, mappedBy = "dependencySet", cascade = CascadeType.ALL)
```
rnc, line 31: @stuartwdouglas I'm not entirely sure I've got the correct json annotations here (and in IdentifiedDependency below)

java-components/management-console/src/main/java/com/redhat/hacbs/management/model/ContainerImage.java
```diff
@@ -58,7 +60,45 @@ public static ContainerImage findImage(String image) {
         return containerImage;
     }
 
+    public static ContainerImage get(String image) {
+        return get(null, image);
+    }
+
+    public static ContainerImage get(ContainerImageRepository repository, String image) {
```
rnc, line 67: @stuartwdouglas Might be worth speaking about the logic for image lookups and what formats should be acceptable.

stuartwdouglas: In the DB you really need the digest, unless you are happy to get multiple results back. Tags can change so digest is the only really unique way to identify the image.

java-components/management-console/src/main/java/com/redhat/hacbs/management/model/DependencySet.java
stuartwdouglas: Why do you need this, is there something in ImageDTO that is not needed?

In general I am not a fan of using database backed types in JSON endpoints. It is really easy to cause security issues as Jackson can follow relationships and include more data than you expect.

It can also be possible to do the opposite, and update different entities to what you expect if a POST includes updates to related entities.

java-components/management-console/src/main/java/com/redhat/hacbs/management/model/ContainerImage.java
rnc: Yeah. Its not obvious on addImage in the UI what formats are acceptable. Also, if I'm using jib to push an image/tag to Quay I don't know the digest (not sure if I can get jib to tell me the digest of the image it just pushed)

java-components/management-console/src/main/java/com/redhat/hacbs/management/model/DependencySet.java
rnc: Yeah looks like ImageDTO with `analysisFailed` and then another call of `getDependencySet` to get `DependencySetDTO`

java-components/management-console/src/main/java/com/redhat/hacbs/management/model/ContainerImage.java
stuartwdouglas: You can get the digest of the image, it is included in the pipeline results.

PR #1328, opened 2024-02-26
Remove use of GME

rnc:
While this has worked on the test builds I have done I suspect there may be failures if some plugins are not being handled. Equally it should fix some of the failures we are seeing though.

Related : https://github.com/redhat-appstudio/jvm-build-service-builder-images/pull/53

### DIFF
--- a/pkg/reconciler/dependencybuild/scripts/gradle-build.sh
+++ b/pkg/reconciler/dependencybuild/scripts/gradle-build.sh
@@ -10,8 +10,6 @@ cp -r /maven-artifacts/.m2/* "$HOME/.m2/" || true
 cat > "${GRADLE_USER_HOME}"/gradle.properties << EOF
 org.gradle.caching=false
 org.gradle.console=plain
-# This prevents the daemon from running (which is unnecessary in one-off builds) and increases the memory allocation
-org.gradle.daemon=false
 # For Spring/Nebula Release Plugins
 release.useLastTag=true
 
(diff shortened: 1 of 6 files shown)


pkg/reconciler/dependencybuild/scripts/gradle-build.sh
```diff
@@ -10,8 +10,6 @@ cp -r /maven-artifacts/.m2/* "$HOME/.m2/" || true
 cat > "${GRADLE_USER_HOME}"/gradle.properties << EOF
 org.gradle.caching=false
 org.gradle.console=plain
-# This prevents the daemon from running (which is unnecessary in one-off builds) and increases the memory allocation
-org.gradle.daemon=false
```
rnc, line 14: Gradle recommendation is to keep the daemon running now ( e.g. https://docs.gradle.org/current/userguide/gradle_daemon.html#continuous_integration and this is also what Travis CI does).

pkg/reconciler/dependencybuild/scripts/gradle-build.sh
```diff
@@ -116,13 +105,11 @@ echo "INIT SCRIPTS: $INIT_SCRIPTS"
...
 rm -f gradle/verification-metadata.xml
 
-gradle-manipulator $INIT_SCRIPTS -DAProxDeployUrl=file:$(workspaces.source.path)/artifacts --no-colour --info --stacktrace -l "${GRADLE_HOME}" -DdependencySource=NONE -DignoreUnresolvableDependencies=true -DpluginRemoval=ALL -DversionModification=false "${ADDITIONAL_ARGS}" generateAlignmentMetadata  | tee $(workspaces.source.path)/logs/gme.log
-
 echo "Running Gradle command with arguments: $@"
 if [ ! -d $(workspaces.source.path)/source ]; then
   cp -r $(workspaces.source.path)/workspace $(workspaces.source.path)/source
 fi
-gradle $INIT_SCRIPTS -DAProxDeployUrl=file:$(workspaces.source.path)/artifacts --info --stacktrace -Prelease.version=$(params.ENFORCE_VERSION) "$@"  | tee $(workspaces.source.path)/logs/gradle.log
+gradle $INIT_SCRIPTS -Dmaven.repo.local=$(workspaces.source.path)/artifacts --info --stacktrace "$@"  | tee $(workspaces.source.path)/logs/gradle.log
```
rnc, line 112: I think that with the tagging and use of `useLastTag` then `-Prelease.version` should not be required.

PR #1337, opened 2024-02-28
Add copy artifacts command

dwalluck:


stuartwdouglas:
The command needs to be added to https://github.com/stuartwdouglas/jvm-build-service/blob/89ec977c3886ef61bfeccf695d2376622fc82d6f/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/EntryPoint.java#L20

PR #1355, opened 2024-03-01
Make regexp more accurate for release property to avoid conflict with nebula

rnc:
Requires #1353 merged first.


stuartwdouglas:
Failures look related

PR #1378, opened 2024-03-06
Fix hermetic gradle builds

stuartwdouglas:


deploy/overlays/dev-template/config.yaml
```diff
@@ -19,3 +19,4 @@ spec:
     url: ${GIT_DEPLOY_URL}
     disableSSLVerification: ${GIT_DISABLE_SSL_VERIFICATION}
   requireArtifactVerification: true
+  hermeticBuilds: Required
```
rnc, line 22: Given its eventually evaluated to a boolean in createPipelineSpec would it be better to make it a boolean here and in jbsconfig_types?

PR #1386, opened 2024-03-08
Add filtering by build tool

rnc:


java-components/management-console/src/main/java/com/redhat/hacbs/management/resources/BuildHistoryResource.java
```diff
@@ -100,10 +102,25 @@ public PageParameters<BuildListDTO> all(@QueryParam("page") int page, @QueryPara
...
                     .map(BuildAttemptDTO::of)
                     .orElse(null);
-            ret.add(new BuildListDTO(build.id, build.buildIdentifier.dependencyBuildName, build.buildIdentifier.repository.url,
-                    build.buildIdentifier.tag, build.succeeded, build.contaminated,
-                    successfulBuild == null || successfulBuild.passedVerification(), inQueue,
-                    build.creationTimestamp.toEpochMilli()));
+
+            if (tool != null && !tool.isEmpty()) {
+                Optional<BuildAttempt> buildAttempt = build.buildAttempts.stream().findFirst();
+                if (buildAttempt.isPresent() && buildAttempt.get().tool.equals(tool)) {
```
stuartwdouglas, line 108: This won't always work if multiple tools are present (e.g. when there are multiple build files and we try both).

Also this breaks pagination (i.e. you will get less results that what you have selected).

That said we can merge this and improve it later if need be, as this is just for internal use anyway.

rnc: Agreed; I can push up a small improvement to check all buildAttempts which would be a small improvement.

PR #1404, opened 2024-03-13
Fix toolEnv substitution with PipelineParamEnforceVersion

rnc:


pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -104,7 +104,7 @@ func createPipelineSpec(tool string, commitTime int64, jbsConfig *v1alpha12.JBSC
 	toolEnv = append(toolEnv, v1.EnvVar{Name: PipelineParamToolVersion, Value: recipe.ToolVersion})
 	toolEnv = append(toolEnv, v1.EnvVar{Name: PipelineParamProjectVersion, Value: db.Spec.Version})
 	toolEnv = append(toolEnv, v1.EnvVar{Name: JavaHome, Value: javaHome})
-	toolEnv = append(toolEnv, v1.EnvVar{Name: PipelineParamEnforceVersion, Value: "$(params." + PipelineParamEnforceVersion + ")"})
```
stuartwdouglas, line 107: Unfortunately we are probably going to need to go the opposite way at some point, and try and make the task definitions more static with everything passed in as params

PR #1406, opened 2024-03-13
Prefer assemble to build

rnc:
Build will also run tests which we don't need. 

In some cases (graphql) build versus assemble versus shadowJar combined with publishToMavenLocal depending upon the order can lead to unexpected failures.


rnc:
I think this is failing as it needs #1405 to be merged first.

PR #1462, opened 2024-03-26
Remove CPU limits

rnc:
Fixes https://github.com/apheleia-project/jbs-issues/issues/12

### DIFF
--- a/deploy/overlays/dev-template/config.yaml
+++ b/deploy/overlays/dev-template/config.yaml
@@ -19,3 +19,7 @@ spec:
     url: ${GIT_DEPLOY_URL}
     disableSSLVerification: ${GIT_DISABLE_SSL_VERIFICATION}
   requireArtifactVerification: true
+  buildSettings:
+    buildRequestCPU: "2"
+    taskRequestCPU: "300m"
+    taskLimitCPU: "1"


java-components/resource-model/src/main/resources/crds/jvmbuildservice.io_jbsconfigs.yaml
```diff
@@ -45,23 +45,13 @@ spec:
                 type: array
               buildSettings:
                 properties:
-                  buildRequestCPU:
```
stuartwdouglas, line 48: We can't do this for backward compatibility reasons (or at least I would need to make sure that these are not configured anywhere important). I think we should leave the ability to configure limits, but just leave it empty if unconfigured.

rnc: Ok. Do you mean change the code around `resource.ParseQuantity(settingOrDefault(jbsConfig.Spec.BuildSettings.TaskRequestCPU, "10m"))` so that if it hasn't been configured don't configure it (and don't set a default) ? 

From reading further on this, I'm thinking we should actually set a requests value (perhaps to `1`) but no limits value.

rnc: Hmm. Setting only requests gets
`Invalid request resource value: 500m must be less or equal to limit 0`
and the limit has been implicitly set to zero. Furthermore in the build step the build request value is copied over to the limits value:
```
          steps:
            - timeout: 6h0m0s
              resources:
                limits:
                  cpu: 300m
                  memory: 5Gi
                requests:
                  cpu: 300m
                  memory: 5Gi
              script: /original-content/build.sh "$@"
```

rnc: Looks like it was our code taking the defaultRequestsLimit and making that the hard limit.

rnc: @stuartwdouglas Should we add a separate ticket for `We can't do this for backward compatibility reasons (or at least I would need to make sure that these are not configured anywhere important). ` so we can deprecate and remove these?

stuartwdouglas: I don't think it is bad to leave this as configurable, there may be times when it is what we people want.

stuartwdouglas: Actually this PR means they now do nothing, so they should definitively be deprecated. I don't think it is worth worrying about wiring them up as an optional limit.

stuartwdouglas:
I think we might actually need CPU limits for the e2e tests to continue to work. I am seeing the limit being set to 2 by the LimitRange plugin anyway. 

Maybe we should just abandon this and set it to a much higher value in our dev scripts / the production cluster, while leaving it small for the e2e tests.

rnc:
> I think we might actually need CPU limits for the e2e tests to continue to work. I am seeing the limit being set to 2 by the LimitRange plugin anyway.
> 
> Maybe we should just abandon this and set it to a much higher value in our dev scripts / the production cluster, while leaving it small for the e2e tests.

I've tried experimentally changing this in the `dev-tempate/config.yaml` instead (I need to squash this PR).

I'm not sure what the limits are with minikube and whether this will exceed them.

I have noticed that the buildRequestCPU is used both as a request value and a limit value - was that on purpose?

PR #1466, opened 2024-03-27
Fix variable substitution within diagnostic files

rnc:


rnc:
@stuartwdouglas Can we hold merging this until https://github.com/apheleia-project/jbs-issues/issues/13 is resolved.

stuartwdouglas:
I think we should move to disallowing any tekton substitution in these, and have a pre-approved / documented list of env vars instead.

rnc:
Closing this as redundant with https://github.com/apheleia-project/jbs-build-issues/issues/13

PR #1481, opened 2024-04-02
Fix issues related to Maven GAV/path conversion

dwalluck:


stuartwdouglas:
The failure looks related

PR #1610, opened 2024-04-24
Create a dependencyBuild directly

rnc:


pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -1054,6 +1058,27 @@ func (r *ReconcileDependencyBuild) createRebuiltArtifacts(ctx context.Context, l
...
+func (r *ReconcileDependencyBuild) createArtifacts(ctx context.Context, log logr.Logger, pr *tektonpipeline.PipelineRun, db *v1alpha1.DependencyBuild, deployed []string) error {
+	db.Status.DeployedArtifacts = deployed
+
+	for _, i := range deployed {
+		ab := v1alpha1.ArtifactBuild{}
+		ab.Namespace = pr.Namespace
+		ab.Name = artifactbuild.CreateABRName(i)
+		ab.Annotations = map[string]string{artifactbuild.DependencyAnnotation: db.Name}
+		ab.Spec.GAV = i
+		if err := controllerutil.SetOwnerReference(&ab, db, r.scheme); err != nil {
```
rnc, line 1070: After further testing this isn't correct I think. It needs to be `controllerutil.SetOwnerReference(db, &ab, r.scheme)` but I don't think we can have it in this function as the caller calls patchStatus.

rnc: I am trying an experimental change to update it elsewhere but currently hitting a mixture of rbac problems and 'object has been modified' ....

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -788,6 +788,10 @@ func (r *ReconcileDependencyBuild) handleBuildPipelineRunReceived(ctx context.Co
 			if err != nil {
 				return reconcile.Result{}, err
 			}
+			err = r.createArtifacts(ctx, log, pr, db, deployed)
```
stuartwdouglas, line 791: I don't think we need to create `ArtifactBuild` objects for this case, we already have `RebuiltArtifact` as a marker that the objects were created.

java-components/cache/src/main/java/com/redhat/hacbs/artifactcache/artifactwatch/ScmLookup.java
```diff
@@ -80,27 +82,72 @@ public void onAdd(ArtifactBuild newObj) {
...
-                                scm.setScmURL(result.getRepoInfo().getUri());
-                                scm.setCommitHash(result.getHash());
-                                scm.setPath(result.getRepoInfo().getPath());
-                                scm.set_private(result.getRepoInfo().isPrivateRepo());
-                                scm.setTag(result.getTag());
-                                Log.infof("Adding artifactBuild with GAV %s with URI %s and hash %s ",
-                                        newObj.getSpec().getGav(), result.getRepoInfo().getUri(), result.getHash());
+
+                                // If the DependencyBuild was created directly no need to look up the GAV source, instead gather
+                                // from existing dependency.
```
stuartwdouglas, line 88: Like I mentioned below I don't think we need to handle this case.

stuartwdouglas (changes requested):
I think that in addition to the CLI changes we also need a way to rebuild the DependencyBuild objects, at the moment this is handled via annotations on the ArtifactBuild, but if there is not corresponding ArtifactBuild then we can't directly rebuild them.

PR #1632, opened 2024-04-30
Build CPU limits

stuartwdouglas:


pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -607,6 +608,10 @@ func memoryLimits(jbsConfig *v1alpha1.JBSConfig, additionalMemory int) (*memLimi
 	if err != nil {
 		return nil, err
 	}
+	limits.buildLimitCPU, err = resource.ParseQuantity(settingOrDefault(jbsConfig.Spec.BuildSettings.BuildLimitCPU, "2"))
```
rnc, line 611: Was this meant to be different to what is configured in the dev-template/config.yaml?

deploy/overlays/dev-template/config.yaml
```diff
@@ -20,6 +20,7 @@ spec:
     disableSSLVerification: ${GIT_DISABLE_SSL_VERIFICATION}
   requireArtifactVerification: true
   buildSettings:
-    buildRequestCPU: "2"
+    buildRequestCPU: "300m"
+    buildLimitCPU: "4"
```
rnc, line 24: Do our dev clusters have 4 capability (especially with other things running)?

stuartwdouglas: It should not matter for the limit, our dev cluster have 4 cores so this will let it use as much as possible. Even if other things are running it won't matter.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
stuartwdouglas: I don't know what is a reasonable default, I just went a bit more conservative than the dev script.

PR #1634, opened 2024-04-30
Add a new stage to push preprocess and JBS Containerfile changes to git

rnc:
Needs #1580 and #1610

### DIFF
--- a/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/DeployCommand.java
+++ b/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/DeployCommand.java
@@ -23,7 +23,6 @@ import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import jakarta.enterprise.inject.spi.BeanManager;
-import jakarta.inject.Inject;
 
 import org.apache.commons.lang3.StringUtils;
 import org.cyclonedx.BomGeneratorFactory;
@@ -40,7 +39,6 @@ import com.redhat.hacbs.recipes.util.FileUtil;
 import com.redhat.hacbs.resources.model.v1alpha1.dependencybuildstatus.Contaminates;
 import com.redhat.hacbs.resources.util.HashUtil;
 
-import io.quarkus.bootstrap.resolver.maven.BootstrapMavenContext;
 import io.quarkus.logging.Log;
 import picocli.CommandLine;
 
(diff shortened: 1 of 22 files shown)


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/KonfluxCommand.java
```diff
@@ -0,0 +1,91 @@
...
+    public void run() {
+        try {
+            // Konflux files should be in the source code in <root>/.jbs/Containerfile|run-build.sh
+            Log.warnf("Source list files %s", sourcePath.toFile().listFiles());
+            Log.warnf("#### DOCKER:\n" + Files.readString(Path.of(sourcePath.toString(), ".jbs/Containerfile")));
+
+            Git.GitStatus archivedSourceTags = new Git.GitStatus();
+
+            // Save the source first regardless of deployment checks
+            if (isNotEmpty(gitIdentity) && gitToken.isPresent()) {
```
rnc, line 72: If git information is not configured the stage becomes a no-op. This is akin to where the source is saved within the DeployCommand ( https://github.com/redhat-appstudio/jvm-build-service/blob/main/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/DeployCommand.java#L151 ) - the difference is the latter will also include any recipe build script post/pre changes.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/DeployCommand.java
```diff
@@ -123,6 +120,9 @@ public class DeployCommand implements Runnable {
     @CommandLine.Option(names = "--git-disable-ssl-verification")
     boolean gitDisableSSLVerification;
 
+    @CommandLine.Option(names = "--git-reuse-repository")
```
stuartwdouglas, line 123: Why is this nessesary? The repo is already re-used if it exists?

rnc: Yes exactly. If the dependencybuild is created directly there is an option in the CLI to reuse the existing scm to push back changes i.e. if we're using a midstream instead of a pristine upstream. This is then cascaded through to here so that not only the prebuild-source task can push back to midstream but so can deploy (which currently runs the pre/post build scripts).

PR #1691, opened 2024-05-09
Gradle & EnforceVersion

rnc:
Fixes https://github.com/apheleia-project/jbs-build-issues/issues/5

This resolves building btf and changes some Gradle build behaviour - it will only create a tag if enforceVersion is set now (as I don't believe the other route is required). Finally, while I would have preferred to change enforceVersion to a bool throughout, I realise that could affect the crds and backwards compatibility so I've just changed it look like a bool and be consistent from the jvm-build-data source.

### DIFF
--- a/pkg/reconciler/dependencybuild/scripts/gradle-build.sh
+++ b/pkg/reconciler/dependencybuild/scripts/gradle-build.sh
@@ -27,6 +27,10 @@ RELEASE_REPOSITORY_URL=file:$(workspaces.source.path)/artifacts
 RELEASE_SIGNING_ENABLED=false
 mavenCentralUsername=
 mavenCentralPassword=
+
+# Default values for common enforced properties
+sonatypeUsername=jbs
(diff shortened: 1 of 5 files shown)


pkg/reconciler/dependencybuild/scripts/gradle-build.sh
```diff
@@ -27,6 +27,10 @@ RELEASE_REPOSITORY_URL=file:$(workspaces.source.path)/artifacts
 RELEASE_SIGNING_ENABLED=false
 mavenCentralUsername=
 mavenCentralPassword=
+
+# Default values for common enforced properties
+sonatypeUsername=jbs
+sonatypePassword=jbs
```
rnc, line 33: As I've seen several projects including https://github.com/fge/btf/blob/1.2/build.gradle#L130 have checks for these variables it seems simpler to dummy up some values.

PR #1714, opened 2024-05-15
Handle Nebula and StageVoteRelease plugins

rnc:
~~Ideally we would be able to encapsulate this within an init script and only activate it if the Nebula plugin is being used ; however I couldn't find any way of setting release.version there as opposed to from the cmd line.~~

This is for micrometer and pgjdbc builds.

### DIFF
--- a/pkg/reconciler/dependencybuild/scripts/gradle-build.sh
+++ b/pkg/reconciler/dependencybuild/scripts/gradle-build.sh
@@ -9,9 +9,6 @@ cp -r /maven-artifacts/.m2/* "$HOME/.m2/" || true
 
 cat > "${GRADLE_USER_HOME}"/gradle.properties << EOF
 org.gradle.console=plain
-# For Spring/Nebula Release Plugins
-release.useLastTag=true
-release.stage=final
 
 # For https://github.com/Kotlin/kotlinx.team.infra
 versionSuffix=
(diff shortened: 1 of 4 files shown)


pkg/reconciler/dependencybuild/scripts/gradle-build.sh
```diff
@@ -69,7 +66,8 @@ echo "Running Gradle command with arguments: $@"
 if [ ! -d $(workspaces.source.path)/source ]; then
   cp -r $(workspaces.source.path)/workspace $(workspaces.source.path)/source
 fi
-gradle -Dmaven.repo.local=$(workspaces.source.path)/artifacts --info --stacktrace "$@"  | tee $(workspaces.source.path)/logs/gradle.log
+# release.version is only for Nebula Release plugin.
+gradle -Dmaven.repo.local=$(workspaces.source.path)/artifacts --info --stacktrace -Prelease.version=$(params.PROJECT_VERSION) "$@" | tee $(workspaces.source.path)/logs/gradle.log
```
stuartwdouglas, line 70: Can this go into build discovery instead so it ends up in the recipe? Also in general we should be referencing the env var version of PROJECT_VERSION rather than using the params. We should try and remove all copies of $(params) from the scripts and replace them with env vars at some point.

stuartwdouglas (changes requested):

PR #1782, opened 2024-05-30
Skip spotless

rnc:
For https://github.com/apache/maven-resolver.git@maven-resolver-1.9.20 which inherits https://github.com/apache/maven-parent/blob/maven-parent-42/pom.xml which applies spotless to everything.


rnc:
If this is merged we should revert https://github.com/redhat-appstudio/jvm-build-data/pull/208

PR #1783, opened 2024-05-30
Use maven-release plugin API to rewrite poms

rnc:
@stuartwdouglas I noticed when investigating another issue that the MavenPrepareCommand used MavenXpp3Writer to write the poms back. The problem is that doesn't maintain comments (or order). I have tried switching to use the underlying code from the Maven Release Plugin (as it too has to rewrite poms). This is essentially what PME does with some smalls changes (as we're working with just Models and not parsed projects). I'd be open to alternate methods to rewrite the poms if you know of any.


stuartwdouglas:
This looks like a much better approach.

PR #1823, opened 2024-06-13
Update image controller

stuartwdouglas:


rnc:
Does this replace https://github.com/redhat-appstudio/jvm-build-service/pull/1761/files ? Can the baseline go be updated to 1.21?

stuartwdouglas:
Actually I think https://github.com/redhat-appstudio/jvm-build-service/pull/1761 can be merged instead, there is no reason why we can't update the baseline AFAIK.

stuartwdouglas:
Actually there is a problem with the e2e tests, something there is still using an old version of go.

rnc:
> Actually there is a problem with the e2e tests, something there is still using an old version of go.

I suspect that is what is causing all the other PRs to fail in their e2e (as they all want to update the baseline go)

stuartwdouglas:
https://github.com/openshift/release/pull/53327 might fix it

PR #1824, opened 2024-06-13
Switch to using OCI Artifacts via build-trusted-artifacts rather than Jib

rnc:
As https://github.com/GoogleContainerTools/jib doesn't preserve symbolic links this switches to https://github.com/konflux-ci/build-trusted-artifacts/ which is based upon https://oras.land/

Requires https://github.com/redhat-appstudio/jvm-build-service/pull/1842 merged first.


pkg/reconciler/dependencybuild/scripts/Dockerfile.build-trusted-artifacts
```diff
@@ -0,0 +1,3 @@
+#This file is used to enable renovate to update the digests. It is updated by renovate then substituted into the golang code.
+
+FROM quay.io/redhat-user-workloads/rhtap-build-tenant/trusted-artifacts/trusted-artifacts@sha256:9e2ffee0cb28f8a0ed7895a357c73f006005b26ef143f00df067f090282e8cbd
```
stuartwdouglas, line 3: This should really be provided by config, rather than being baked into the binary. This is fine for now, but we should look at this later, as we will need a general solution for Konflux tasks we are referencing.

rnc: Sure. I did consider adding to e.g. system-config.yaml. I'm not actually renovate will catch this (but equally I know we have to update the builder images manually which is not ideal). I don't know how Konflux keeps the references up to date.

stuartwdouglas: I think we can configure renovate to handle this. Konflux runs renovate manually to update references, its run by the build service as a scheduled task.

stuartwdouglas:
The minikube tests are failing with:

```
Creating pre-build-image archive
jq: error: Could not open file /root/.docker/config.json: No such file or directory
jq: error (at <stdin>:1): null (null) and null (null) cannot be multiplied
```

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/DeployCommand.java
```diff
@@ -1,108 +1,52 @@
...
+import com.redhat.hacbs.container.deploy.mavenrepository.CodeArtifactRepository;
+import com.redhat.hacbs.container.deploy.mavenrepository.MavenRepositoryDeployer;
 
+import io.quarkus.bootstrap.resolver.maven.BootstrapMavenContext;
 import io.quarkus.logging.Log;
 import picocli.CommandLine;
 
-@SuppressWarnings("OptionalUsedAsFieldOrParameterType")
 @CommandLine.Command(name = "deploy")
 public class DeployCommand implements Runnable {
```
rnc, line 32: The git diff isn't showing this very well. I renamed MavenDeployCommand to DeployCommand and the original DeployCommand to VerifyCommand.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -338,22 +356,26 @@ func createPipelineSpec(log logr.Logger, tool string, commitTime int64, jbsConfi
 			{Name: PipelineResultImageDigest},
 			{Name: PipelineResultPassedVerification},
 			{Name: PipelineResultVerificationResult},
-			{Name: PipelineResultGitArchive},
+			// TODO: ### DeployPreBuildSource and Deploy push to git. Currently the former is used for GitArchive results.
+			//			{Name: PipelineResultGitArchive},
```
rnc, line 360: Previously `PipelineResultGitArchive` was storing the result of Deploy's git push while DeployPreBuildSource git push was 'lost'. _Currently_ `PipelineResultGitArchive` is now storing the result of `DeployPreBuildSource`. I haven't (yet) added another pipeline result for the Deploy pipeline to also store the result of its git archiving.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -786,87 +871,65 @@ func imageRegistryCommands(imageId string, db *v1alpha1.DependencyBuild, jbsConf
...
+	// 'else' :
+	// No need to pass a default port (of 443) and its not supported by select-oci-auth.sh according
+	// to the tests in https://github.com/konflux-ci/build-trusted-artifacts/pull/103
+	registryArgs.WriteString("/")
 	if imageRegistry.Owner != "" {
-		registryArgs = append(registryArgs, "--registry-owner="+imageRegistry.Owner)
+		registryArgs.WriteString(imageRegistry.Owner)
+	} else {
+		// TODO: Should we keep the owner defaulted as 'hacbs' (or even jbs) ? I think this is legacy and should be removed?
+		registryArgs.WriteString("hacbs")
```
rnc, line 902: In the java code the owner was defaulted e.g.
```
    @CommandLine.Option(names = "--registry-owner", defaultValue = "hacbs")
    String owner;
```
Should we keep this or make it an error condition if the owner is not specified?

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -1434,6 +1442,11 @@ func (r *ReconcileDependencyBuild) handleDeployPipelineRunReceived(ctx context.C
 
 		success := pr.Status.GetCondition(apis.ConditionSucceeded).IsTrue()
 		if success {
+
+			fmt.Printf("### handleDeployPipelineRunReceived pr.Name %#v \n", pr.Name)
+			attempt := db.Status.BuildAttempts[len(db.Status.BuildAttempts)-1]
+			fmt.Printf("### handleDeployPipelineRunReceived attempt %#v \n", attempt.Build)
+
```
rnc, line 1449: ( TODO: Remove )

rnc:
Current status - there is an issue with rebuilding contaminants

pkg/reconciler/dependencybuild/buildrecipeyaml.go
stuartwdouglas: We can remove this IMHO

java-components/cache/src/test/java/com/redhat/hacbs/artifactcache/ContainerRegistryTestResourceManager.java
```diff
@@ -153,7 +153,10 @@ private List<Path> createTestLayers() throws IOException {
             Files.writeString(shaFile, sha1);
         }
 
-        return List.of(layer1Path, layer2Path, layer3Path);
+        // HACK : OCIRepositoryClient assumes that the artifacts are in a directory and it then places them
+        //        within 'artifacts/...'. This is due to build-trusted-artifact changes as its storage stores
+        //        the contents of a directory not including the directory itself.
+        return List.of(layer1Path, layer2Path, layer3Path.resolve("com"));
```
rnc, line 159: Not overly fond of this, but the current test doesn't match how oras stores (it stores the _contents_ of a directory versus the directory itself).

pkg/reconciler/dependencybuild/buildrecipeyaml.go
stuartwdouglas: I think this can come later. It is not 100% clear if we will be pushing to git at all, or if that will come from repour/PNC.

PR #1842, opened 2024-06-17
Fix hermetic builds

rnc:
This is partly to fix issues due to my changes in https://github.com/redhat-appstudio/jvm-build-service/pull/1806

It also adds the ability to configure hermetic builds when deploying.


pkg/reconciler/dependencybuild/scripts/gradle-build.sh
```diff
@@ -1,11 +1,11 @@
...
 export GRADLE_USER_HOME="$(workspaces.build-settings.path)/.gradle"
 mkdir -p "${GRADLE_USER_HOME}"
-mkdir -p "$HOME/.m2/"
+mkdir -p "${HOME}/.m2/repository"
 
 #copy back the gradle folder for hermetic
-cp -r /maven-artifacts/.gradle/* "$GRADLE_USER_HOME/" || true
-cp -r /maven-artifacts/.m2/* "$HOME/.m2/" || true
+cp -r $(workspaces.source.path)/source/maven-artifacts/.gradle/* "$GRADLE_USER_HOME/" || true
+cp -r $(workspaces.source.path)/source/maven-artifacts/* "$HOME/.m2/repository/" || true
```
rnc, line 8: This is slightly inconsistent in that the gradle directory is stored as .gradle while the Maven m2 is stored differently (note that [pkg/reconciler/dependencybuild/scripts/maven-build.sh](https://github.com/redhat-appstudio/jvm-build-service/pull/1842/files#diff-62c2c28ec336e6db726208663357f1f70d9fca6e3ef5dcc5b82df5a40fe5e2d5) matches )

stuartwdouglas:
LGTM, but after discussions last week we are going to move away from the 'build twice' approach, and instead use the domain socket based proxy to emulate firewalling. This should result in a lot of simplification.

PR #1862, opened 2024-06-20
Refactor to simplify module structure

rnc:


tecarter94:
@stuartwdouglas FYI. You'd told me to create common-sbom. @rnc The intent was that I'd add more sbom related code to the module I'd created, so I'm not sure this is quite the right approach?

stuartwdouglas:
@rnc The reason I did not go with the common-resources approach was because we will want the domain socket proxy to be as lightweight as possible, so I don't want it pulling in additional dependencies such as the kubernetes APIs. 

The domain socket proxy will need to run inside the build task so memory usage will be key.

rnc:
> @rnc The reason I did not go with the common-resources approach was because we will want the domain socket proxy to be as lightweight as possible, so I don't want it pulling in additional dependencies such as the kubernetes APIs.
> 
> The domain socket proxy will need to run inside the build task so memory usage will be key.

Fair enough, I didn't know that. I think it was the existing package structure putting a Maven GAV into .hacbs.common.sbom (rather than e.g. .hacbs.common.maven) that 'felt weird' to me. The existing proxy:server dependency tree has quite a bit in it?

PR #1866, opened 2024-06-21
Refactor CLI to use same package structure as rest of repository

rnc:
(Split from a prior PR.)

Note that all of the poms still have e.g. 
```
    <groupId>io.github.redhat-appstudio.jvmbuild</groupId>
    <artifactId>jvm-build-service-parent</artifactId>
```


stuartwdouglas:
`io.github.redhatappstudio.jvmbuild` is probably the more correct package name, but the appstudio name is being phased out. There is a good change this repo will be moved to either konflux-ci or somewhere else, in which case we should get a new package name.

That said we might as well make it consistent for now.

rnc:
> `io.github.redhatappstudio.jvmbuild` is probably the more correct package name, but the appstudio name is being phased out. There is a good change this repo will be moved to either konflux-ci or somewhere else, in which case we should get a new package name.
> 
> That said we might as well make it consistent for now.

Agreed, I didn't change the poms as until naming has fully settled it seemed redundant - hence just making it consistent for now.

PR #1884, opened 2024-06-26
Configure build recipe

tecarter94:
Added configuring of build recipe via mounting file as config map.


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/analyser/build/LookupBuildInfoCommand.java
```diff
@@ -131,36 +134,42 @@ public class LookupBuildInfoCommand implements Runnable {
...
 
-            CacheBuildInfoLocator buildInfoLocator = RestClientBuilder.newBuilder().baseUri(new URI(cacheUrl))
+            if (buildRecipePath != null) {
+                Log.infof("Using build recipe from %s", buildRecipePath);
+                infoString = Files.readString(Path.of(buildRecipePath));
+                Log.infof("Reading %s", infoString);
+            } else {
+                CacheBuildInfoLocator buildInfoLocator = RestClientBuilder.newBuilder().baseUri(new URI(cacheUrl))
                     .build(CacheBuildInfoLocator.class);
-            BuildRecipeInfo buildRecipeInfo = buildInfoLocator.resolveBuildInfo(scmInfo.getUri(), version);
```
stuartwdouglas, line 140: The infoString should be deserialized into a BuildRecipeInfo, and then used instead of this lookup. I.e. rather than skipping all the logic if it is provided, in should just replace what is looked up from the git repo.

tecarter94:
Tests are WIP.

rnc:
@tecarter94 I think for the tests, you'll need to rebase.

rnc:
@tecarter94 Wondering - if the config map is on dependencybuild would it be useful to enhance https://github.com/redhat-appstudio/jvm-build-service/blob/main/java-components/cli/src/main/java/com/redhat/hacbs/cli/builds/BuildCreateCommand.java to allow someone to run this from the CLI?

tecarter94:
> @tecarter94 Wondering - if the config map is on dependencybuild would it be useful to enhance https://github.com/redhat-appstudio/jvm-build-service/blob/main/java-components/cli/src/main/java/com/redhat/hacbs/cli/builds/BuildCreateCommand.java to allow someone to run this from the CLI?

Ooh good point. Will add to this PR.

stuartwdouglas (approved):

PR #1892, opened 2024-06-27
Add jvm-build-image-secrets to pipeline to avoid using REGISTRY_TOKEN

rnc:
Following up on #1824 

This links to secret to the serviceaccount being used by the pipeline. As the secret is a dockerconfigjson this is then used by Tekton to create a `.docker/config.json`. See https://tekton.dev/docs/pipelines/auth/#configuring-docker-authentication-for-docker

### DIFF
--- a/deploy/base/sa.yaml
+++ b/deploy/base/sa.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pipeline
+  namespace: ${JBS_WORKER_NAMESPACE}
+secrets:
+  - name: jvm-build-image-secrets
--- a/pkg/reconciler/dependencybuild/dependencybuild.go
+++ b/pkg/reconciler/dependencybuild/dependencybuild.go
@@ -1181,6 +1181,8 @@ func (r *ReconcileDependencyBuild) createLookupBuildInfoPipeline(ctx context.Con
 		{Name: "GIT_TOKEN", ValueFrom: &v1.EnvVarSource{SecretKeyRef: &v1.SecretKeySelector{LocalObjectReference: v1.LocalObjectReference{Name: v1alpha1.GitSecretName}, Key: v1alpha1.GitSecretTokenKey, Optional: &trueBool}}},
 	}
 	if jbsConfig.ImageRegistry().SecretName != "" {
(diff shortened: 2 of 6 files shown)


deploy/base/sa.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pipeline
```
rnc, line 4: Default service account name is 'pipeline' for OpenShift Pipelines ( https://docs.openshift.com/pipelines/1.14/install_config/customizing-configurations-in-the-tektonconfig-cr.html#op-changing-default-service-account_customizing-configurations-in-the-tektonconfig-cr ). 
( For future : This may be different under Konflux - according to @stuartwdouglas its `appstudio-pipeline` )

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -1181,6 +1181,8 @@ func (r *ReconcileDependencyBuild) createLookupBuildInfoPipeline(ctx context.Con
 		{Name: "GIT_TOKEN", ValueFrom: &v1.EnvVarSource{SecretKeyRef: &v1.SecretKeySelector{LocalObjectReference: v1.LocalObjectReference{Name: v1alpha1.GitSecretName}, Key: v1alpha1.GitSecretTokenKey, Optional: &trueBool}}},
 	}
 	if jbsConfig.ImageRegistry().SecretName != "" {
+		// Builds or tooling mostly use the .docker/config.json directly which is updated via Tekton/Kubernetes secrets. But the
+		// Java code may require the token as well.
 		envVars = append(envVars, v1.EnvVar{Name: "REGISTRY_TOKEN", ValueFrom: &v1.EnvVarSource{SecretKeyRef: &v1.SecretKeySelector{LocalObjectReference: v1.LocalObjectReference{Name: jbsConfig.ImageRegistry().SecretName}, Key: v1alpha1.ImageSecretTokenKey, Optional: &secretOptional}}})
 	}
```
rnc, line 1187: Haven't completely purged the use of REGISTRY_TOKEN as it might have to passed through as an environment variable to the Java tooling (which looks for `registry.token`).

PR #1895, opened 2024-06-28
Remove relocations

stuartwdouglas:
The idea of this was to be able to transparently sub in PNC built artifacts when building managed services. This use case no longer exists and these have never been used, so there is no point keeping this code around.


pkg/apis/jvmbuildservice/v1alpha1/jbsconfig_types.go
```diff
@@ -47,12 +47,13 @@ type JBSConfigSpec struct {
...
-	CacheSettings      CacheSettings              `json:"cacheSettings,omitempty"`
-	BuildSettings      BuildSettings              `json:"buildSettings,omitempty"`
+	SharedRegistries []ImageRegistry   `json:"sharedRegistries,omitempty"`
+	Registry         ImageRegistrySpec `json:"registry,omitempty"`
+	MavenDeployment  MavenDeployment   `json:"mavenDeployment,omitempty"`
+	GitSourceArchive GitSourceArchive  `json:"gitSourceArchive,omitempty"`
+	CacheSettings    CacheSettings     `json:"cacheSettings,omitempty"`
+	BuildSettings    BuildSettings     `json:"buildSettings,omitempty"`
+	// Deprecated
 	RelocationPatterns []RelocationPatternElement `json:"relocationPatterns,omitempty"`
```
rnc, line 57: What should the future policy be on deprecated code or code we want to remove within types like this i.e. should we have a cut-off point and purge stuff (even though its non-backwards compatible)
cc @vibe13

stuartwdouglas: The problem is the CRDs on the cluster, we will need a way to make sure they are handled cleanly. I guess I should have done what we did for the other ones and make sure it is cleaned.

vibe13: yes if we need to nuke things for the work ahead, let's make a pre-cleanup tag as a backup. cannot tell now, but it could be that we will need to bring some things back in the future...

PR #1921, opened 2024-07-16
Update controller-gen

stuartwdouglas:


rnc:
Can you explain this change a bit more?

stuartwdouglas:
It is just a version update, of the kubebuilder generator (i.e. the bit that generates the CRDs from the model classes).

rnc:
For posterity, the failing test was fixed in #1960

PR #1932, opened 2024-07-16
Rename workspace and mount paths

rnc:
This is meant to change our layout to be closer to the Konflux standard (which bases sources in `$(workspaces.source.path)/source` and stores the git repository in the oci artifact in the root directory not in a subdirectory.


vibe13 (approved):
FWIW, makes sense to me :-)

PR #1955, opened 2024-07-17
Update to OpenShift Pipelines 1.15

rnc:
Follows on from the 1.14 upgrade in #1915


rnc:
I think the changes in https://github.com/redhat-appstudio/jvm-build-service/pull/1921 to update the controller-gen are needed (See also https://www.github.com/kubernetes-sigs/controller-tools/issues/880)

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/redhat-appstudio/jvm-build-service
 
-go 1.21
+go 1.22.0
```
stuartwdouglas, line 3: Won't this also require an updated Dockerfile with the new go version?

rnc: Ah, yes, I'd missed there was a go reference in the top level Dockerfile.

rnc: Although, whether go1.22 is available is another question...

PR #1958, opened 2024-07-17
Remove Hermetic Builds

stuartwdouglas:
The approach of building twice is too slow and unwieldly, instead it was decided to use a proxy based approach that will limit and track the downloaded dependencies.

Removing this option greatly simplifies the pipeline code.


vibe13 (approved):
Makes sense to me!

PR #1965, opened 2024-07-19
Update owners to avoid spamming Stuart

rnc:
Trying to avoid the openshift ci bot spamming Stuart all of the time which could get annoying :-(


vibe13 (approved):
looks good to me! We will let @stuartwdouglas decide if he wants to be in the `emeritus_approvers` list :-). Thanks!

PR #1977, opened 2024-07-23
Containerbuild / konflux

rnc:


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/TagDeployCommand.java
```diff
@@ -86,20 +85,25 @@ public void run() {
...
+//                var git = Git.builder(gitURL, gitIdentity, gitToken.get(), gitDisableSSLVerification);
+//                if (reuseRepository) {
+//                    git.initialise(scmUri);
+//                } else {
+//                    Log.warnf("Not reusing repository; creating under %s", scmUri);
+//                    git.create(scmUri);
+//                }
+//                Log.infof("Pushing changes back to URL %s", git.getName());
+//                archivedSourceTags = git.add(sourcePath, commit, imageId);
+//            }
```
rnc, line 106: @vibe13 To discuss ; I don't think we need the Git push changes here (they are after the preprocess changes (which are also stored in git) but also include any pre-build script changes.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/BuildVerifyCommand.java
```diff
@@ -241,6 +240,7 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
         }
     }
 
+    // TODO: ### For container-builds, should sbom generation be delegated to the task within that? If it supports it?
     private void generateBuildSbom() {
```
rnc, line 244: @vibe13 To discuss ; Previously JBS handled SBOM after the build - I suspect now Konflux should handle that?

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -507,8 +432,240 @@ mv $(workspaces.source.path)/source/.jbs/build.sh $(workspaces.source.path)`, or
...
+			Params: []tektonpipeline.Param{
+				{
+					Name: "url",
+					Value: tektonpipeline.ParamValue{
+						Type:      tektonpipeline.ParamTypeString,
+						StringVal: v1alpha1.KonfluxBuildDefinitions,
+					},
+				},
+				{
+					// TODO: ### Currently always using 'head' of branch.
```
rnc, line 450: I'm hard-coding `main` just below as the build-definitions repository is not tagged and everyone (I think) uses latest.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -507,8 +432,240 @@ mv $(workspaces.source.path)/source/.jbs/build.sh $(workspaces.source.path)`, or
...
+					},
+				},
+
+				// TODO: ### How to pass build-settings/tls information to buildah task?
+				//       Note - buildah-oci-ta task has no defined workspace
+				//Workspaces: []tektonpipeline.WorkspacePipelineTaskBinding{
+				//	//{Name: WorkspaceBuildSettings, Workspace: WorkspaceBuildSettings},
+				//	{Name: WorkspaceSource, Workspace: WorkspaceSource},
+				//	//{Name: WorkspaceTls, Workspace: WorkspaceTls},
+				//},
```
rnc, line 504: Currently I haven't figured out how (if at all) we can pass the setting.xml/tls shared workspaces - I suspect we'll need a different approach.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -507,8 +432,240 @@ mv $(workspaces.source.path)/source/.jbs/build.sh $(workspaces.source.path)`, or
...
+			//	Name:            "create-post-build-image",
+			//	Image:           strings.TrimSpace(strings.Split(buildTrustedArtifacts, "FROM")[1]),
+			//	ImagePullPolicy: v1.PullIfNotPresent,
+			//	SecurityContext: &v1.SecurityContext{RunAsUser: &zero},
+			//	Env:             secretVariables,
+			//	ComputeResources: v1.ResourceRequirements{
+			//		Requests: v1.ResourceList{"memory": limits.defaultBuildRequestMemory, "cpu": limits.defaultRequestCPU},
+			//		Limits:   v1.ResourceList{"memory": limits.defaultBuildRequestMemory, "cpu": limits.defaultLimitCPU},
+			//	},
+			//	Script: postBuildImageArgs,
```
rnc, line 643: @vibe13 to be discussed - I don't think we need this anymore.

vibe13: Ok to remove the storage of the artifacts which are already stored with "store-post-build-artifacts" task. I would store any change to the scripts, for a full attestation. Wrt the logs, not sure I completely follow, why is it irrelevant if done in a container? I would still store them (regardless if it's a container build or a normal build). Additionally we will store logs in PNC DB. If later on we feel it's not necessary anymore, we can remove the storage of logs here.

vibe13 (approved):
I have put some more questions for the sake of my understanding but overall it makes sense to me, thanks Nick!

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/BuildVerifyCommand.java
```diff
@@ -214,8 +214,7 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
             }
             for (var i : contaminatedGavs.entrySet()) {
                 if (!i.getValue().getAllowed()) {
-                    gavs.removeAll(i.getValue().getContaminatedArtifacts());
```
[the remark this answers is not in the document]
rnc: IntelliJ recommended as it said the other had performance issues.

java-components/resource-model/src/main/resources/crds/jvmbuildservice.io_jbsconfigs.yaml
```diff
@@ -92,6 +92,10 @@ spec:
                   workerThreads:
                     type: string
                 type: object
+              containerBuilds:
```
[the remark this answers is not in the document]
rnc: None. One is a copy of the other. I think from memory Stuart set them up like that - I think its due to the fact that the Dockerfile can't reference outside of its context (I asked this in https://github.com/redhat-appstudio/jvm-build-service/pull/796 and its due to context of `java_components` (I think) in https://github.com/openshift/release/blob/master/ci-operator/config/redhat-appstudio/jvm-build-service/redhat-appstudio-jvm-build-service-main.yaml#L14 )

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/BuildVerifyCommand.java
rnc: At the moment we don't have any way of handling / storing the sbom - this was handled by BuildVerifyCommand but as we don't have a way in the buildah based task to access the .m2/gradle repositories I think this now needs to be handled specifically by Konflux.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/deploy/TagDeployCommand.java
rnc: I think this is one I wanted to discuss offline. It relates to [this](https://docs.google.com/document/d/1qJ0rCdBs3iB0ETkxxNaIqRxYpoiMmpcVzPCgheaotgw/edit#bookmark=id.lis275csct3t) - specifically storing sources after preprocess and after pre-build but before build. PNC doesn't do that and doesn't store any changes user's script so I think we can drop it.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -507,8 +432,240 @@ mv $(workspaces.source.path)/source/.jbs/build.sh $(workspaces.source.path)`, or
...
+
+		// Results for https://github.com/konflux-ci/build-definitions/tree/main/task/buildah-oci-ta/0.2
+		// IMAGE_DIGEST
+		// IMAGE_URL
+		ps.Results = append(ps.Results, tektonpipeline.PipelineResult{Name: PipelineResultImage, Value: tektonpipeline.ResultValue{Type: tektonpipeline.ParamTypeString, StringVal: "$(tasks." + BuildTaskName + ".results." + PipelineResultImage + ")"}})
+		ps.Results = append(ps.Results, tektonpipeline.PipelineResult{Name: PipelineResultImageDigest, Value: tektonpipeline.ResultValue{Type: tektonpipeline.ParamTypeString, StringVal: "$(tasks." + BuildTaskName + ".results." + PipelineResultImageDigest + ")"}})
+	} else {
+		fmt.Printf("### Not running container builds \n")
+
+		buildTask := tektonpipeline.TaskSpec{
```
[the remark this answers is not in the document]
rnc: Agreed. It usefully allows comparisons to make sure I haven't broken anything!

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -785,10 +784,6 @@ func (r *ReconcileDependencyBuild) handleBuildPipelineRunReceived(ctx context.Co
 					// 	But this is now stored as
 					// 		"verificationFailures": "{\"commons-lang:commons-lang:jar:2.5\":[]}"
 					verificationResults = i.Value.StringVal
-				} else if i.Name == PipelineResultGavs {
```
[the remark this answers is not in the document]
rnc: No it was just code that I think is irrelevant and could never be called. Nothing ever sets `PipelineResultGavs` and it was always handled by `PipelineResultDeployedResources`. I think (!)

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -287,16 +295,36 @@ func createPipelineSpec(log logr.Logger, tool string, commitTime int64, jbsConfi
...
+			"\nCOPY --from=0 /root/project/ /root/project/" +
+			// Don't think we need to mess with keystore as copy-artifacts is simply calling copy commands.
+			"\nRUN /opt/jboss/container/java/run/run-java.sh " + doSubstitution(strings.Join(copyArtifactsArgs, " "), []tektonpipeline.Param{}, commitTime, buildRepos) +
+			"\nFROM scratch" +
+			"\nCOPY --from=1 /root/project/artifacts /"
+	} else {
+		kf = kf +
+			"\nFROM scratch" +
+			"\nCOPY --from=0 /root/project/artifacts /"
+	}
```
rnc, line 327: This above block is somewhat of a hack. The original code that @dwalluck added ran in the **verify** phase - but as we can't use a shared workspace it needs to run in the build phase hence appending it to the Containerfile. However its written in Java so I have to use another `FROM`. In future we should probably look to see if we can migrate it to shell.

deploy/minikube-ci.sh
```diff
@@ -23,7 +23,6 @@ echo "Tekton controller is running"
 
 #CRDS are sometimes racey
 kubectl apply -k $DIR/crds/base
-kubectl apply -f https://raw.githubusercontent.com/openshift/api/master/quota/v1/0000_03_quota-openshift_01_clusterresourcequota.crd.yaml
 sleep 2
```
dwalluck, line 26: Does anyone know why we need these `sleep` commands after `kubectl apply`? I know it wasn't added in this PR, but I am curious.

rnc: Minikube is slow to load the CRDs

pkg/reconciler/dependencybuild/buildrecipeyaml.go
rnc: Yes `CopyArtifactsCommand`. Its not a major issue and is low priority as I have 'if-blocked' it so the secondary FROM isn't used under maven/gradle

PR #2000, opened 2024-07-30
Run reposilite and deploy to during minikube pipeline

tecarter94:


tecarter94:
Not ready for review yet.

deploy/maven-repo.yaml
```diff
@@ -0,0 +1,67 @@
...
+    metadata:
+      labels:
+        app: jvm-build-maven-repo
+    spec:
+      serviceAccountName: pipeline
+      securityContext:
+        runAsUser: 0
+      containers:
+        - name: jvm-build-maven-repo
+          image: dzikoysk/reposilite:3.5.14
```
rnc, line 23: I wonder if this should be `dzikoysk/reposilite:latest` so we pick up eg 3.5.16. Equally I guess that might introduce some instability risk. The alternative would be to use a scheme like ./pkg/reconciler/dependencybuild/scripts/Dockerfile.build-trusted-artifacts and have a Dockerfile.reposilite and substitute in via a variable. This image would then be updated by renovate.

deploy/maven-repo.yaml
```diff
@@ -0,0 +1,67 @@
...
+          imagePullPolicy: IfNotPresent
+          ports:
+            - name: http
+              containerPort: 8080
```
rnc, line 27: Is this file / deployment only used for development? In which case perhaps a more unique port should be used as 8080 is often used for standard deployments during development?

deploy/openshift-development.sh
```diff
@@ -1,4 +1,17 @@
...
+fi
+if [ -z "$MAVEN_REPOSITORY" ]; then
+    export MAVEN_REPOSITORY='http://jvm-build-maven-repo.$(context.taskRun.namespace).svc.cluster.local/releases'
+fi
+
 DIR=`dirname $0`
 $DIR/base-development.sh $1
+
+kubectl delete --ignore-not-found deployment,service jvm-build-maven-repo -n test-jvm-namespace
+cat $DIR/maven-repo.yaml | envsubst '${MAVEN_PASSWORD} ${MAVEN_USERNAME}' | kubectl create -f -
```
rnc, line 17: Slightly pedantic but I think if someone has already set their own maven username/psasword/repository they are probably using their own and reposilite shouldn't be deployed? For example see https://github.com/rnc/jbs-demo/blob/main/setup.sh#L63

pkg/apis/jvmbuildservice/v1alpha1/jbsconfig_types.go
```diff
@@ -114,6 +115,7 @@ type ImageRegistry struct {
 type MavenDeployment struct {
 	Username   string `json:"username,omitempty"`
 	Repository string `json:"repository,omitempty"`
+	OnlyDeploy bool   `json:"onlyDeploy,omitempty"` // we don't want cache attempting to download from empty repo
```
rnc, line 118: What happens without this?

rnc:
In general this is great! Just a few small questions/comments. I think perhaps we should squash the commits a bit - as 60+ is a bit much for a merge :-)

tecarter94:
> In general this is great! Just a few small questions/comments. I think perhaps we should squash the commits a bit - as 60+ is a bit much for a merge :-)

Yeah, 100% agree regarding squashing commits! If there's some way to develop against the minikube tests locally I'm all ears because it's been such a pain pushing all of these commits and waiting for everything to spin up.

pkg/apis/jvmbuildservice/v1alpha1/jbsconfig_types.go
tecarter94: Without this the maven env variables get picked up by the cache pod too and it'll attempt to download every dependency for maven builds from our empty reposilite repo. If we _really_ want to we could potentially configure reposilite to mirror https://repo.maven.apache.org/maven2, but that would require creating config on a local machine and then mounting it as a volume and probably isn't worth it for mere testing. But do let me know if you think otherwise. :slightly_smiling_face:

deploy/openshift-development.sh
tecarter94: I agree, will add some logic around that. :+1:

deploy/maven-repo.yaml
tecarter94: Yes it does, as the same port is used for UI management and upload/download requests.

deploy/maven-repo.yaml
tecarter94: It could potentially introduce some instability risk pinning 'latest' version that's true. Do we actually need to keep the image up to date given that we only use it for testing?

deploy/openshift-development.sh
tecarter94: Done.

rnc:
> > In general this is great! Just a few small questions/comments. I think perhaps we should squash the commits a bit - as 60+ is a bit much for a merge :-)
> 
> Yeah @rnc, 100% agree regarding squashing commits! If there's some way to develop against the minikube tests locally I'm all ears because it's been such a pain pushing all of these commits and waiting for everything to spin up.

While it _should_ be possible to run minikube locally (esp. given there is a minikube-dev.sh script) I've never tried either running minikube locally or trying to get the tests running locally. This is a good point and I htink we should log a ticket for this work.

openshift-with-appstudio-test/e2e/basictests.go
```diff
@@ -181,6 +181,27 @@ func runAbTests(path string, testSet string, pipeline string, ta *testArgs) {
 		}
 	})
 
+	mavenRepoDetails := getMavenRepoDetails(ta)
+
+	ta.t.Run("Maven repo contains artifacts for dependency builds", func(t *testing.T) {
```
rnc, line 186: This test is running (and failing) under the e2e test which is why you aren't getting a clean run.

deploy/maven-repo.yaml
rnc: I meant onto e.g. console-openshift-console.apps..... I think a route generation might be useful ? Something like https://github.com/redhat-appstudio/jvm-build-service/blob/main/deploy/console/base/route.yaml but 
```
kind: Route
apiVersion: route.openshift.io/v1
metadata:
  name: reposilite
  namespace: test-jvm-namespace
  labels:
    app: jvm-build-maven-repo
  annotations:
    openshift.io/host.generated: 'true'
spec:
  to:
    kind: Service
    name: jvm-build-maven-repo
    weight: 100
  port:
    targetPort: http
  tls:
    termination: edge
    insecureEdgeTerminationPolicy: Redirect
  wildcardPolicy: None
```

deploy/maven-repo.yaml
rnc: Thats a very good question! In general it probably shouldn't matter however given 3.5.12 (for example) had a number of vulnerabilities I'm thinking we should at least have a way of it being updated to avoid being flagged for using vulnerable dependencies. However, I'm ok if this is moved to a separate piece of work (as both the above reference and the one in the test suite would need to be handled) if you want (and a ticket is entered)

openshift-with-appstudio-test/e2e/basictests.go
tecarter94: Interesting... that e2e test appears to not use minikube, hence why it fails when I attempt to tunnel from the maven repo service using minikube. May be tricky to figure out how to do this generically. I have no knowledge of how this e2e test works so this should be fun.

deploy/maven-repo.yaml
tecarter94: Ah, gotcha. I'll give it a go.

openshift-with-appstudio-test/e2e/basictests.go
rnc: I was thinking initially exclude that test under e2e? i.e. make it minikube only?

tecarter94: Yes, I had considered that too. One very lazy way would be to check for the presence of the minikube binary.

deploy/maven-repo.yaml
tecarter94: Have added route for UI to 'maven.repo.yaml'. Seems to work fine for accessing UI.

deploy/maven-repo.yaml
tecarter94: I believe I've made the necessary changes for renovate to update reposilite to latest version. Please let me know if anything is messing!

openshift-with-appstudio-test/e2e/basictests.go
rnc: Reading a bit further it looks like initially golang build constraints are being used which unfortunately are per file based (see the Makefile). Where is junit when we need it! I don't know if its feasible to move your test into minikube_test.go (but is there ordering issues then?) Alternatively perhaps in https://github.com/redhat-appstudio/jvm-build-service/blob/main/Makefile#L43 set a variable e.g. `MINIKUBE=true` and use a `t.Skip` ( https://konradreiche.com/blog/how-to-separate-integration-tests-in-go ) if that variable isn't set for your test?

deploy/maven-repo.yaml
rnc: That looks great!

deploy/maven-repo.yaml
rnc: Excellent!

pkg/apis/jvmbuildservice/v1alpha1/jbsconfig_types.go
vibe13: First of all, massive work, thanks @tecarter94 ! Secondly, I think we will remove the cache soonish anyway, so the flag `OnlyDeply` will be removed soon as well? I think we can leave it as it is and make updates in subsequent PRs. WDYT @rnc ? Thanks!

rnc: I agree with having `onlyDeploy` in for now and revisting to how JBS uses the cache later.

tecarter94: Makes sense. Happy to leave the flag in place for now. :+1:

openshift-with-appstudio-test/e2e/basictests.go
tecarter94: Must have missed this comment before @rnc, my bad! I was thinking of setting a flag to skip the test if I can't get port forwarding to work without using kubectl/minikube, so we were thinking along the same lines. Unsure if moving the test to minikube_test.go is feasible.

tecarter94: Port forwarding within e2e env is no longer the issue it seems, but in fact the repo pod not having started. Will look into tomorrow.

rnc: I think the `MINIKUBE=true` and a `t.skip` is the simplest option IMHO.

PR #2076, opened 2024-09-05
Update to the latest jvm-build-service-builder-images

vibe13:


rnc (approved):

rnc (changes requested):
On second thoughts, I think we're missing a change to https://github.com/redhat-appstudio/jvm-build-service-builder-images/blob/main/image-config.yaml#L13 and to the other references to 1.10.13 (in this repo) so I'm sorry but I think this needs further work.

rnc (approved):

PR #2110, opened 2024-09-23
Remove SBOM handling

tecarter94:
https://issues.redhat.com/browse/KJB-38


vibe13:
@rnc @tecarter94 I think it makes sense to remove the SBOM generation part and the jvmimagescan, but I would still keep the console around because it could still be useful? And if we later decide to remove it, I would open another PR.
Since this is a big change, it might be worth creating a tag of jvm-build-service to have a reference of the "old" version.

rnc (approved):

PR #2117, opened 2024-09-25
Convert pre-build task to yaml.

rnc:


deploy/tasks/pre-build.yaml
```diff
@@ -0,0 +1,145 @@
...
+      securityContext:
+        runAsUser: 0
+      env:
+        - name: GIT_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: jvm-build-git-secrets
+              key: .git-credentials
+      script: |
+        $(params.GIT_SCRIPT)
```
rnc, line 91: Currently **not** using Konflux git clone task but the original simple git clone from JBS and passing it through as a script from the operator.

deploy/tasks/pre-build.yaml
```diff
@@ -0,0 +1,145 @@
...
+        runAsUser: 0
+      computeResources:
+        limits:
+          cpu: 300m
+          memory: 512Mi
+        requests:
+          cpu: 10m
+          memory: 512Mi
+      script: |
+        $(params.BUILD_SCRIPT)
```
rnc, line 104: To embed the Containerfile / build-script into the repository **currently** passing them through from the operator (which is generating them). Eventually I think the preprocessor should generate them (for the following steps (`create-pre-build-source` and `create-pre-build-image`) to use).

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -362,87 +359,145 @@ func createPipelineSpec(log logr.Logger, tool string, commitTime int64, jbsConfi
...
-				{Name: PipelineResultGitArchive, Type: tektonpipeline.ResultsTypeString},
+		resolver := tektonpipeline.ResolverRef{
+			// We can use either a http or git resolver. Using http as avoids cloning an entire repository.
+			Resolver: "http",
+			Params: []tektonpipeline.Param{
+				{
+					Name: "url",
+					Value: tektonpipeline.ParamValue{
+						Type:      tektonpipeline.ParamTypeString,
+						StringVal: "https://raw.githubusercontent.com/rnc/jvm-build-service/KJB11/deploy/tasks/pre-build.yaml",
```
rnc, line 374: As per prior PRs will need a follow-up PR to correct the branch/repo.

openshift-with-appstudio-test/e2e/types.go
```diff
@@ -18,7 +18,3 @@ type testArgs struct {
 	pipeline *tektonpipeline.Pipeline
 	run      *tektonpipeline.PipelineRun
 }
-
-func (ta *testArgs) Logf(msg string) {
-	ta.t.Logf("time: %s: %s", time.Now().String(), msg)
-}
```
rnc, line 24: The testing logging formatter does stack frame depth calculation to print out where the logging is from. If its redirected that breaks.

deploy/tasks/pre-build.yaml
vibe13: Yes I believe it would be cleaner that the preprocessor generates the build script and propagates it as a result to following steps.

deploy/tasks/pre-build.yaml
vibe13: The more we use Konflux's tasks the better I think, but not a problem for now to use our own. Can be a subsequent PR.

pkg/reconciler/dependencybuild/buildrecipeyaml.go
vibe13: +1

vibe13 (approved):
Looks good!! Agree on the suggestions to use konflux's tasks and to propagate the build-script from the preprocessor task, can be other PRs.

PR #2119, opened 2024-09-30
Inline Containerfile generation

rnc:
This moves Containerfile generation from the operator to the Java request-processor (specifically the preprocessor step) so it can be reused outside of JBS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,8 +65,8 @@ dev-image:
             echo "ERROR: QUAY_USERNAME is not set"; \
             exit 1; \
     fi
-	docker build . -t quay.io/$(QUAY_USERNAME)/hacbs-jvm-controller:dev
-	docker push quay.io/$(QUAY_USERNAME)/hacbs-jvm-controller:dev
+	docker build . -t quay.io/$(QUAY_USERNAME)/hacbs-jvm-controller:"$${JBS_QUAY_IMAGE_TAG:-dev}"
+	docker push quay.io/$(QUAY_USERNAME)/hacbs-jvm-controller:"$${JBS_QUAY_IMAGE_TAG:-dev}"
 
 dev: dev-image
 	cd java-components && mvn clean install -Dlocal -DskipTests -Ddev
--- a/deploy/tasks/pre-build.yaml
+++ b/deploy/tasks/pre-build.yaml
@@ -46,10 +46,13 @@ spec:
       type: string
     - name: RECIPE_IMAGE
       description: The image from the build recipe to use
+    - name: BUILD_TOOL
+      description: The build tool to use.
+    - name: BUILD_PLUGINS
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -65,8 +65,8 @@ dev-image:
             echo "ERROR: QUAY_USERNAME is not set"; \
             exit 1; \
     fi
-	docker build . -t quay.io/$(QUAY_USERNAME)/hacbs-jvm-controller:dev
-	docker push quay.io/$(QUAY_USERNAME)/hacbs-jvm-controller:dev
+	docker build . -t quay.io/$(QUAY_USERNAME)/hacbs-jvm-controller:"$${JBS_QUAY_IMAGE_TAG:-dev}"
+	docker push quay.io/$(QUAY_USERNAME)/hacbs-jvm-controller:"$${JBS_QUAY_IMAGE_TAG:-dev}"
```
rnc, line 69: Unless the JBS_QUAY_IMAGE_TAG is specified we default to `dev`. With Tekton pipelines, if tag is `latest` the it will use PullAlways. Hence for pipelines setting JBS_QUAY_IMAGE_TAG=latest is advised. We should consider whether to change the default to `latest` (perhaps in another iteration).

deploy/tasks/pre-build.yaml
```diff
@@ -103,7 +106,7 @@ spec:
           memory: 512Mi
       script: |
         $(params.BUILD_SCRIPT)
-        /opt/jboss/container/java/run/run-java.sh $(params.PREPROCESSOR_ARGS)
+        /opt/jboss/container/java/run/run-java.sh $(params.BUILD_TOOL)-prepare $(workspaces.source.path)/source --recipe-image=$(params.RECIPE_IMAGE) --request-processor-image=$(params.JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE) --disabled-plugins=$(params.BUILD_PLUGINS)
```
rnc, line 109: This completely inlines the request processor arguments (which makes it easier for PNC to use)

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
```diff
@@ -1,21 +1,83 @@
...
+                """
+                    FROM scratch
+                    COPY --from=0 /var/workdir/workspace/artifacts /""";
+        }
+        try {
+            Files.writeString(Paths.get(jbsDirectory.toString(), "Containerfile"), containerFile);
+        } catch (IOException e) {
+            Log.errorf("Unable to write Containerfile", e);
+            throw new RuntimeException(e);
+        }
```
rnc, line 81: This inlines the Containerfile generation into the source directory for the next steps (the git archiving and image generation) in the task to use.

PR #2142, opened 2024-10-03
MW pipeline with Konflux git clone.

matejonnet:


deploy/tasks/pre-build-minimal.yaml
```diff
@@ -0,0 +1,72 @@
...
+  #   computeResources: {}
+  #   volumeMounts:
+  #     - mountPath: /var/workdir/source
+  #       name: source
+  workspaces:
+    - name: source
+      description: The git repo will be cloned onto the volume backing this Workspace.
+      # mountPath: /var/workdir
+    # - name: tls
+  steps:
```
rnc, line 45: Few comments:

- In https://github.com/redhat-appstudio/jvm-build-service/pull/2144 I've created a more 'minimal' version of the prebuild task (and also marking more parameters as optional). I think it would be better to try to adopt/improve that than creating another version?
- I'm not sure why you needed to remove the workspace mountpath?
- I agree tls workspace could be removed (potentially as another iteration).

rnc:
Given changes in https://github.com/redhat-appstudio/jvm-build-service/pull/2286 and in gitlab I think we can close this.

PR #2153, opened 2024-10-08
Domain proxy integration

tecarter94:
Convert domain proxy from Java to Go. Integrate domain proxy into buildah-oci-ta. Add generic internal proxy support to domain proxy. Integrate building of domain proxy into pipeline build processes. Implement domain proxy integration tests.

https://issues.redhat.com/browse/JBS-54


pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -516,6 +516,13 @@ func createPipelineSpec(log logr.Logger, tool string, commitTime int64, jbsConfi
 						StringVal: tlsVerify,
 					},
 				},
+				{
+					Name: "JVM_BUILD_SERVICE_DOMAIN_PROXY_IMAGE",
```
matejonnet, line 520: This would probably be an image providing build enviroment. Meaning all build environment images, must have the domain-proxy client. Or do I miss something ?

java-components/domain-proxy/.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -0,0 +1,98 @@
...
+import java.net.Authenticator;
+import java.net.PasswordAuthentication;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+
+public final class MavenWrapperDownloader
+{
```
matejonnet, line 31: What's the purpose of this class ?

matejonnet (changes requested):

pkg/reconciler/dependencybuild/buildrecipeyaml.go
tecarter94: @matejonnet this PR isn't ready for review. It's still wip. Please ignore it for now, thanks.

tecarter94: Still lots of changes to be made.

java-components/domain-proxy/common/src/main/java/com/redhat/hacbs/domainproxy/common/CommonIOUtil.java
```diff
@@ -1,84 +1,153 @@
...
+            int bytesWrittenLeft = 0;
+            int bytesWrittenRight = 0;
+            final ByteBuffer buffer = ByteBuffer.allocate(byteBufferSize);
+            try (final Selector selector = Selector.open()) {
+                leftChannel.configureBlocking(false);
+                rightChannel.configureBlocking(false);
+                leftChannel.register(selector, SelectionKey.OP_READ);
+                rightChannel.register(selector, SelectionKey.OP_WRITE);
+                long bytesTransferredTime = System.currentTimeMillis();
+                while (leftChannel.isOpen() && rightChannel.isOpen()) {
```
matejonnet, line 50: This while loop exits on time-out or closed connections, but the connections are closed only after this loop ends?

tecarter94: That's right. We return early upon timeout, which causes the channels to be closed. We also close the channels explicitly if one of them closes (ending while loop iteration) - one circumstance where this might happen is if a client (i.e. maven or curl) terminates early.

domain-proxy/Dockerfile.all-in-one
```diff
@@ -0,0 +1,15 @@
+FROM golang:alpine AS builder
```
rnc, line 1: Can this not be from a RH golang?

domain-proxy/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/redhat-appstudio/jvm-build-service/domain-proxy
+
+go 1.23
```
rnc, line 3: Is there a reason this is using a different version of golang to the top level (although there is https://github.com/redhat-appstudio/jvm-build-service/pull/2045 I guess )

domain-proxy/pkg/server/server.go
```diff
@@ -0,0 +1,246 @@
...
+	"strings"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	HttpPort                     = 80
+	HttpsPort                    = 443
+	ProxyTargetWhitelistKey      = "PROXY_TARGET_WHITELIST"
+	DefaultProxyTargetWhitelist  = "gariscus.com,neverssl.com,repo1.maven.org,repo.maven.apache.org,repository.jboss.org,packages.confluent.io,jitpack.io,repo.gradle.org,plugins.gradle.org"
```
rnc, line 20: What are the first two?

deploy/tasks/buildah-oci-ta.yaml
```diff
@@ -0,0 +1,744 @@
...
+      description: The byte buffer size to use for the domain proxy.
+      type: string
+      default: 32768
+    - name: PROXY_TARGET_WHITELIST
+      description: Comma separated whitelist of target hosts for the domain proxy.
+      type: string
+    - name: INTERNAL_PROXY_ADDRESS
+      description: Host and port of proxy used internally by the domain proxy.
+      type: string
+    - name: INTERNAL_PROXY_USER
```
rnc, line 152: Is this an up-to-date copy of buildah-oci-ta? I'm not sure if the copy in gitlab has any changes? Can you summarise the changes (somewhere?!?) ?

domain-proxy/Dockerfile.all-in-one
```diff
@@ -0,0 +1,15 @@
...
+RUN go build -o domain-proxy-server github.com/redhat-appstudio/jvm-build-service/domain-proxy/cmd/server
+RUN go build -o domain-proxy-client github.com/redhat-appstudio/jvm-build-service/domain-proxy/cmd/client
+
+FROM quay.io/konflux-ci/buildah-task:latest@sha256:5cbd487022fb7ac476cbfdea25513b810f7e343ec48f89dc6a4e8c3c39fa37a2
+USER 0
+RUN microdnf install -y iproute
+WORKDIR /work/
+
+COPY --from=builder /work/domain-proxy-server /app/domain-proxy-server
+COPY --from=builder /work/domain-proxy-client /app/domain-proxy-client
```
rnc, line 15: Rather than splitting my comments across multiple locations which will get rather confusing I've embedded it all here. 

I am unclear as to why Dockerfile.all-in-one and Dockerfile.local are required?

In terms of the package layout I also have a some thoughts -

- Can we move the common code (e.g. pkg/client...) under the pkg structure (e.g. into pkg/domain-proxy/client/...)?
- Can the domain-proxy/cmd/client and  domain-proxy/cmd/server be moved under cmd (next to cmd/controller)?
- Do we need multiple go.mod files i.e. as the above simplifies the structure down? If the golang version needs to be 1.23 (See my other comment) we can always merge that associated PR.
- I don't think the executables need to be output straight into the working directory ; instead use similar to e.g.
```
go build -o out/jvmbuildservice
go build -o out/domain-proxy-server
```
   This then negates the `.gitignore` change
- In the Makefile I don't think there is a need to change `dev: dev-image-controller` ; instead move the golang compiling under the `build` target and the quay image generation  for both controller and domain-proxy can be under `dev-image`?

domain-proxy/pkg/client/client.go
```diff
@@ -0,0 +1,94 @@
+package client
+
+import (
+	"fmt"
+	. "github.com/redhat-appstudio/jvm-build-service/domain-proxy/pkg/common"
```
rnc, line 5: Using this import structure doesn't work for me in intellij hence my other restructuring thoughts :-)

domain-proxy/pkg/server/server.go
tecarter94: Those shouldn't have been committed. They were part of a separate changelist I had for testing but then got committed when I was doing refactoring. They'll be removed.

domain-proxy/Dockerfile.all-in-one
tecarter94: The only option is [rhel8/go-toolset](https://catalog.redhat.com/software/containers/rhel8/go-toolset/5b9c810add19c70b45cbd666), which requires either use of registry tokens or RH login to pull.

domain-proxy/go.mod
tecarter94: The reason is to use the latest. It's a separate application to the 'top level'. We could update the 'top level' too I suppose.

deploy/tasks/buildah-oci-ta.yaml
tecarter94: It is an up to date copy. I have no idea what the one in gitlab has.

Screenshot From 2024-11-29 11-23-20
Screenshot From 2024-11-29 11-23-07
Screenshot From 2024-11-29 11-22-57
Screenshot From 2024-11-29 11-22-49

domain-proxy/pkg/client/client.go
tecarter94: Interesting. I could simply change the module name to fix this for you.

domain-proxy/Dockerfile.all-in-one
tecarter94: > I am unclear as to why Dockerfile.all-in-one and Dockerfile.local are required?

'all-in-one' is used to build the domain proxy container as part of Github CI for use with minikube and buildah-oci-ta.
'local' is used to build the domain proxy container as part of local testing with minikube and buildah-oci-ta.

> Can we move the common code (e.g. pkg/client...) under the pkg structure (e.g. into pkg/domain-proxy/client/...)?
Can the domain-proxy/cmd/client and domain-proxy/cmd/server be moved under cmd (next to cmd/controller)?
Do we need multiple go.mod files i.e. as the above simplifies the structure down? If the golang version needs to be 1.23 (See my other comment) we can always merge that associated PR.

I don't think it makes sense to merge them all. They're separate applications and will be built differently. I could understand putting them all in the same structure if they directly depended on Go packages in the existing pkg structure. It's my understanding that a large amount of that existing Go code (maybe all?) will disappear at some point anyway with JBS disappearing.

> I don't think the executables need to be output straight into the working directory ; instead use similar to e.g. out
In the Makefile I don't think there is a need to change dev: dev-image-controller ; instead move the golang compiling under the build target and the quay image generation for both controller and domain-proxy can be under dev-image?

Both of these make sense to me. Will look at doing so.

domain-proxy/pkg/server/server.go
tecarter94: They've been removed.

domain-proxy/Dockerfile.all-in-one
tecarter94: Ultimately I found it way easier to go with all of your suggestions. :smile:

domain-proxy/Dockerfile.all-in-one
tecarter94: Which apparently works just fine...

pkg/reconciler/dependencybuild/buildrecipeyaml.go
tecarter94: Comment is somewhat outdated now. But yes, the domain proxy will need to be part of the image referenced to by the BUILD_IMAGE param, assuming HERMETIC and ENABLE_DOMAIN_PROXY params are enabled.

tecarter94:
The domain proxy specific buildah-oci-ta changes.  @rnc @matejonnet 

Screenshot From 2024-12-11 13-45-27
Screenshot From 2024-12-11 13-40-43
Screenshot From 2024-12-11 13-36-33

rnc:
This all looks goodat first glance. While I realise the e2e tests are failing given STONEBLD-2711 I don't think we should be too concerned about them tbh.

PR #2156, opened 2024-10-08
Use standard buildah-oci-ta

rnc:


pkg/apis/jvmbuildservice/v1alpha1/systemconfig_types.go
```diff
@@ -44,3 +44,10 @@ type SystemConfigList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []SystemConfig `json:"items"`
 }
+
+const (
+	KonfluxGitDefinition          = "https://raw.githubusercontent.com/konflux-ci/build-definitions/refs/heads/main/task/git-clone/0.1/git-clone.yaml"
+	KonfluxPreBuildDefinitions    = "https://raw.githubusercontent.com/rnc/jvm-build-service/main/deploy/tasks/pre-build.yaml"
+	KonfluxBuildDefinitions       = "https://raw.githubusercontent.com/konflux-ci/build-definitions/refs/heads/main/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml"
+	KonfluxMavenDeployDefinitions = "https://raw.githubusercontent.com/rnc/jvm-build-service/main/deploy/tasks/maven-deployment.yaml"
```
rnc, line 52: @tecarter94 I realise this conflicts somewhat with your proxy PR. The problem is Konflux are constantly committing to the `buildah-oci-ta` task and it has diverged since I forked it. This PR removes the final modifications I made to the fork isolating them into the JBS code. 
I think what is best **after** this, is to refork / update the buildah task again (and update the above `KonfluxBuildDefinitions`) so that we're running with a recent copy.

pkg/reconciler/dependencybuild/dependencybuild.go
```diff
@@ -646,6 +647,17 @@ func (r *ReconcileDependencyBuild) handleStateBuilding(ctx context.Context, db *
 			},
 		},
 	}
+	// TODO: ### Enclose this within an annotation to denote test CI system in use?
```
rnc, line 650: Really we only need to limit the task resources on CI due to the fact that minikube and the e2e can't allocate the CPU/memory that the buildah-oci-ta task specifies by default. I think I could create another annotation to represent the CI tests being used in order to wrap this block...

pkg/apis/jvmbuildservice/v1alpha1/systemconfig_types.go
tecarter94: Cheers. It might be a while till I proceed with #2153, so I'll revisit this at that point.

deploy/tasks/buildah-oci-ta.yaml
```diff
@@ -398,11 +392,6 @@ spec:
           done < <(find $ADDITIONAL_SECRET_TMP -maxdepth 1 -type f -exec basename {} \;)
         fi
 
-        # TODO: Rename to JBS_CACHE_URL?
-        if [ -n "$CACHE_URL" ]; then
-          BUILDAH_ARGS+=("--build-arg=CACHE_URL=$CACHE_URL")
-        fi
-
         unshare -Uf $UNSHARE_ARGS --keep-caps -r --map-users 1,1,65536 --map-groups 1,1,65536 -w ${SOURCE_CODE_DIR}/$CONTEXT -- buildah build \
```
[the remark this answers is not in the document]
rnc: I've now removed it in this PR to make things clearer.

pkg/reconciler/dependencybuild/dependencybuild.go
rnc: Latest commit has wrapped them in a new annotation.

.github/workflows/minikube.yaml
```diff
@@ -4,7 +4,7 @@ on:
     branches: [ main ]
 jobs:
   wait-for-images:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
rnc, line 7: Solves the problems with minikube / oc due to https://www.github.com/actions/runner-images/issues/10636

rnc:
Discussed with @matejonnet offline - merging this to move us forward.

PR #2172, opened 2024-10-12
Generate build script within preprocessor

rnc:
…ol_version


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
```diff
@@ -27,36 +34,177 @@ public abstract class AbstractPreprocessor implements Runnable {
...
+            # useful if Gradle/Ant also requires Maven configured.
+            export MAVEN_HOME=/opt/maven/3.8.8
+            # If we run out of memory we want the JVM to die with error code 134
+            export MAVEN_OPTS="-XX:+CrashOnOutOfMemoryError"
+            # If we run out of memory we want the JVM to die with error code 134
+            export JAVA_OPTS="-XX:+CrashOnOutOfMemoryError"
+            export %s_HOME=/opt/%s/%s
+
+            mkdir -p /var/workdir/workspace/logs /var/workdir/workspace/packages ${HOME}/.sbt/1.0 ${GRADLE_USER_HOME} ${HOME}/.m2
+            cd /var/workdir/workspace/source
```
rnc, line 111: I wonder if I should make all of these values customisable variables so that a person could use the script locally to reproduce issues?

deploy/tasks/pre-build.yaml
```diff
@@ -45,6 +45,10 @@ spec:
       description: The image from the build recipe to use
     - name: BUILD_TOOL
       description: The build tool to use.
```
rnc, line 47: Perhaps I should enhance these descriptions to give examples of allowed values.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
```diff
@@ -76,11 +224,188 @@ public void run() {
...
+              </activation>
+              <properties>
+                <altDeploymentRepository>
+                  local::file:///var/workdir/workspace/artifacts
+                </altDeploymentRepository>
+              </properties>
+            </profile>
+          </profiles>
+        </settings>
+        EOF
```
matejonnet, line 303: The settings.xml for Indy, I took FROM PNC and merged with your
```
<settings>
  <mirrors>
    <mirror>
      <id>indy-mvn</id>
      <mirrorOf>*</mirrorOf>
      <url>
        ${MVN_REPO_DEPENDENCIES_URL}
      </url>
    </mirror>
  </mirrors>

  <servers>
    <server>
      <id>indy-mvn</id>
      <configuration>
        <connectionTimeout>60000</connectionTimeout>
        <httpHeaders>
          <property>
            <name>Authorization</name>
            <value>Bearer ${MVN_TOKEN}</value>
          </property>
        </httpHeaders>
      </configuration>
    </server>
  </servers>

  <proxies>
    <proxy>
      <id>indy-http</id>
      <active>true</active>
      <protocol>http</protocol>
      <host>indy-generic-proxy</host>
      <port>80</port>
      <!-- <username>build-ADDTW3JAGHYAA+tracking</username> -->
      <username>${BUILD_ID}+tracking</username>
      <password>${MVN_TOKEN}</password>
      <nonProxyHosts>indy|localhost</nonProxyHosts>
    </proxy>
    <proxy>
      <id>indy-https</id>
      <active>true</active>
      <protocol>https</protocol>
      <host>indy-generic-proxy</host>
      <port>80</port>
      <username>${BUILD_ID}+tracking</username>
      <password>${MVN_TOKEN}</password>
      <nonProxyHosts>indy|localhost</nonProxyHosts>
    </proxy>
  </proxies>

  <profiles>
    <profile>
      <id>deployment</id>
      <properties>
        <altDeploymentRepository>
          local::file:///var/workdir/workspace/artifacts
        </altDeploymentRepository>
      </properties>
    </profile>
  </profiles>

  <activeProfiles>
    <activeProfile>deployment</activeProfile>
  </activeProfiles>

  <interactiveMode>false</interactiveMode>
  <localRepository>/tmp/.m2/repository</localRepository>

</settings>
```

matejonnet: I would bake the ${MVN_REPO_DEPENDENCIES_URL} and ${BUILD_ID} in the OCI source image as that is required for the reproducibility and leave only ${MVN_TOKEN} (probably not the best name) configurable.

rnc: I think we need to discuss this; I'm not entirely sure I follow or what you're trying to modify?

I don't think the server needs to be configured here as its not deploying? 
I'm not entirely sure about how the proxy section is used either.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
matejonnet: Just realized, the build debug is quite tricky. In addition to getting into a container being locally or remote and to the unshare islotaed namespace. We need to provide an option to interactively connect to the container creating an image by the buildah build command, so the users can resume long running builds once thay fail in the middle eg. mvn -rf.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
matejonnet: > I don't think the server needs to be configured here as its not deploying?
I need to verify this, it might be we need auth for read operation from a per-build repo too

> I'm not entirely sure about how the proxy section is used either.
this is a http proxy, used by the tools that are started by maven, eg. a maven task using wget.
Similar to maven there is an http "repo" (persistent cache) per build.

I believe this is all we need for a build using Indy, as I said it's a copy form PNC configured build, except the altDeploymentRepository

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
matejonnet: A sleep can be added after the mvn command (mvn must not cause exit on failure) and `buildah run <container-id> bash` can be used to get it.

Also added a comment to the JBS-11.

rnc: I don't think we should try and deal with debug in this iteration. We still don't know what is 'allowed' in Konflux so IMHO I don't think we should code to preempt that but establish what can be done first.

rnc: I've changed the script generation to make it easier to override / reuse the run-build.sh.

matejonnet (approved):

PR #2199, opened 2024-10-18
Domain proxy standalone improvements

tecarter94:
Make byte buffer size configurable. Add configurable whitelist of proxy targets. Factor out common IO code. Switch to using Vert.x server. Add support for HTTPS. Handle exceptions seen when connection dies between client and server. Refactored domain proxy in general for consistency.

https://issues.redhat.com/browse/JBS-23


tecarter94:
Tests still need to be added.

matejonnet (approved):

PR #2224, opened 2024-10-24
Use single pipeline

rnc:
Changes to the task yamls are to allow the settings.xml to be stored in the OCI image as well thereby making it available to the deploy step. This will be useful when deploying to Indy.

`tls` workspace references been removed from the tasks as they aren't required (within JBS we're running with `disableTLS` currently so they are unused and wouldn't work in the new layout anyway).  

For JBS, this normalizes to a single pipeline (note that not all of the code from the secondary pipeline added in #1597 and #1625 has been removed/reverted). I also isolated the storage for the verify step so that any testing does not affect the deploy step.


pkg/apis/jvmbuildservice/v1alpha1/systemconfig_types.go
```diff
@@ -47,7 +47,7 @@ type SystemConfigList struct {
 
 const (
 	KonfluxGitDefinition          = "https://raw.githubusercontent.com/konflux-ci/build-definitions/refs/heads/main/task/git-clone/0.1/git-clone.yaml"
-	KonfluxPreBuildDefinitions    = "https://raw.githubusercontent.com/redhat-appstudio/jvm-build-service/main/deploy/tasks/pre-build.yaml"
+	KonfluxPreBuildDefinitions    = "https://raw.githubusercontent.com/rnc/jvm-build-service/PIPELINE/deploy/tasks/pre-build.yaml"
```
matejonnet, line 50: Using `rnc` repo ?

rnc: Task changes always have to be done in two stages - first with the reference to my branch then onces it merged into main it a subsequent PR to change the reference back. Its due to using the git resolver.

PR #2263, opened 2024-11-06
Conditionally activate proxies

rnc:
Potentially a better solution than #2261


rnc:
Closing this in favour of #2261 after discussion with @matejonnet

PR #2327, opened 2024-12-17
Save and copy build.log

ruhan1:
This is for JBS-70. According to @rnc 's comment that directly changing buildah task would "...affect multiple teams. we should...keeping the changes isolated to only our code base", I move the change here.

### DIFF
--- a/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
+++ b/java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
@@ -189,7 +189,7 @@ public abstract class AbstractPreprocessor implements Runnable {
             ENV PROXY_URL=$PROXY_URL
             COPY .jbs/run-build.sh /var/workdir
             COPY . /var/workdir/workspace/source/
-            RUN /var/workdir/run-build.sh
+            RUN /var/workdir/run-build.sh | tee /var/workdir/build.log
             """.formatted(recipeImage);
 
         if (type == ToolType.ANT) {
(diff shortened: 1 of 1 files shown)


java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
```diff
@@ -189,7 +189,7 @@ private String getContainerFile() {
             ENV PROXY_URL=$PROXY_URL
             COPY .jbs/run-build.sh /var/workdir
             COPY . /var/workdir/workspace/source/
-            RUN /var/workdir/run-build.sh
+            RUN /var/workdir/run-build.sh | tee /var/workdir/build.log
```
rnc, line 192: What about error output ?

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
```diff
@@ -203,12 +203,14 @@ private String getContainerFile() {
                     RUN /opt/jboss/container/java/run/run-java.sh copy-artifacts --source-path=/var/workdir/workspace/source --deploy-path=/var/workdir/workspace/artifacts
                     FROM scratch
                     COPY --from=1 /var/workdir/workspace/artifacts /deployment/
+                    COPY --from=0 /var/workdir/build.log /log/
```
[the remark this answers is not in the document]
ruhan1: It seems the build happens in 0 and we should copy from 0.

java-components/build-request-processor/src/main/java/com/redhat/hacbs/container/build/preprocessor/AbstractPreprocessor.java
ruhan1: I guess mostly what we need in the log file is the mvn output. Not sure if we need to care about stderr.

rnc: I'm not sure if we can guarantee that Ant/SBT/Maven/Gradle do **not** send any parts of any logs to stderr ? i.e. I'm just trying to make sure we don't miss anything.

ruhan1: I google and it seem tee can easily do it, like './xx.sh 2>&1 | tee log' . I will update it.

ruhan1: updated.

rnc (approved):

rnc:
Note : the e2e tests will fail due to the addition of domain proxy but JBS is being removed from Konflux so those tests will go away anyway.

PR #2328, opened 2024-12-19
Remove git from prebuild and use PNC Konflux tooling

rnc:
This PR:
* simplifies current pre-build by removing git source functionality from it
* Ensures the mw-pipeline sets a default of the PNC konflux tooling
* keeps JBS current functionality by moving the git source commit into another task and tacking it onto the pipeline


pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -366,87 +380,118 @@ func createPipelineSpec(log logr.Logger, tool string, commitTime int64, jbsConfi
...
+				// This will break various parts of JBS as the PNC tooling does NOT support disabling plugins on
+				// a per build basis.
+				//
+				//{
+				//	Name: "BUILD_PLUGINS",
+				//	Value: tektonpipeline.ParamValue{
+				//		Type:      tektonpipeline.ParamTypeString,
+				//		StringVal: strings.Join(recipe.DisabledPlugins, ","),
+				//	},
+				//},
```
rnc, line 426: Not ideal but now the prebuild is targeted at PNC not JBS and therefore the ability to dynamically remove plugins isn't available (plugin manipulation will be handled earlier in repour).

pkg/reconciler/dependencybuild/buildrecipeyaml.go
```diff
@@ -366,87 +380,118 @@ func createPipelineSpec(log logr.Logger, tool string, commitTime int64, jbsConfi
...
 				{
-					Name: "SCM_HASH",
+					Name: "JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE",
 					Value: tektonpipeline.ParamValue{
 						Type:      tektonpipeline.ParamTypeString,
-						StringVal: db.Spec.ScmInfo.CommitHash,
+						StringVal: "quay.io/redhat-user-workloads/konflux-jbs-pnc-tenant/konflux-tooling:latest",
```
rnc, line 431: Bit of hard-coding here - normally `JVM_BUILD_SERVICE_REQPROCESSOR_IMAGE` is used but now prebuild is targeted at PNC and uses the separate https://github.com/project-ncl/konflux-tooling project we need to split them out and hardcode this.

PR #2338, opened 2025-01-13
Rename whitelist to allowlist

tecarter94:
Rename whitelist to allowlist. Restore ability to override domain proxy version.


rnc (approved):

rnc:
I presume we need https://github.com/project-ncl/domain-proxy/pull/7 merged first?

PR #2354, opened 2025-02-11
Remove references to konflux-tooling image and use released version

rnc:


deploy/tasks/pre-build.yaml
```diff
@@ -38,7 +35,7 @@ spec:
       mountPath: /var/workdir
   steps:
     - name: preprocessor
-      image: $(params.PNC_KONFLUX_TOOLING_IMAGE)
+      image: quay.io/konflux-ci/pnc-konflux-tooling:latest
```
matejonnet, line 38: We should use a version tag instead of latest. IMO the control when to update (or roll-back) should be in the task definition rather than depending on the repository state.

pkg/apis/jvmbuildservice/v1alpha1/systemconfig_types.go
```diff
@@ -47,8 +47,8 @@ type SystemConfigList struct {
 
 const (
 	KonfluxGitDefinition          = "https://raw.githubusercontent.com/konflux-ci/build-definitions/refs/heads/main/task/git-clone/0.1/git-clone.yaml"
-	KonfluxPreBuildDefinitions    = "https://raw.githubusercontent.com/redhat-appstudio/jvm-build-service/main/deploy/tasks/pre-build.yaml"
+	KonfluxPreBuildDefinitions    = "https://raw.githubusercontent.com/rnc/jvm-build-service/TOOLING/deploy/tasks/pre-build.yaml"
```
matejonnet, line 50: Can you use some shared repository instead of personal (rnc)?

deploy/tasks/pre-build.yaml
rnc: Konflux will update the tag to avoid me having to find the SHA

rnc: Konflux pinning https://github.com/redhat-appstudio/jvm-build-service/pull/2315

pkg/apis/jvmbuildservice/v1alpha1/systemconfig_types.go
rnc: https://github.com/redhat-appstudio/jvm-build-service/pull/2376

